Issue #66, opened 2022-04-06
Module Validation - The file "main.json" is modified or outdated.


Gordonby:
During PR of #63 an [error](https://github.com/Azure/bicep-registry-modules/runs/5858711116?check_suite_focus=true#step:6:6) is raised

> Run brm validate
The file "/home/runner/work/bicep-registry-modules/bicep-registry-modules/modules/resources/deployment-scripts-import-acr/main.json" is modified or outdated. Please regenerate the file to fix it.

However the file is up to date, and local validation succeeds
![image](https://user-images.githubusercontent.com/17914476/162074691-17dc030b-5b7d-4c07-827f-973071e07eec.png)

Gordonby:
@shenglol - I pulled this change to my local branch, however on my next push it still fails with the same error.

shenglol:
`.gitattributes` doesn't affect files in working tree. If you clone the repository again it should normalize line endings of all `.bicep` files. Here's the line endings output of a fresh clone on my machine:

```bash
> git ls-files --eol
i/lf    w/crlf  attr/text=auto          .editorconfig
i/lf    w/crlf  attr/text=auto          .eslintrc.js
i/lf    w/crlf  attr/text=auto          .gitattributes
i/lf    w/crlf  attr/text=auto          .github/CODEOWNERS
i/lf    w/crlf  attr/text=auto          .github/ISSUE_TEMPLATE/bug_report.yml
i/lf    w/crlf  attr/text=auto          .github/ISSUE_TEMPLATE/module_proposal.yml
i/lf    w/crlf  attr/text=auto          .github/fabricbot.json
i/lf    w/crlf  attr/text=auto          .github/pull_request_template.md
i/lf    w/crlf  attr/text=auto          .github/workflows/get-changed-module.yml
i/lf    w/crlf  attr/text=auto          .github/workflows/on-pull-request.yml
i/lf    w/crlf  attr/text=auto          .github/workflows/on-push-main.yml
i/lf    w/crlf  attr/text=auto          .github/workflows/publish-module.yml
i/lf    w/crlf  attr/text=auto          .github/workflows/refresh-module-table.yml
i/lf    w/crlf  attr/text=auto          .gitignore
i/lf    w/crlf  attr/text=auto          .prettierignore
i/lf    w/crlf  attr/text=auto          CODE_OF_CONDUCT.md
i/lf    w/crlf  attr/text=auto          CONTRIBUTING.md
[...]

PR #65, opened 2022-04-06
Deployment script module - ACR Import

Gordonby:
## Description

closes #60 

Reraising as when the repo went public all forks were detached. Old PR = #62 

## Adding a new module

- [x] A proposal has been submitted and approved.
- [x] I have included "Closes #{module_proposal_issue_number}" in the PR description.
- [x] I have run `brm validate` locally to verify the module files.
- [x] I have run deployment tests locally to ensure the module is deployable.
- [x] README.md is complete and ready for review

PR #67, opened 2022-04-06
Add .gitattributes

shenglol:
We need to stop Git from changing line endings of Bicep files when checking out the repo to prevent errors like https://github.com/Azure/bicep-registry-modules/pull/65#discussion_r844481706 from happening.


Gordonby:
This closes #66

majastrz (approved):

PR #70, opened 2022-04-07
Codespace config

Gordonby:
## Description

This PR introduces GitHub Codespaces (dev container) configuration that meets all of the prerequisites for `brm`.

This is helpful when the local development environment is restricted or has environmental issues.

The Codespaces container is based on Ubuntu and includes...

- dotnet 6 🖥️ 
- az cli ☁️ 
- bicep 💪
- brm 🦾

### DIFF
--- a/.github/workflows/on-pull-request.yml
+++ b/.github/workflows/on-pull-request.yml
@@ -4,7 +4,7 @@ on:
   pull_request:
     branches:
       - main
-      
+
 jobs:
   get-module-to-validate:
     uses: ./.github/workflows/get-changed-module.yml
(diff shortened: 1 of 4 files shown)


.github/workflows/on-pull-request.yml
```diff
@@ -4,7 +4,9 @@ on:
   pull_request:
     branches:
       - main
-
+    paths:
+      - "modules/**"
```
shenglol, line 8: Need to remove this. The CI job also checks style for non-module files.

shenglol: Another reason is that if the workflow is a required check. Pull requests that do not include changes to modules cannot be merged with the path filtering. See: https://github.community/t/path-filtering-on-required-pull-request-checks/18402.

package-lock.json
```diff
@@ -1,1178 +1,8 @@
 {
   "name": "bicep-registry-modules",
   "version": "1.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
shenglol, line 4: Could we restore the package-lock file? The version change might be due to an old `npm` version. Upgrading `npm` to > v7 should solve the problem.

shenglol (approved):
LGTM. Thanks for adding the codespace config!

Issue #72, opened 2022-04-08
[Enhancement] Add custom removal for additional resources


MrMCake:
The ResourceGroup removal logic used in this repo is already very detailed. However, there are a few more modules worth adding to the current [removal script](https://github.com/Azure/bicep-registry-modules/blob/main/scripts/azure-pipelines/Remove-TestResourceGroup.ps1).

For example
- KeyVault
- Cognitive Services
- API Management
- Log Analytics Workspaces

If we don't implement a corresponding logic, a 2nd test of e.g. a test file deploying a key vault will fail because it does not automatically recover deleted resources.

To speed things up, I'd suggest to re-use parts of the removal logic used in [CARML](https://github.com/Azure/ResourceModules/blob/main/utilities/pipelines/resourceRemoval/helper/Invoke-ResourcePostRemoval.ps1#L99) and align it to the structure used in the [Bicep-Registry-Modules](https://github.com/Azure/bicep-registry-modules/blob/main/scripts/azure-pipelines/Remove-TestResourceGroup.ps1) repository.

cc: @shenglol, @eriqua

shenglol:
Thanks for sharing this!

akata72:
Why not use azure/ResourceModules as the starting point for the public registry. It seems strange to start from «scratch» when most of the basic modules are already created. 😀😀

alex-frankel:
We plan to re-use as much as possible. We are working closely with @MrMCake and the CARML team to get their content hosted in the registry. Both the `virtual-network` and `availability-set` modules come from CARML.

PR #74, opened 2022-04-13
Handle soft delete enabled resource types

shenglol:
Update the resource group cleanup scripts to purge resources that may be soft deleted.

Closes https://github.com/Azure/bicep-registry-modules/issues/72.

### DIFF
--- a/scripts/azure-pipelines/utils/AzureResourceUtils.psm1
+++ b/scripts/azure-pipelines/utils/AzureResourceUtils.psm1
@@ -29,9 +29,9 @@ function Clear-AzRecoveryServicesVault {
     }
     #Invoking API to disable enhanced security
     $body = @{properties = @{enhancedSecurityState = "Disabled" } }
-    $restUri = $Vault.ID + "/backupconfig/vaultconfig?api-version=2020-05-13"
+    $vaultPath = $Vault.ID + "/backupconfig/vaultconfig?api-version=2020-05-13"
 
-    Invoke-RestMethod -Uri $restUri -Body ($body | ConvertTo-JSON -Depth 5) -Method PATCH
+    Invoke-AzRestMethod -Method "PATCH" -Path $vaultPath -Payload ($body | ConvertTo-JSON -Depth 5)
 
     #Fetch all protected items and servers
     $backupItemsVM = Get-AzRecoveryServicesBackupItem -BackupManagementType AzureVM -WorkloadType AzureVM -VaultId $Vault.ID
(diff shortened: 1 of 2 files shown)


shenglol:
@MrMCake do mind taking a look?

majastrz (approved):
:shipit:

scripts/azure-pipelines/utils/AzureResourceUtils.psm1
```diff
@@ -446,6 +446,80 @@ function Remove-AzPrivateEndpointConnectionInResourceGroup {
...
+    [string]$ResourceGroupName
+  )
+  
+  $keyVaults = Get-AzKeyVault -ResourceGroupName $ResourceGroupName
+  
+  foreach ($keyVault in $keyVaults) {
+    Write-Host "Removing Key vault" $keyVault.VaultName "..."
+    Remove-AzKeyVault -VaultName $keyVault.VaultName -ResourceGroupName $ResourceGroupName -Force
+
+    if (-not $keyVault.EnablePurgeProtection) {
```
MrMCake, line 462: if enabled, would it maybe be a good idea to return a warning? For example:

```suggestion
    if ($keyVault.EnablePurgeProtection) {
       Write-Warning ('Key vault [{0}] had purge protection enabled. The retention time is [{1}]. Please wait until after this period before re-running the test.' -f  $keyVault.VaultName, $keyVault.SoftDeleteRetentionInDays)
    }
    else {
```

Alternatively, you could also move the purge outside the parent loop, fetch all purged key vaults via `Get-AzKeyVault -InRemovedState` and iterate over them to print the date when they are deleted (as it is not available otherwise):

```PowerShell
$softDeletedKeyVaults = Get-AzKeyVault -InRemovedState | Where-Object { $_.VaultName -in $keyVaults.VaultName }

foreach($softDeletedKeyVault in $softDeletedKeyVaults) {
  if (-not ($keyVaults | Where-Object { $_.VaultName -eq $softDeletedKeyVault.VaultName}).EnablePurgeProtection) {
      Write-Host "Purging Key vault" $softDeletedKeyVault.VaultName "..."
      Remove-AzKeyVault -VaultName $softDeletedKeyVault.VaultName -ResourceGroupName $ResourceGroupName -InRemovedState -Force
    } else {
       Write-Warning ('Key vault [{0}] had purge protection enabled. It will automatically be purged by [{1}]' -f  $softDeletedKeyVault .VaultName, $softDeletedKeyVault .ScheduledPurgeDate)
    }
}
```

shenglol: Good suggestion! Added a warning for purge protection enabled vaults. The alternative solution also reminds me that we cannot GET the soft delete enabled resources once the test resource group is removed, so I changed `Remove-TestResourceGroup.ps1` to purge those resources before removing the resource group.

MrMCake (approved):

PR #81, opened 2022-04-16
Remove -ResourceGroupName from purging kv cmdlet

shenglol:
Verified via a local command.

### DIFF
--- a/scripts/azure-pipelines/utils/AzureResourceUtils.psm1
+++ b/scripts/azure-pipelines/utils/AzureResourceUtils.psm1
@@ -464,7 +464,7 @@ function Remove-AzKeyVaultInResourceGroup {
     }
     else {
       Write-Host "Purging Key vault" $keyVault.VaultName "..."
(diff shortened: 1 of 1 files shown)


scripts/azure-pipelines/utils/AzureResourceUtils.psm1
```diff
@@ -464,7 +464,7 @@ function Remove-AzKeyVaultInResourceGroup {
     }
     else {
       Write-Host "Purging Key vault" $keyVault.VaultName "..."
-      Remove-AzKeyVault -VaultName $keyVault.VaultName -Location $keyVault.Location -ResourceGroupName $ResourceGroupName -InRemovedState -Force
```
alex-frankel, line 467: Are key vaults unique to a subscription? I'm surprised this works, but as long as we've checked it :)

alex-frankel (approved):

scripts/azure-pipelines/utils/AzureResourceUtils.psm1
shenglol: They are globally unique.

PR #84, opened 2022-04-19
Deployment script module - AKV certificates

### LINKED ISSUE #69: [Module Proposal]: Azure CLI DeploymentScript Module - AKV Certs in AGW

Yes

### Why is the module needed?

Split from #60

[DeploymentScripts](https://docs.microsoft.com/en-us/azure/templates/microsoft.resources/deploymentscripts?tabs=bicep) typically require several Azure resources to be deployed together to execute a meaningful deployment (ID, RBAC and the Script Itself). It lends itself to be in a registry because of this nature.

As well as serving as a meaningful module for customers, it will be directly leveraged in the [Azure/Aks-Construction repo](https://github.com/Azure/AKS-Construction) to reduce complexity in post-deployment automation steps;
- Using the AZ CLI to create self-signed certificates in an Azure Key Vault for use by an Azure Application Gateway Ingress Controller.

> DeploymentScripts are a powerful, yet underutilised resource in Arm, we need more living samples of this resource type being used. 

### Module path

deployment-scripts/create-agw-kv-certificate

### Describe the module

A single bicep file that will deploy a managed identity, apply rbac permissions on existing resources and run an Azure CLI script.

The Azure CLI script will
- Instruct KeyVault to create a self signed certificate based on its default certificate policy
- Add either a root **or** frontend certificate to an existing Application Gateway, referencing the KeyVault certificate

Gordonby:
## Description

Adds a new module which leverages Key Vaults ability to create self-signed certificates.
~~It is also able to use these generated certificates in Azure Application Gateway as either frontend or backend certificates.~~

The script is derived from what is used in [this sample](https://github.com/Azure-Samples/java-aks-keyvault-tls/tree/main/scripts).

Closes #69

## Adding a new module


Gordonby:
Closing this down for the moment as i've still got an optimisation to do in the AGW script. oops.

Gordonby:
Re-Opening as the script looks fixed.

I'm anticipating feedback on "what if the AKV is in a different RG to the AGW" - but am interested in other thoughts/feedback before I add parameters for this use case.

alex-frankel (commented):
Looks good, but paging @bmoore-msft to check if you know there is no way to accomplish this in bicep.

Also, can you confirm this script is idempotent, meaning if this script executes a second time, the end state (including the outputs) will be the same

modules/deployment-scripts/create-agw-kv-certificate/README.md
```diff
@@ -0,0 +1,102 @@
...
+### Example 2
+
+Create multiple self-signed certificates in Azure KeyVault
+
+```bicep
+param location string = resourceGroup().location
+param akvName string =  'yourAzureKeyVault'
+param certificateNames array = [
+  'myapp'
+  'myotherapp
```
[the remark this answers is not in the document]
Gordonby: good spot... shame vscode doesn't problem-validate the bicep that's in a markdown bicep code block 😆

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
```diff
@@ -0,0 +1,229 @@
...
+  scope: agwId
+  name: guid(agwId.id, managedIdentityOperator, useExistingManagedIdentity ? existingDepScriptId.id : newDepScriptId.id)
+  properties: {
+    roleDefinitionId: managedIdentityOperator
+    principalId: useExistingManagedIdentity ? existingDepScriptId.properties.principalId : newDepScriptId.properties.principalId
+    principalType: 'ServicePrincipal'
+  }
+}
+
+@batchSize(1)
```
bmoore-msft, line 127: why 1?  Also, wouldn't it make more sense to loop in the script itself?

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
```diff
@@ -0,0 +1,229 @@
...
+        name: 'certificateWait'
+        value: agwCertImportDelay
+      }
+    ]
+    scriptContent: '''
+      #!/bin/bash
+      set -e
+
+      echo "Waiting on Identity RBAC replication ($initialDelay)"
+      sleep $initialDelay
```
bmoore-msft, line 182: we can be more deterministic by doing a GET on the roleAssignment and breaking if we only see 200's

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
```diff
@@ -0,0 +1,229 @@
...
+
+      echo "Creating AKV Cert $certName"
+      az keyvault certificate create --vault-name $akvName -n $certName -p "$(az keyvault certificate get-default-policy | sed -e s/CN=CLIGetDefaultPolicy/CN=${certName}/g )";
+
+      echo "Waiting for AKV Cert Generation ($certificateWait)"
+      sleep $certificateWait
+
+      echo "getting akv secretid for $certName to add to $agwName";
```
bmoore-msft, line 190: this needs a loop, cert is not guaranteed to be ready...

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
```diff
@@ -0,0 +1,229 @@
...
+      echo $jsonOutputString > $AZ_SCRIPTS_OUTPUT_PATH
+    '''
+    cleanupPreference: cleanupPreference
+  }
+}]
+
+@description('Array of info from each Certificate Deployment Script')
+output createdCertificates array = [for (certName, i) in certNames: {
+  certName: certName
+  DeploymentScriptName: createImportCert[i].name
```
bmoore-msft, line 225: I would limit outputs to those that are relevant to the scenario... if debugging is needed then a few thoughts
1) the script isn't robust enough - e.g. you wouldn't need this info output for a deployment of another resource type...
2) the subscription will still have details of the script that can be used to help debug

IOW, we're letting the implementation (of the deploymentScript) bleed through the module.

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
```diff
@@ -0,0 +1,229 @@
...
+      unversionedSecretId=$(echo $versionedSecretId | cut -d'/' -f-5) # remove the version from the url;
+
+      case $agwCertType in
+        none)
+          echo "AppGw Certificate Creation not configured"
+          ;;
+
+        root-cert)
+          echo "Creating root certificate in application gateway";
+          rootcertcmd="az network application-gateway root-cert create --gateway-name $agwName -g $RG -n $certName --key-vault-secret-id $unversionedSecretId";
```
bmoore-msft, line 201: adding the cert is a PUT as I understand it, which means this should be done outside of the script...

Gordonby:
Some great feedback from @bmoore-msft 
I think this module will be best down-scoped to just KeyVault.
I'll work on the suggested changes, then respond to the PR discussions individually. I think all but 2 i'm happy to implement without further discussion.

modules/deployment-scripts/create-agw-kv-certificate/README.md
```diff
@@ -0,0 +1,102 @@
+# Key Vault Certificate Creation
+
+Create Key Vault self-signed certificates and optionally integrate with Application Gateway
```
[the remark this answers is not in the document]
Gordonby: Agreed. I will down-scope this PR to just KeyVault

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
```diff
@@ -0,0 +1,229 @@
...
+resource akv 'Microsoft.KeyVault/vaults@2021-11-01-preview' existing = {
+  name: akvName
+}
+
+resource agw 'Microsoft.Network/applicationGateways@2021-05-01' existing = if (!empty(agwName) && agwCertType != 'none') {
+  name: agwName
+}
+
+@description('The User Assigned Identity of the Azure Application Gateway')
+param agwIdName string = ''
```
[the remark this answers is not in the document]
Gordonby: AGW will be pulled out, but yes... I have a certain reluctance to add all the params to the top of the file. It feels like a hangup from Arm templates, but i do understand we need some convention to these modules.

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
```diff
@@ -0,0 +1,229 @@
...
+param agwName string = ''
+
+@description('The location to deploy the resources to')
+param location string = resourceGroup().location
+
+@description('How the deployment script should be forced to execute')
+param forceUpdateTag  string = utcNow()
+
+@description('Version of the Azure CLI to use')
+param azCliVersion string = '2.30.0'
```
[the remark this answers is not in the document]
Gordonby: I think i have a reluctance to hard code these when they can be defaulted and exposed as params.
The retention can be useful when testing, when you might want a shorter period.

I do take the point that we're not creating a generic module, and the properties should be specific to the module to reduce confusion. I'll make the changes.

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
```diff
@@ -0,0 +1,229 @@
...
+param retention string = 'P1D'
+
+@description('An array of Azure Key Vault RoleIds that are required for the DeploymentScript resource')
+param rbacRolesNeededOnKV array = [
+  'a4417e6f-fecd-4de8-b567-7b0420556985' //KeyVault Certificate Officer
+]
+
+@description('An array of Azure Application Gateway RoleIds that are required for the DeploymentScript resource')
+param rbacRolesNeededOnAppGw array = [
+  'b24988ac-6180-42a0-ab88-20f7382dd24c' //Contributor
```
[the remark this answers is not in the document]
Gordonby: AppGw is going to be removed from this module

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
Gordonby: AppGw seems to have concurrency problems with multiple updates at the same time. Batchsize was added to mitigate this

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
Gordonby: I'll remove the datetime info, but the KV id's will be useful to modules that consume this.

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
```diff
@@ -0,0 +1,229 @@
...
+  'Always'
+])
+@description('When the script resource is cleaned up')
+param cleanupPreference string = 'OnSuccess'
+
+resource akv 'Microsoft.KeyVault/vaults@2021-11-01-preview' existing = {
+  name: akvName
+}
+
+resource agw 'Microsoft.Network/applicationGateways@2021-05-01' existing = if (!empty(agwName) && agwCertType != 'none') {
```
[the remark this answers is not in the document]
Gordonby: this allowed the module to function to just create certificates.
since this module is going to be downscoped to just KV, i'll resolve this discussion

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
```diff
@@ -0,0 +1,229 @@
...
+  location: location
+}
+
+@description('An existing managed identity that could exist in another sub/rg')
+resource existingDepScriptId 'Microsoft.ManagedIdentity/userAssignedIdentities@2018-11-30' existing = if (useExistingManagedIdentity ) {
+  name: managedIdentityName
+  scope: resourceGroup(existingManagedIdentitySubId, existingManagedIdentityResourceGroupName)
+}
+
+resource rbacKv 'Microsoft.Authorization/roleAssignments@2020-08-01-preview' = [for roleDefId in rbacRolesNeededOnKV: {
```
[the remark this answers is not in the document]
Gordonby: unsupported scenario. I'll make this clear in the readme

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
```diff
@@ -0,0 +1,229 @@
...
+  'ssl-cert'
+])
+@description('Configured certificate in Application Gateway as Frontend (ssl-cert) or Backend (root-cert)')
+param agwCertType string = 'none'
+
+@description('A delay before the script import operation starts. Primarily to allow Azure AAD Role Assignments to propagate')
+param initialScriptDelay string = '30s'
+
+@description('A delay before the Application Gateway imports the Certificate from KeyVault. Primarily to allow the certificate creation operation to complete')
+param agwCertImportDelay string = '120s'
```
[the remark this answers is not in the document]
Gordonby: removed.

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
```diff
@@ -0,0 +1,229 @@
...
+param azCliVersion string = '2.30.0'
+
+@description('Deployment Script timeout')
+param timeout string = 'PT30M'
+
+@description('The retention period for the deployment script')
+param retention string = 'P1D'
+
+@description('An array of Azure Key Vault RoleIds that are required for the DeploymentScript resource')
+param rbacRolesNeededOnKV array = [
```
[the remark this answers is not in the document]
Gordonby: the edge case here is a slim one.
if custom roles are being used, having it as an array would allow multiple to be provided.

happy to scale this back down to a single role. 🚀 

multiple roles is more likely in the other modules i'm writing, but using arrays for RBAC is just my normal workflow now 🐱

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
```diff
@@ -0,0 +1,229 @@
...
+  properties: {
+    roleDefinitionId: resourceId('Microsoft.Authorization/roleDefinitions', roleDefId)
+    principalId: useExistingManagedIdentity ? existingDepScriptId.properties.principalId : newDepScriptId.properties.principalId
+    principalType: 'ServicePrincipal'
+  }
+}]
+
+var managedIdentityOperator = resourceId('Microsoft.Authorization/roleDefinitions', 'f1a07417-d97a-45cb-824c-7a7467783830')
+@description('Managed identity requires "Managed Identity Operator" permission over the user assigned identity of the Application Gateway.')
+resource agwIdRbac 'Microsoft.Authorization/roleAssignments@2020-08-01-preview' =  if (!empty(agwName) && agwCertType != 'none') {
```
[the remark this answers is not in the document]
Gordonby: This was a nuance of AppGw... Not worth worrying about now since we're just doing keyvault in this module

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
Gordonby: agreed

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
Gordonby: Aren't there nuances if the SP will be querying its own role assignments? 
Just running a;

```azurecli
az role assignment list --assignee $spClientId
```

returns an empty array.

I'm not sure if adding more RBAC just so the SP can check it's own RBAC is going to be worth it.

If you have a sample of this working well @bmoore-msft i'd be keen to look

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
Gordonby: i've removed the sleep.

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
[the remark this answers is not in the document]
Gordonby: going to resolve this discussion as it's covered by this issue #88

Gordonby:
Getting another strange CI failure @shenglol - any insight? It works fine locally.

> An error occurred: Exception calling "Invoke" with "0" argument(s): "The running command stopped because the preference variable "ErrorActionPreference" or common parameter is set to Stop: 
Cannot find deleted vault 'akvtest3vbarm34t26ck' in location 'westus'"

alex-frankel:
Thanks for all the feedback @bmoore-msft and iteration @Gordonby.

In general, I am aligned with Brian's thinking, though I had one thought that may split the difference:

Should we have a `generic-deployment-script` module that is the foundation for all these specific scenarios in which to use a deployment script. The `generic-deployment-script` would expose everything (or mostly everything) with good defaults, then the `create-agw-kv-certificate` would use that module in its definition and not expose the same degree of parameters. Thoughts?

shenglol:
> Getting another strange CI failure @shenglol - any insight? It works fine locally.
> 
> > An error occurred: Exception calling "Invoke" with "0" argument(s): "The running command stopped because the preference variable "ErrorActionPreference" or common parameter is set to Stop:
> > Cannot find deleted vault 'akvtest3vbarm34t26ck' in location 'westus'"

That's an error in cleaning up the resource group. It seems that deleting a key vault is an async operation inside the backend and we should wait some time before purging the key vault. I'm working on a fix now.

shenglol:
Closing and reopening the PR to run the latest Deployment Test - CI.

modules/deployment-scripts/create-kv-certificate/test/main.test.bicep
```diff
@@ -0,0 +1,47 @@
...
+resource akv 'Microsoft.KeyVault/vaults@2021-11-01-preview' = {
+  name: akvName
+  location: location
+  properties: {
+    sku: {
+      family: 'A'
+      name: 'standard'
+    }
+    tenantId: subscription().tenantId
+    enableSoftDelete: false
```
shenglol, line 17: Ahh this explains why the CI failed to purge the key vault. Because soft delete is disabled, the CI will never find the key vault in deleted state. To fix the CI failure we can change value to true (or just remove the property as the default is true). Key vault is also going to deprecate the ability to opt out of soft-delete soon: https://docs.microsoft.com/en-us/azure/key-vault/general/soft-delete-change.

I will update the CI to skip purging for soft-delete disabled key vaults.

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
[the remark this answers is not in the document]
bmoore-msft: I have similar reluctance, though I think my desire for a "clean contract" is a bit stronger...  definitely need more discussion as we're blazing trails here...

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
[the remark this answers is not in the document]
bmoore-msft: We can always expand later...

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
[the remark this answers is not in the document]
bmoore-msft: I think RBAC only is fine, we should just make it clear...  also wondering if we need to do something beyond the readme (full disclosure: no idea what that would be...)

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
bmoore-msft: We do something similar here: https://github.com/Azure/azure-quickstart-templates/blob/master/test/ci-scripts/Create-ResourceGroupAssignment.ps1 - pass in the ```rbacKv.id``` and just do a few GETs on it.  But I think I agree with you on the extra perms the MSI would need to query the graph, it's not as elegant.

The other thing we could do is try/catch on the ```az keyvault certificate create```.  If it's failing with 4xx, then we know we can retry within reason.  Chances are, by the time the deploymentScript resource spins up, the role has replicated...  So you wouldn't need many retries IMO.

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
bmoore-msft: agreed - was thinking more of just the deploymentScript goo...

modules/deployment-scripts/create-kv-certificate/test/main.test.bicep
bmoore-msft: Don't we deploy each pipeline run to a unique RG?  Then it's just a matter of using uniqueString() to name resources with global names and we can let the platform deal with the purge (and authors don't have to worry about how to set this for every given resource).

bmoore-msft:
> Thanks for all the feedback @bmoore-msft and iteration @Gordonby.
> 
> In general, I am aligned with Brian's thinking, though I had one thought that may split the difference:
> 
> Should we have a `generic-deployment-script` module that is the foundation for all these specific scenarios in which to use a deployment script. The `generic-deployment-script` would expose everything (or mostly everything) with good defaults, then the `create-kv-certificate` would use that module in its definition and not expose the same degree of parameters. Thoughts?

Do we have a ```generic deployment script``` module?  If so, I do love the idea...

modules/deployment-scripts/create-kv-certificate/main.bicep
```diff
@@ -0,0 +1,143 @@
...
+
+      jsonOutputString=$(jq -n --arg cn "$certName" --arg sid "$unversionedSecretId" --arg vsid "$versionedSecretId" '{certName: $cn, certSecretId: {versioned: $vsid, unversioned: $sid }}')
+      echo $jsonOutputString > $AZ_SCRIPTS_OUTPUT_PATH
+    '''
+    cleanupPreference: cleanupPreference
+  }
+}]
+
+@description('Array of info from each Certificate Deployment Script')
+output createdCertificates array = [for (certName, i) in certNames: {
```
bmoore-msft, line 139: another thought -- we should output the thumbprint too... many scenarios that need certs need that and there's no way to retrieve it via the control plane.  It does mean, the script would have to poll to wait until the cert is created.

Another sample: https://github.com/bmoore-msft/AzureRM-Samples/blob/master/ignite-2019/Create-Cert.ps1#L17

alex-frankel:
> > Thanks for all the feedback @bmoore-msft
> >             
> >             Brian Moore (AZURE RESOURCE MANAGER)
> >             FTE and iteration @Gordonby
> >             
> >             Gordon Byers
> >             FTE.
> > In general, I am aligned with Brian's thinking, though I had one thought that may split the difference:
> > Should we have a `generic-deployment-script` module that is the foundation for all these specific scenarios in which to use a deployment script. The `generic-deployment-script` would expose everything (or mostly everything) with good defaults, then the `create-kv-certificate` would use that module in its definition and not expose the same degree of parameters. Thoughts?
> 
> Do we have a `generic deployment script` module? If so, I do love the idea...

We don't but we can make one, or @Gordonby or @MrMCake can have right of first refusal :)

modules/deployment-scripts/create-kv-certificate/test/main.test.bicep
shenglol: Yeah, but we may need a linter rule to enforce that. I think we should still prevent the CI from failing in case the name of the key vault is hard coded in the test file. If the key vault is not purged, the subsequent deployment will fail. This was suggested by the CARML team btw.

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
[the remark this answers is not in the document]
Gordonby: YAGNI. 😉

modules/deployment-scripts/create-kv-certificate/main.bicep
Gordonby: added the thumbprint as output.

also wrapped this in a loop, but the `az keyvault certificate create` operation seems to be synchronous - so the loop seems unnecessary.

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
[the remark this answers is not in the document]
Gordonby: i can get behind that approach 👍 

i've made the changes to the param surface area.

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
Gordonby: Try/Catch seems the best option here.
I'll leave the initialDelay parameter in there as a cushion though.

Bash doesn't really do try/catch, so i've looped it with a maximum retry attempt.
Tested, and works nice. I even specifically denied the RBAC assignment and it fails gracefully.

Gordonby:
> We don't but we can make one, or @Gordonby or @MrMCake can have right of first refusal :)

There is a module for this resource in CARML. If it's suitable, then @MrMCake should probably PR.

modules/deployment-scripts/create-agw-kv-certificate/main.bicep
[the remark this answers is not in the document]
Gordonby: Not sure there's a good answer here.

We could elevate the Identity permission to be able to check if the AKV is using RBAC instead of access policies in the script... Definitely increasing the fault surface area though.

The *best* way would be to reference the akv.properties.enableRbacAuthorization as a resource condition. However it needs to be calculated at the start.

Suggest that we just document in the readme for the time being. I'll leave this discussion open until we have a conclusion

modules/deployment-scripts/create-kv-certificate/test/main.test.bicep
Gordonby: > Ahh this explains why the CI failed to purge the key vault. Because soft delete is disabled, the CI will never find the key vault in deleted state. To fix the CI failure we can change value to true (or just remove the property as the default is true). Key vault is also going to deprecate the ability to opt out of soft-delete soon: https://docs.microsoft.com/en-us/azure/key-vault/general/soft-delete-change.

For the local development loop, avoiding soft delete is a must.
Since the team's going to deprecate that ability, i'll base that value off the existance of a tag to get the best experience in CI and locally.

```bicep
enableSoftDelete: ! contains(resourceGroup().tags,'kvSoftDelete')
```

Gordonby: @shenglol - Another CI issue i'm afraid. Works locally, but not in CI.... The classic, "It works on my machine 👅 "

> Unable to process template language expressions for resource '/subscriptions/3ce675b8-fcea-4732-a965-393bbe02d7ad/resourceGroups/84-e23e630-20220422123703-8fe43199e63746e6bcdf864009574c0d/providers/Microsoft.KeyVault/vaults/akvtestjj2zsmhppdpww' at line '24' and column '9'. 'The language expression property 'tags' doesn't exist, available properties are 'id, name, type, location, properties'.

modules/deployment-scripts/create-kv-certificate/README.md
```diff
@@ -0,0 +1,74 @@
...
+| `managedIdentityName`                      | `string` | No       | Name of the Managed Identity resource                                                                         |
+| `existingManagedIdentitySubId`             | `string` | No       | For an existing Managed Identity, the Subscription Id it is located in                                        |
+| `existingManagedIdentityResourceGroupName` | `string` | No       | For an existing Managed Identity, the Resource Group it is located in                                         |
+| `certNames`                                | `array`  | Yes      | An array of Certificate names to create                                                                       |
+| `initialScriptDelay`                       | `string` | No       | A delay before the script import operation starts. Primarily to allow Azure AAD Role Assignments to propagate |
+| `cleanupPreference`                        | `string` | No       | When the script resource is cleaned up                                                                        |
+
+## Outputs
```
bmoore-msft, line 21: need to update outputs section - but see comment below first

modules/deployment-scripts/create-kv-certificate/main.bicep
```diff
@@ -0,0 +1,148 @@
...
+      jsonOutputString=$(echo $createdCert | jq --arg usid $unversionedSecretId '{name: .name ,certSecretId: {versioned: .sid, unversioned: $usid }, thumbprint: .x509Thumbprint, thumbprintHex: .x509ThumbprintHex}')
+      echo $jsonOutputString > $AZ_SCRIPTS_OUTPUT_PATH
+    '''
+    cleanupPreference: cleanupPreference
+  }
+}]
+
+@description('Array of info from each Certificate Deployment Script')
+output createdCertificates array = [for (certName, i) in certNames: {
+  certName: certName
```
bmoore-msft, line 145: Ok, now seeing this really come together and looking at the samples in the readme I had a few more thoughts...

1) what's the difference between `certName` and `deploymentScriptName`?  Seems they would be identical and it means the user has to understand the difference and which to use?
2) having an output named `deploymentScript*` seemed odd to me, again it really exposes the implementation which leads me to (one of) 2 things:
a) let's just call the outputs the specific things we think are relevant to the scenario (versionedUri, thumbprint) rather than a custom object
b) perhaps even better, simply output the response from `az keyvault certificate show` - that way 1) all of the properties will be available (if we miss a scenario) and 2) it's not a custom object - we return the documented response from the platform verbatim.  We can simply point to the REST api documentation and should the day come where this is available in the template language, switching costs are trivial - i.e. developers wouldn't have to update any of the code that consumes outputs.

Thoughts?

bmoore-msft:
> That's an error in cleaning up the resource group. It seems that deleting a key vault is an async operation inside the backend and we should wait some time before purging the key vault. I'm working on a fix now.

Why would we need to purge the vault?

bmoore-msft (changes requested):

modules/deployment-scripts/create-kv-certificate/README.md
Gordonby: I think what complicates the outputs is the fact that it works with **arrays** of certificates.
I'm starting to wonder if it should just work with a single certificate, and the calling module can do the loops.
@bmoore-msft

bmoore-msft: Oooooo.... I think you might be right.  That simplifies things quite a bit throughout.  Parameters (type safety), the script itself and outputs.  Let's do it.

modules/deployment-scripts/create-kv-certificate/main.bicep
bmoore-msft: Crap - you're right (about outputs).  But what's the secret property(s) in the get?  The cer data? (everything else looks fine but I don't know a ton about certs).  But also don't want to optimize too much for one scenario (AppGW)...

Failing that we can also just use the same schema of the response - so as we add properties. Add them with the same property name and json path.  But in practice I only know a a need for id, kid, sid, and x5t... so making each one of those a string is a great place for us to land too...

modules/deployment-scripts/create-kv-certificate/README.md
Gordonby: done!

modules/deployment-scripts/create-kv-certificate/test/main.test.bicep
Gordonby: @shenglol  - i've reverted the change that was causing the issue, but this might still be a good one to look at.

bmoore-msft (approved):

modules/deployment-scripts/create-kv-certificate/main.bicep
bmoore-msft: LGTM

bmoore-msft (approved):

modules/deployment-scripts/create-kv-certificate/test/main.test.bicep
shenglol: According to https://github.com/MicrosoftDocs/azure-docs/issues/59033#issuecomment-662274297 the error will be returned if we try to call `resourceGroup().tags` on a resource group that was created without specifying the tags property (seems to be a bug in the resource group type🤦‍♂️). I'll change the CI to set the tags property for each test resource group.

shenglol:
I'll need to update the MCR manifest before merging the PR since the module name is changed.

shenglol (approved):
Thanks for adding the module!

PR #102, opened 2022-04-25
Contribute ManagedIdentity module - Closes #98

### LINKED ISSUE #98: [Module Proposal]: Managed Identity

Yes

### Why is the module needed?

We need a module to simplify Managed Identity creation.

### Module path

managed-identity/user-assigned-identities

### Describe the module

Taken from the [CARML](https://aka.ms/CARML) repository, the module contains a Managed Identity resource. It sets default values for common properties and allows users to override them via parameters. It further inclused extension resources such as RBAC & locks.

MrMCake:
## Description

Contribute module [CARML module](https://github.com/Azure/ResourceModules/tree/main/arm/Microsoft.ManagedIdentity/userAssignedIdentities) `ManagedIdentities` with issue #98 .

## Adding a new module

<details>

<summary>Successful test</summary>

```
> New-AzResourceGroupDeployment -ResourceGroupName 'test-Rg' -TemplateFile '.\test\main.test.bicep' -Verbose
VERBOSE: Performing the operation "Creating Deployment" on target "test-Rg".
VERBOSE: 22:02:52 - Template is valid.
VERBOSE: 22:02:54 - Create template deployment 'main.test'
VERBOSE: 22:02:54 - Checking deployment status in 5 seconds
VERBOSE: 22:02:59 - Resource Microsoft.Resources/deployments 'bku2phqmh3aei-gen-uai' provisioning status is running
VERBOSE: 22:02:59 - Resource Microsoft.ManagedIdentity/userAssignedIdentities 'adp-avs-az-msi-x-01' provisioning status is succeeded
VERBOSE: 22:02:59 - Resource Microsoft.Resources/deployments 'bku2phqmh3aei-min' provisioning status is running
VERBOSE: 22:02:59 - Resource Microsoft.ManagedIdentity/userAssignedIdentities 'avs-az-uai-min-01' provisioning status is succeeded
VERBOSE: 22:02:59 - Checking deployment status in 14 seconds
VERBOSE: 22:03:14 - Resource Microsoft.Resources/deployments 'enurxnncijege-UserMSI-Rbac-0' provisioning status is running
VERBOSE: 22:03:14 - Resource Microsoft.Authorization/roleAssignments 'ab48a99c-6adc-5024-9d90-4e3affb1c43c' provisioning status is succeeded
VERBOSE: 22:03:14 - Resource Microsoft.ManagedIdentity/userAssignedIdentities 'avs-az-uai-gen-01' provisioning status is succeeded
VERBOSE: 22:03:14 - Resource Microsoft.Resources/deployments 'bku2phqmh3aei-min' provisioning status is succeeded
VERBOSE: 22:03:14 - Checking deployment status in 15 seconds
VERBOSE: 22:03:30 - Resource Microsoft.Resources/deployments 'bku2phqmh3aei-gen-uai' provisioning status is succeeded
VERBOSE: 22:03:30 - Resource Microsoft.Resources/deployments 'enurxnncijege-UserMSI-Rbac-0' provisioning status is succeeded

DeploymentName          : main.test
ResourceGroupName       : test-Rg
ProvisioningState       : Succeeded
Timestamp               : 25.04.2022 20:03:28
Mode                    : Incremental
TemplateLink            : 
Parameters              : 
                          Name             Type                       Value
                          ===============  =========================  ==========
                          location         String                     "westeurope"
                          serviceShort     String                     "avs"

Outputs                 : 
DeploymentDebugLogLevel : 
```

</details>

### DIFF
--- a/modules/managed-identity/user-assigned-identity/.bicep/nested_rbac.bicep
+++ b/modules/managed-identity/user-assigned-identity/.bicep/nested_rbac.bicep
@@ -1,0 +1,55 @@
+@sys.description('Required. The IDs of the prinicpals to assign to role to')
+param principalIds array
+
+@sys.description('Required. The name of the role to assign. If it cannot be found you can specify the role definition ID instead')
+param roleDefinitionIdOrName string
+
+@sys.description('Required. The resource ID of the resource to apply the role assignment to')
+param resourceId string
+
+@sys.description('Optional. The principal type of the assigned principal ID.')
+@allowed([
+  'ServicePrincipal'
+  'Group'
+  'User'
+  'ForeignGroup'
+  'Device'
+  ''
+])
+param principalType string = ''
+
+@sys.description('Optional. Description of role assignment')
+param description string = ''
+
+var builtInRoleNames = {
+  'Owner': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '8e3af657-a8ff-443c-a75c-2fe8c4bcb635')
+  'Contributor': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', 'b24988ac-6180-42a0-ab88-20f7382dd24c')
+  'Reader': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', 'acdd72a7-3385-48ef-bd42-f606fba81ae7')
+  'Log Analytics Contributor': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '92aaf0da-9dab-42b6-94a3-d43ce8d16293')
+  'Log Analytics Reader': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '73c42c96-874c-492b-b04d-ab87d138a893')
+  'Managed Application Contributor Role': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '641177b8-a67a-45b9-a033-47bc880bb21e')
+  'Managed Application Operator Role': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', 'c7393b34-138c-406f-901b-d8cf2b17e6ae')
+  'Managed Applications Reader': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', 'b9331d33-8a36-4f8c-b097-4f54124fdb44')
+  'Managed Identity Contributor': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', 'e40ec5ca-96e0-45a2-b4ff-59039f2c2b59')
+  'Managed Identity Operator': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', 'f1a07417-d97a-45cb-824c-7a7467783830')
+  'Monitoring Contributor': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '749f88d5-cbae-40b8-bcfc-e573ddc772fa')
+  'Monitoring Metrics Publisher': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '3913510d-42f4-4e42-8a64-420c390055eb')
+  'Monitoring Reader': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '43d0d8ad-25c7-4714-9337-8ba259a9fe05')
+  'Resource Policy Contributor': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '36243c78-bf99-498c-9df9-86d9f8d28608')
+  'User Access Administrator': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '18d7d88d-d35e-4fb5-a5c3-7773c20a72d9')
+}
+
+resource userMsi 'Microsoft.ManagedIdentity/userAssignedIdentities@2018-11-30' existing = {
+  name: last(split(resourceId, '/'))
+}
+
+resource roleAssignment 'Microsoft.Authorization/roleAssignments@2021-04-01-preview' = [for principalId in principalIds: {
+  name: guid(userMsi.name, principalId, roleDefinitionIdOrName)
+  properties: {
+    description: description
+    roleDefinitionId: contains(builtInRoleNames, roleDefinitionIdOrName) ? builtInRoleNames[roleDefinitionIdOrName] : roleDefinitionIdOrName
+    principalId: principalId
+    principalType: !empty(principalType) ? principalType : null
+  }
+  scope: userMsi
+}]
--- a/modules/managed-identity/user-assigned-identity/README.md
+++ b/modules/managed-identity/user-assigned-identity/README.md
@@ -6,7 +6,7 @@ This module deploys Microsoft.ManagedIdentity User Assigned Identities and optio
 
 | Name              | Type     | Required | Description                                                                                                                                                                                                                                                                                                                                                                                                    |
 | :---------------- | :------: | :------: | :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `name`            | `string` | No       | Optional. Name of the User Assigned Identity.                                                                                                                                                                                                                                                                                                                                                                  |
+| `name`            | `string` | Yes      | Required. Name of the User Assigned Identity.                                                                                                                                                                                                                                                                                                                                                                  |
 | `location`        | `string` | No       | Optional. Location for all resources.                                                                                                                                                                                                                                                                                                                                                                          |
 | `lock`            | `string` | No       | Optional. Specify the type of lock.                                                                                                                                                                                                                                                                                                                                                                            |
 | `roleAssignments` | `array`  | No       | Optional. Array of role assignment objects that contain the 'roleDefinitionIdOrName' and 'principalId' to define RBAC role assignments on this resource. In the roleDefinitionIdOrName attribute, you can provide either the display name of the role definition, or its fully qualified ID in the following format: '/providers/Microsoft.Authorization/roleDefinitions/c2f4ef07-c644-48eb-af81-4b1b4947fb11' |
@@ -28,7 +28,7 @@ This module deploys Microsoft.ManagedIdentity User Assigned Identities and optio
 Deployment of `1` user assigned identity using the minimum required parameters.
 
 ```bicep
-module minMSI 'br/public:managed-identity/user-assigned-identity:1.0' = {
+module minMSI 'br/public:managed-identity/user-assigned-identity:1.0.1' = {
   name: '${uniqueString(deployment().name, 'WestEurope')}-min-msi'
   params: {
     name: 'carml-az-uai-min-01'
@@ -45,7 +45,7 @@ Deployment of `1` user assigned identity with
 - `1` role assignment for `1` identity
 
 ```bicep
-module genMSI '../main.bicep' = {
+module genMSI 'br/public:managed-identity/user-assigned-identity:1.0.1' = {
   name: '${uniqueString(deployment().name, 'WestEurope')}-gen-msi'
   params: {
     name: 'carml-az-uai-gen-01'
--- a/modules/managed-identity/user-assigned-identity/README.md
+++ b/modules/managed-identity/user-assigned-identity/README.md
@@ -1,0 +1,67 @@
+# User Assigned Identities
+
+This module deploys Microsoft.ManagedIdentity User Assigned Identities and optionally available children or extensions
+
+## Parameters
+
+| Name              | Type     | Required | Description                                                                                                                                                                                                                                                                                                                                                                                                    |
+| :---------------- | :------: | :------: | :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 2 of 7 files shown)


modules/managed-identity/user-assigned-identity/main.bicep
```diff
@@ -0,0 +1,57 @@
+@description('Optional. Name of the User Assigned Identity.')
+param name string = guid(resourceGroup().id)
```
alex-frankel, line 2: Does this need to be more unique? What if I use this module twice in the same RG?

modules/managed-identity/user-assigned-identity/README.md
```diff
@@ -0,0 +1,68 @@
...
+| name              | string | The name of the user assigned identity                          |
+| resourceId        | string | The resource ID of the user assigned identity                   |
+| principalId       | string | The principal ID of the user assigned identity                  |
+| resourceGroupName | string | The resource group the user assigned identity was deployed into |
+
+## Examples
+
+### Example 1
+
+Deployment of `1` user assigned identity using the minimum required parameters.
```
alex-frankel, line 28: Why are the numbers code-fenced?

modules/managed-identity/user-assigned-identity/main.bicep
```diff
@@ -0,0 +1,57 @@
...
+@description('Optional. Specify the type of lock.')
+param lock string = 'NotSpecified'
+
+@description('Optional. Array of role assignment objects that contain the \'roleDefinitionIdOrName\' and \'principalId\' to define RBAC role assignments on this resource. In the roleDefinitionIdOrName attribute, you can provide either the display name of the role definition, or its fully qualified ID in the following format: \'/providers/Microsoft.Authorization/roleDefinitions/c2f4ef07-c644-48eb-af81-4b1b4947fb11\'')
+param roleAssignments array = []
+
+@description('Optional. Tags of the resource.')
+param tags object = {}
+
+resource userMsi 'Microsoft.ManagedIdentity/userAssignedIdentities@2018-11-30' = {
```
alex-frankel, line 21: What is this module helping abstract in this particular case? There are no special restrictions on managed identity resource names, and there are no additional properties to set that are complicated. It seems the "raw" resource is simpler than using this module since with the module I would be exposed to extra parameters for locking and rbac. 

I guess I would not need to set the name & location myself if I use the module, but that is very basic and something we would want all bicep developers to be able to do.

alex-frankel (commented):
Have a question about the general goals and expectations of what this module allows me to do. I may be getting stuck on the inclusion of locks and rbac in a module that is otherwise a very simple resource declaration.

modules/managed-identity/user-assigned-identity/main.bicep
MrMCake: It's a fair point. Can't say I'm a big fan of auto-generated names anyways. Will remove it

modules/managed-identity/user-assigned-identity/README.md
MrMCake: To highlight it. Alternatively I could make them **bold**, but I personally prefer `code-fenced`.  Happy to undo it, if you rather prefer plain text

modules/managed-identity/user-assigned-identity/main.bicep
MrMCake: In this particular case, there are no child-modules, so that leaves us with the extension resources - that is RBAC that you can deploy together with the MSI.

modules/managed-identity/user-assigned-identity/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+```bicep
+module minMSI 'br/public:managed-identity/user-assigned-identity:1.0' = {
```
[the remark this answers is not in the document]
MrMCake: Updated

modules/managed-identity/user-assigned-identity/main.bicep
alex-frankel: I guess I am still ok checking this in for now, but TBH, I'm growing more concerned about this pattern when I see modules like this. 

I'd prefer customers learn how to do extension resources once, then be able to repeat it as they need it. Hiding it inside of a module allows the module to be more functional at the cost of user education. Would you be willing to submit all CARML modules without those extension resources? 

If we have those, then you could in theory make an "L2" module that combines this extension-less modules with rbac + locks. I recognize that is more work, but feels like the right layering. Thoughts?

modules/managed-identity/user-assigned-identity/README.md
MrMCake: done

modules/managed-identity/user-assigned-identity/main.bicep
MrMCake: I agree we shold have a fundamental discussion about some of the patterns we apply (i.e. extension resources, what we optimize for in CARML, and also what vice-versa what you want to go for with the bicep registry). Probably not in a comment section in GitHub though as this would turn into an essay ;)

We could publish without the extension resources (that leaves the child-resources as the biggest addition), but I'd argue that it wouldn't make sense & would not be worthwhile doing as we're then clearly engineering into 2 different directions - making it increasingly difficult & time intensive to re-engineer the modules to account for the registry's intended module design.

MrMCake:
> Have a question about the general goals and expectations of what this module allows me to do. I may be getting stuck on the inclusion of locks and rbac in a module that is otherwise a very simple resource declaration.

In CARML we follow a fundamental approach towards Azure resources, that is, to make the Azure Resource Reference as usable as possible while applying smart & secure defaults along-side added capabilities that one would otherwise need to manually re-invent (e.g. endpoints+DNS, RBAC, Public IPs, Diagnostic Settings etc.) that should cater for as many scenarios as possible (while being optional). 
As such, we don't apply a case-by-case discussion whether or not to e.g. implement RBAC for the MSI module - instead, if we create a module, we aim to add the features by default, because they are supported.

modules/managed-identity/user-assigned-identity/main.bicep
```diff
@@ -0,0 +1,57 @@
+@description('Required. Name of the User Assigned Identity.')
+param name string
+
+@description('Optional. Location for all resources.')
+param location string = resourceGroup().location
+
+@allowed([
+  'CanNotDelete'
+  'NotSpecified'
```
bmoore-msft, line 9: Instead of `NotSpecified` why not `none`?

modules/managed-identity/user-assigned-identity/main.bicep
bmoore-msft: I had a similar thought/reaction as @alex-frankel on this one - on why we're bundling things together like this (which seems the pattern in CARML now that I spend more time with it).  My first confusion was that I expected the roleAssignment to give the MSI access to something - that's the common pattern.  Create the MSI and then give it perms to do something.  Where this seems to be just giving people access to an MSI that doesn't have access to anything (since it was just created).  I "get" the pattern of -- create resources, let someone access them -- but for MSIs this seems less useful without being able to state what the MSI will do (i.e. have access to).

The other thought is whether we want to create new ways for users to do common/basic things...  IOW, do we want another/different way to create a resource and give access to it?  It is a tricky line, if you're familiar with ARM and you understand the simplification, it can be helpful but if you're new it will be confusing because every doc we have will talk about the native way of doing things, not this one.

Interesting (longer term) discussion of whether some of these scenarios (e.g. locks) should be simplified in the platform.

modules/managed-identity/user-assigned-identity/main.bicep
```diff
@@ -0,0 +1,57 @@
...
+  tags: tags
+}
+
+resource userMsi_lock 'Microsoft.Authorization/locks@2017-04-01' = if (lock != 'NotSpecified') {
+  name: '${userMsi.name}-${lock}-lock'
+  properties: {
+    level: lock
+    notes: (lock == 'CanNotDelete') ? 'Cannot delete resource or child resources.' : 'Cannot modify the resource or child resources.'
+  }
+  scope: userMsi
```
bmoore-msft, line 33: nit - scope should at the top

modules/managed-identity/user-assigned-identity/README.md
```diff
@@ -0,0 +1,68 @@
...
+| `lock`            | `string` | No       | Optional. Specify the type of lock.                                                                                                                                                                                                                                                                                                                                                                            |
+| `roleAssignments` | `array`  | No       | Optional. Array of role assignment objects that contain the 'roleDefinitionIdOrName' and 'principalId' to define RBAC role assignments on this resource. In the roleDefinitionIdOrName attribute, you can provide either the display name of the role definition, or its fully qualified ID in the following format: '/providers/Microsoft.Authorization/roleDefinitions/c2f4ef07-c644-48eb-af81-4b1b4947fb11' |
```
bmoore-msft, line 12: it probably makes sense to provide a code sample of the shape of the object for object params...  deciphering in a single line is harder than it should be...  I think we have:
```bicep
{
  principalId: 'foo'
  roleDefinitionIdOrName: 'bar'
}
```
but I also thought that principalId was (or could be) an array...

modules/managed-identity/user-assigned-identity/main.bicep
```diff
@@ -0,0 +1,57 @@
...
+}
+
+module userMsi_rbac '.bicep/nested_rbac.bicep' = [for (roleAssignment, index) in roleAssignments: {
+  name: '${uniqueString(deployment().name, location)}-UserMSI-Rbac-${index}'
+  params: {
+    description: contains(roleAssignment, 'description') ? roleAssignment.description : ''
+    principalIds: roleAssignment.principalIds
+    principalType: contains(roleAssignment, 'principalType') ? roleAssignment.principalType : ''
```
bmoore-msft, line 41: Is the user meant to supply principalType?  It's not in the docs/readme...

modules/managed-identity/user-assigned-identity/main.bicep
```diff
@@ -0,0 +1,57 @@
...
+    notes: (lock == 'CanNotDelete') ? 'Cannot delete resource or child resources.' : 'Cannot modify the resource or child resources.'
+  }
+  scope: userMsi
+}
+
+module userMsi_rbac '.bicep/nested_rbac.bicep' = [for (roleAssignment, index) in roleAssignments: {
+  name: '${uniqueString(deployment().name, location)}-UserMSI-Rbac-${index}'
+  params: {
+    description: contains(roleAssignment, 'description') ? roleAssignment.description : ''
+    principalIds: roleAssignment.principalIds
```
bmoore-msft, line 40: is principalIds plural? metadata description says singular..

modules/managed-identity/user-assigned-identity/.bicep/nested_rbac.bicep
```diff
@@ -0,0 +1,55 @@
+@sys.description('Required. The IDs of the prinicpals to assign to role to')
+param principalIds array
+
+@sys.description('Required. The name of the role to assign. If it cannot be found you can specify the role definition ID instead')
+param roleDefinitionIdOrName string
+
+@sys.description('Required. The resource ID of the resource to apply the role assignment to')
+param resourceId string
```
bmoore-msft, line 8: seems this should be `scope` since that's what it actually is in the API

modules/managed-identity/user-assigned-identity/README.md
bmoore-msft, line 12: re: roleDefinitionIdOrName - a couple thoughts
1) the `name` of a roleDef is the guid or the name property on the resource. the `roleName` is the name of the property we use for the "displayName"
2) when supplying the `id` is there any reason we can't just ask for the guid/name?  i.e. why do we need the full resourceId (which is harder to work with)... 

IOW, the parameter becomes `roleDefinitionName` - to which we can describe was the `roleName` or `name` of the resource?

modules/managed-identity/user-assigned-identity/.bicep/nested_rbac.bicep
```diff
@@ -0,0 +1,55 @@
...
+  'ForeignGroup'
+  'Device'
+  ''
+])
+param principalType string = ''
+
+@sys.description('Optional. Description of role assignment')
+param description string = ''
+
+var builtInRoleNames = {
```
bmoore-msft, line 24: one thing about providing this list (in every module) is that users won't know what's available in any given module, with searching the code.  If we follow this pattern, it feels like it has to work everywhere (in every roleAssignment module) or we should require the resource name and provide a lookup function in ARM.

cc: @alex-frankel

modules/managed-identity/user-assigned-identity/.bicep/nested_rbac.bicep
```diff
@@ -0,0 +1,55 @@
...
+  'Resource Policy Contributor': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '36243c78-bf99-498c-9df9-86d9f8d28608')
+  'User Access Administrator': subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '18d7d88d-d35e-4fb5-a5c3-7773c20a72d9')
+}
+
+resource userMsi 'Microsoft.ManagedIdentity/userAssignedIdentities@2018-11-30' existing = {
+  name: last(split(resourceId, '/'))
+}
+
+resource roleAssignment 'Microsoft.Authorization/roleAssignments@2021-04-01-preview' = [for principalId in principalIds: {
+  name: guid(userMsi.name, principalId, roleDefinitionIdOrName)
```
bmoore-msft, line 47: this will not be sufficiently unique in the tenant - since msiNames are not globally unique...

PR #105, opened 2022-04-27
UPDATE Deployment script module - Import ACR Images

Gordonby:
## Description

- Re-aligning the parameter surface area with that of #84 (This causes a major version bump as we're removing 3 parameters, and changing an array to a string).
- Adding in a retry loop used in #84 to mitigate errors such as RBAC delays
- Shortened the deployment name used to be just the last segment of the fully qualified image name, to give enough headroom from the length limit.

- [x] I have read the [Updating an existing module](../CONTRIBUTING.md#updating-an-existing-module) section in the contributing guide and updated the `version.json` file properly:
  - [x] The PR contains breaking changes, and I have bumped the MAJOR version in `version.json`.


alex-frankel (approved):
Great changes and great test of a breaking change!

PR #115, opened 2022-07-12
Deployment script module - Wait

### LINKED ISSUE #114: [Module Proposal]: Deployment delay

Yes

### Why is the module needed?

When deploying some Azure services together that use AzureAD for RBAC, it is common to receive errors regarding propagation time.

> "Caller is not authorized to perform action on resource. If role assignments, deny assignments or role definitions were changed recently, please observe propagation time."

An example is when Azure Application Gateway is created that uses Key Vault for its SSL Certificates. On some tenants, success rate is at about 20%.

### Module path

deployment-scripts/wait

### Describe the module

This new module would create a sleep for a configurable number of seconds to be used in other resources `DependsOn` in order to delay resource creation. EG. [wait.bicep](https://github.com/Azure/azure-quickstart-templates/pull/12524/files#diff-cbe0f293f3aec177575c790878a3fb776fa65342455911c10f346ffef798217e)

Gordonby:
## Description

This new module would create a sleep for a configurable number of seconds to be used in other resources DependsOn in order to delay resource creation. 

Closes #114 

## Adding a new module

### DIFF
--- a/modules/deployment-scripts/wait/README.md
+++ b/modules/deployment-scripts/wait/README.md
@@ -2,6 +2,10 @@
 
 A Deployment Script that introduces a delay to the deployment process.
 
+## Description
+
+{{ Add detailed description for the module. }}
+
 ## Parameters
 
 | Name          | Type     | Required | Description                             |
--- a/modules/deployment-scripts/wait/README.md
+++ b/modules/deployment-scripts/wait/README.md
@@ -1,0 +1,32 @@
+# Deployment Delay
+
+A Deployment Script that introduces a delay to the deployment process.
+
+## Parameters
+
+| Name          | Type     | Required | Description                             |
+| :------------ | :------: | :------: | :-------------------------------------- |
+| `waitSeconds` | `int`    | Yes      | The number of seconds to wait for       |
+| `location`    | `string` | No       | The location to deploy the resources to |
+
+## Outputs
+
+| Name | Type | Description |
+| :--- | :--: | :---------- |
+
+## Examples
+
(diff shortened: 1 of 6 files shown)


Gordonby:
@shenglol - i've got an interesting CI error.  Could you have a look?

shenglol:
@Gordonby Sure thing. The GitHub token has expired, and I updated it. The GitHub token was added when the repo was private so the CI can download files from pull requests. Since the repo is now public, I'll update the script to not use GitHub token so that we won't hit the same error again.

shenglol (approved):
:shipit:

shenglol:
Need to onboard MCR manifest. Will merge it after that's done.

modules/deployment-scripts/wait/README.md
```diff
@@ -0,0 +1,36 @@
+# Deployment Delay
+
+A Deployment Script that introduces a delay to the deployment process.
+
+## Description
+
+{{ Add detailed description for the module. }}
```
shenglol, line 7: @Gordonby sorry I missed this when I approve it. Can you add detailed description for the module? This is introduced in the latest version of the registry module tool.

shenglol (changes requested):
Please see https://github.com/Azure/bicep-registry-modules/pull/115/files#r924820892.

modules/deployment-scripts/wait/README.md
Gordonby: @shenglol  - done.

shenglol (approved):

PR #130, opened 2022-08-02
Create GitHub auto-action for 'brm generate'

dciborow:
## Description

This adds a GitHub action to auto generate the required files. This eliminates a developers need to open the project and manually run 'brm generate'.

### DIFF
--- a/.github/workflows/auto-fix.yml
+++ b/.github/workflows/auto-fix.yml
@@ -1,0 +1,84 @@
+name: BRM / Automaticly Triggered BRM Generation
+
+on:
+  push:
+    branches-ignore:
+      - main
+    paths: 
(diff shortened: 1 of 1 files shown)


dciborow:
@shenglol , @alex-frankel 

Would like to propose this GitHub action to automate running 'brm generate'. 

Here is an example commit generated when this step found that my files where not up to date.
image

https://github.com/dciborow/bicep-registry-modules/pull/1/commits/4aed441096eb0051a36619f2bea631aeab83aff3

shenglol:
Unfortunately, this is not going to work, since the job won't be able to push commits to forked repositories as it requires write access. We also prefer developers to run the generation step locally to make it easier to track commit history.

PR #200, opened 2022-11-15
Add LZ Vending Module

### LINKED ISSUE #191: Bicep Landing Zone (Subscription) Vending module

Yes

### Why is the module needed?

Customers today have a gap in being able to create subscriptions programmatically with the base networking resources they generally require in a repeatable fashion. 

We in the CAE team have created a Bicep Landing Zone vending module (https://github.com/Azure/bicep-lz-vending/), as well as a Terraform equivalent (HTTPS://aka.ms/LZ-vending/TF), to address this gap. 

We would like to have it consumable in the same way in bicep as it is via terraform in a public module registry to remove an additional step for consumers to have to worry about.  

### Module path

lz/subWithNetworking

### Describe the module

All details can be found in repo here https://github.com/Azure/bicep-lz-vending

jtracey93:
## Description

Closes #191 

## Adding a new module

- [x] A proposal has been submitted and approved. - see #191 


modules/lz/sub-vending/main.bicep
```diff
@@ -0,0 +1,479 @@
...
+
+// PARAMETERS
+
+// Subscription Parameters
+@metadata({
+  example: true
+})
+@sys.description('''Whether to create a new Subscription using the Subscription Alias resource. If `false`, supply an existing Subscription's ID in the parameter named `existingSubscriptionId` instead to deploy resources to an existing Subscription.
+
+- Type: Boolean
```
alex-frankel, line 21: I think we should try to get rid of anything that is only here for PSDocs. Thoughts?

jtracey93: Yup I'm mulling this over hence leaving this PR in draft for now until I finish everything and make a final call. 👍

I have considered migrating to brm to generate the module docs in modules home repo. But I actually really like the per parameter verbose output and the way it creates the parameter file example based on parameter metadata in PSdocs. 

But I think for now I'll just have to maintain 2 copies of the module docs and make some manual changes if we need to cut a release. Mainly because the tests on this repo run a brm and if there it is a diff it will fail. So that pretty much means I have to move to brm generated docs.

I'm happy to go this way for now and happy to participate, provide feedback on changes and assist wherever possible. 

Happy to chat as always 👍👍

jtracey93: have moved to brm docs here :)

jtracey93:
@alex-frankel this is good to go now 🙂

shenglol (approved):
:shipit:

PR #202, opened 2022-11-17
feat(new): Automation Account Runbook

### LINKED ISSUE #199: [Module Proposal]: Automation Runbook

Yes

### Why is the module needed?

Creating automation runbooks with schedules is a laborious task.
There are several resource abstracts to create, the schedule can be particularly onerous with the timezone string property.

This module would streamline the creation process.

The first use case is that of stopping and starting an AKS cluster at the beginning and end of the day. This is a feature that will be added to [AKSC](https://github.com/Azure/Aks-Construction), a bicep module in the registry would increase the reusability of the capability.

### Module path

automation/automationrunbook

### Describe the module

- An Automation Account (with system managed identity)
- A variety of schedules with configurable time zone via allowed param array
- A single runbook (via URL)
- The job, runbook->schedule mapping
- Will take an existing Log analytics workspace id for logging.

When the runbook params are null it will just create the account and scheduled.

The readme will include runbook examples of starting/stopping AKS as well as the most popular runbook from the gallery; stopping/Starting vms.

Gordonby:
## Description

This contribution Closes #199 

`This module does use experimental features: userDefinedTypes`

## Adding a new module


Gordonby:
I've added UserDefinedTypes to this bicep module @alex-frankel - which really helps with the flexibility/maturity of the use case.
However we should hold off on this PR progressing until after the next bicep release.

alex-frankel:
I don't think we should use user-defined types in a public modules until the feature is enabled by default. Otherwise, this module will only work if the preview flag is enabled. Thoughts?

Of course, it's great that you are trying the feature out and glad it is improving usability :)

Gordonby:
> I don't think we should use user-defined types in a public modules until the feature is enabled by default. Otherwise, this module will only work if the preview flag is enabled. Thoughts?

Agreed, public modules should be easy to adopt 👍

Gordonby:
@microsoft-github-policy-service agree

Gordonby:
@dciborow - As we're taking module that use Custom Types now (eg. The CosmosDb one). I would consider this module ready for review/merge.

Gordonby:
@dciborow Have you had a chance to review?

skydiver1980 (approved):

ChrisSidebotham:
Closing as stale

PR #204, opened 2022-11-18
LZ Vending - bumping examples version tag

jtracey93:
LZ Vending - bumping examples version tag as was set to `1.1.0` but when published it was `1.1.1`.

**No changes to bicep code, just docs**

@shenglol @alex-frankel can we get this one approved and merged ASAP so the examples are correct 👍

### DIFF
--- a/modules/lz/sub-vending/README.md
+++ b/modules/lz/sub-vending/README.md
@@ -76,7 +76,7 @@ We would like feedback on what's missing in the module. Please raise an [issue](
 ```bicep
 targetScope = 'managementGroup'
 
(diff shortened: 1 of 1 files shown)


Gordonby (approved):

jtracey93:
@alex-frankel @shenglol unless there is a way to add a version for `v1.1.0` as this then matches the release in the source repo https://github.com/Azure/bicep-lz-vending/releases/tag/V1.1.0 ?

This would nice for consumers IMO

alex-frankel (approved):

PR #232, opened 2023-02-01
Generate modules metadata

bhsubra:
Changes in the PR include:
1. Generate modules metadata that'll be used in completion scenarios
2. I'll have a follow up PR to write the generated data to storage account container

Verified the changes in dogfood via:
https://github.com/bhsubra/bicep-registry-modules/pull/1

### DIFF
--- a/.github/workflows/get-modules-metadata.yml
+++ b/.github/workflows/get-modules-metadata.yml
@@ -1,0 +1,43 @@
+name: Get modules metadata
+on:
+  pull_request:
+    branches:
+      - main
+  workflow_call:
+    secrets:
+      # Secrets must be passed from the caller workflow explicitly.
+      CLIENT_ID:
+        required: true
+      TENANT_ID:
+        required: true
+      SUBSCRIPTION_ID:
+        required: true
+  workflow_dispatch:
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Install packages
+        run: npm ci
+
+      - name: Get modules metadata
+        uses: actions/github-script@v6
+        with:
+          result-encoding: string
+          script: |
+            const script = require("./scripts/github-actions/get-modules-metadata.js")
+            await script({ require, core })
+
+      - name: Upload modules metadata
+        uses: actions/upload-artifact@v3
+        with:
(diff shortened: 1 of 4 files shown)


scripts/github-actions/get-modules-metadata.js
```diff
@@ -0,0 +1,40 @@
+async function getModulesMetadata({ require, core }) {
```
shenglol, line 1: Mind adding a [JSDoc ](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html) comment to annotate the types of parameters?

scripts/github-actions/get-sub-directory-names.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * @param {typeof import("fs")} fs
+ * @param {string} dir
+ */
+function getSubdirNames(dir) {
+  const fs = require("fs");
```
shenglol, line 6: Should `fs` remain a parameter and be passed from the caller?

scripts/github-actions/get-modules-metadata.js
bhsubra: Sure, updated here:
https://github.com/Azure/bicep-registry-modules/commit/20d412d9da3b54faf759004a8783918aee430fe4

Thanks!

scripts/github-actions/get-sub-directory-names.js
bhsubra: Updated here:
https://github.com/Azure/bicep-registry-modules/commit/20d412d9da3b54faf759004a8783918aee430fe4

Thanks!

.github/workflows/get-modules-metadata.yml
```diff
@@ -0,0 +1,40 @@
...
+        required: true
+      TENANT_ID:
+        required: true
+      SUBSCRIPTION_ID:
+        required: true
+  workflow_dispatch:
+
+permissions:
+  id-token: write
+  contents: read
```
shenglol, line 16: Are the secrets needed? If not, we should remove them.

bhsubra: Yes, I'll be sending out a follow up PR to login to azure and write the modules metadata to storage account container. Let me know if you want me to remove it here and include it in other PR.

shenglol: No, in that case we can leave them as is.

shenglol (approved):

PR #233, opened 2023-02-01
Write modules metadata to storage account container

bhsubra:
Write modules metadata to storage account container. This will be used in completion scenarios.


shenglol (approved):
:shipit:

majastrz (approved):
:shipit:

PR #238, opened 2023-02-03
Fix upload module metadata workflow

bhsubra:
Fixes upload module metadata workflow

### DIFF
--- a/scripts/github-actions/upload-modules-metadata.js
+++ b/scripts/github-actions/upload-modules-metadata.js
@@ -1,6 +1,10 @@
-/**
- * @param {typeof import("@actions/core")} core
- */
+ /** 
+  * @typedef Params 
+  * @property {typeof require} require 
+  * @property {typeof import("@actions/core")} core 
+  * 
+  * @param {Params} params 
+  */ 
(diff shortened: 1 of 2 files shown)


scripts/github-actions/upload-modules-metadata.js
```diff
@@ -1,7 +1,7 @@
 /**
  * @param {typeof import("@actions/core")} core
  */
-async function uploadModulesMetadata(core) {
+async function uploadModulesMetadata({ require, core }) {
```
shenglol, line 4: Can you also update the JSDoc comment? Here's an example: https://github.com/Azure/bicep-registry-modules/blob/917735e53c4de2b6b995a1bf527eb51f5c96fc5a/scripts/github-actions/get-changed-module.js#L1-L10

shenglol (approved):
Looks good to me. Only one comment about the JSDoc.

scripts/github-actions/upload-modules-metadata.js
bhsubra: Updated, thanks!

PR #252, opened 2023-03-03
feat: new Azure Key Vault Module

### LINKED ISSUE #244: [Module Proposal]: Azure KeyVault

Yes

### Why is the module needed?

We require a base module for managing Azure KeyVault and one is not currently available in the public registry. It will follow the concept of the Azure Container Registry module that I previously created via #166.

### Module path

security/keyvault

### Describe the module

This module will allow for provisioning and management of Azure KeyVault, in line with a similar experience as using Azure Portal/CLI, for example: support for integrating in VNETs via Private Endpoint.

If there is a recommendation for a more relevant module path, please let me know.

dciborow:
## Description

Closes #244

If you haven't already, read the full [contribution guide](../CONTRIBUTING.md). The guide may have changed since the last time you read it, so please double-check. Once you are done and ready to submit your PR, edit the PR description and run through the relevant checklist below.

Enable GitHub Worksflows in your fork to enable auto-generation of assets with our [GitHub Action](/.github/workflows/push-auto-generate.yml).
To trigger GitHub Actions after auto-generation, [add a GitHub PAT](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token) as a secret in your forked repository called `PAT`.

## Adding a new module


dciborow:
@shenglol , this one is also ready for review

dciborow:
Closes #244

modules/security/keyvault/main.bicep
```diff
@@ -0,0 +1,84 @@
...
+
+@description('List of RBAC policies to assign to the Key Vault')
+param rbacPolicies array = []
+
+@allowed([ 'new', 'existing'])
+@description('Specifies whether to create a new Key Vault or use an existing one. Use "new" to create a new Key Vault or "existing" to use an existing one.')
+param newOrExisting string = 'new'
+
+@description('Enable role assignment for the Key Vault')
+param assignRole bool = true
```
wsmelton, line 24: Why not offer customization of what role is used? This locks it down where I'd have to run it again to assign Key access.

dciborow: had to break things out into a second module for the double for loop, but was able to get this done.

modules/security/keyvault/modules/secrets.bicep
alex-frankel, line 1: I do wonder if creation of secrets should be its own module, but don't feel strongly enough to require a change. Just calling it out as something I took note of.

modules/security/keyvault/main.bicep
```diff
@@ -0,0 +1,98 @@
...
+  '4633458b-17de-408a-b874-0445c86b69e6' // rbacSecretsReaderRole
+  'a4417e6f-fecd-4de8-b567-7b0420556985' // rbacCertificateOfficerRole
+]
+
+@allowed([ 'new', 'existing' ])
+@description('Whether to create a new Key Vault or use an existing one')
+param newOrExisting string = 'new'
+
+@description('List of secrets to create in the Key Vault [ { secretName: string, secretValue: string }]')
+param secrets array = []
```
alex-frankel, line 39: Isn't this going to expose secrets?

alex-frankel (dismissed):

modules/security/keyvault/main.bicep
wsmelton: The only way to keep the value secure would be having it as a separate module and require secure string to be provided. So users consuming could call it per secret.

The secure object I found from research (we had internal request to use this for similar reasons) will keep the key-value pair secure until something requires that the key or the value be converted to plaintext, at that that point the whole object is converted to plaintext.

modules/security/keyvault/README.md
```diff
@@ -0,0 +1,75 @@
...
+module keyVault 'br/public:security/keyvault:0.0.1' = {
+  name: 'myKeyVault'
+  params: {
+    location: location
+    name: name
+  }
+}
+```
+
+### Example 2
```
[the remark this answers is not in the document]
dciborow: added example 3 and 4.

modules/security/keyvault/modules/secrets.bicep
dciborow: A dedicated module for secrets would look something like this.

```bicep
param keyVaultName string
param secretName string

@secure
param secretValue string

resource vault 'Microsoft.KeyVault/vaults@2022-11-01' existing = {
  name: keyVaultName
}

resource secret 'Microsoft.KeyVault/vaults/secrets@2022-11-01' = {
  parent: vault
  name: secretName
  properties: {
    value: secretValue
  }
}
```

And it use it with this module would be this.

```bicep
param location
param subscriptionId string = subscription().subscriptionId
param resourceGroupName string = resourceGroup().name
param keyVaultName string
param secretName string

@secure
param secretValue string

module keyVault 'br/public:security/keyvault:0.0.1' = {
  name: 'myKeyVault'
  params: {
    location: location
    name: keyVaultName
    subscriptionId: subscriptionId
    resourceGroupName: resourceGroupName
  }
}

module secret 'br/public:security/keyvault-secret:0.0.1' = {
  name: guid(keyVault.name, 'secrets')
  scope: resourceGroup(subscriptionId, resourceGroupName)
  params: {
    keyVaultName: keyVault.outputs.name
    secretName : secretName 
    secretValue: secretValue
  }
}
```

If we put it together

```bicep
param location
param subscriptionId string = subscription().subscriptionId
param resourceGroupName string = resourceGroup().name
param keyVaultName string
param secretName string

@secure
param secretValue string

module keyVault 'br/public:security/keyvault:0.0.1' = {
  name: 'myKeyVault'
  params: {
    location: location
    name: keyVaultName
    subscriptionId: subscriptionId
    resourceGroupName: resourceGroupName
    secrets: [{secretName: secretName, secretValue: secretValue }]
  }
}
```

dciborow: I like how we can turn everything into parameters for the users and hide all the scope configurations in the module.

modules/security/keyvault/main.bicep
dciborow: im rolling back the complex object here. Let me see how far I can get only taking in on secret at a time.

modules/security/keyvault/main.bicep
```diff
@@ -0,0 +1,100 @@
...
+@allowed(['A', 'B'])
+@description('The SKU family of the Key Vault.')
+param skuFamily string = 'A'
+
+@description('Specifies whether RBAC authorization should be enabled for the Key Vault.')
+param enableRbacAuthorization bool = true
+
+module keyVault 'modules/vaults.bicep' = {
+  name: guid(name, 'deploy')
+  scope: resourceGroup(subscriptionId, resourceGroupName)
```
wsmelton, line 61: Including this in Bicep can cause conflicts based on how the module is consumed. I do not see other modules using this and being that a KV only resides in a given Resource Group I'm not sure there is a need to have scope defined.

modules/security/keyvault/main.bicep
```diff
@@ -0,0 +1,100 @@
+@description('Deployment Location')
+param location string
+
+@description('Prefix of Cosmos DB Resource Name')
+param prefix string = 'kv'
+
+@description('Name of the Key Vault')
+param name string = '${prefix}-${uniqueString(resourceGroup().id)}'
```
wsmelton, line 8:
```suggestion
@minLength(3)
@maxLength(24)
param name string = '${prefix}-${uniqueString(resourceGroup().id)}'
```

modules/security/keyvault/main.bicep
```diff
@@ -0,0 +1,100 @@
...
+@allowed(['standard', 'premium'])
+@description('The SKU name of the Key Vault.')
+param skuName string = 'standard'
+
+@allowed(['A', 'B'])
+@description('The SKU family of the Key Vault.')
+param skuFamily string = 'A'
+
+@description('Specifies whether RBAC authorization should be enabled for the Key Vault.')
+param enableRbacAuthorization bool = true
```
wsmelton, line 57: Is this really needed since there is no option to configure the KV with Access Policies?

modules/security/keyvault/main.bicep
```diff
@@ -0,0 +1,100 @@
...
+param subscriptionId string = subscription().subscriptionId
+
+@description('For an existing Managed Identity, the Resource Group it is located in')
+param resourceGroupName string = resourceGroup().name
+
+@description('Subnet ID for the Key Vault')
+param subnetID string = ''
+
+@description('List of RBAC policies to assign to the Key Vault')
+param rbacPolicies array = []
```
wsmelton, line 23: This may be a bit confusing to some considering RBAC assignment is not policy drive in KVs. It is just assigning an RBAC role. This array is being passed in to the nested module for role assignment and requires the Principal IDs that are to be assigned the given roles.

However, if I pass in multiple Principal IDs that means they are each going to be granted the default roleAssignments.

I would think it may be easier to create this as an array of objects that defined the Principal ID and the Role together. Example:

```
var rbacAssignment = [
    {
        principalId: '0000-000-000-00000-0000'
        role: 'Key Vault Secrest Officer'
    }
    {
        principalId: '00001-000-000-000-000000'
        role: 'Key Vault Secrets User'
    }
]
```

This lets them have more granular control over the RBAC assignments.

modules/security/keyvault/main.bicep
```diff
@@ -0,0 +1,100 @@
...
+  name: guid(keyVault.name, rbacRole)
+  scope: resourceGroup(subscriptionId, resourceGroupName)
+  params: {
+    keyVaultName: keyVault.name
+    rbacPolicies: rbacPolicies
+    rbacRole: rbacRole
+  }
+}]
+
+module secret 'modules/secrets.bicep' = if (secretName != '' && secretValue != '') {
```
wsmelton, line 86: The default value on `secretValue` should be set to the value you are checking for. Right now it auto-generates a secret name so this would never return true would it?

modules/security/keyvault/main.bicep
```diff
@@ -0,0 +1,100 @@
+@description('Deployment Location')
+param location string
+
+@description('Prefix of Cosmos DB Resource Name')
```
wsmelton, line 4:
```suggestion
@description('Prefix of Azure Key Vault Resource Name')
```

modules/security/keyvault/main.bicep
```diff
@@ -0,0 +1,100 @@
+@description('Deployment Location')
+param location string
```
wsmelton, line 2: I see a mix of some modules defaulting this value to the `resourceGroup().location` and some that don't. It would be nice if this was more consistent or set as a standard in the contributing guide.

modules/security/keyvault/modules/secrets.bicep
```diff
@@ -0,0 +1,19 @@
...
+  name: keyVaultName
+}
+
+resource secret 'Microsoft.KeyVault/vaults/secrets@2022-11-01' = {
+  parent: vault
+  name: secretName
+  properties: {
+    value: secretValue
+  }
+}
```
wsmelton, line 19: Would it be worth having an output for the secret URI in the event the reference needed to be used downstream in a user's deployment?

modules/security/keyvault/main.bicep
```diff
@@ -0,0 +1,100 @@
...
+@description('The SKU family of the Key Vault.')
+param skuFamily string = 'A'
+
+@description('Specifies whether RBAC authorization should be enabled for the Key Vault.')
+param enableRbacAuthorization bool = true
+
+module keyVault 'modules/vaults.bicep' = {
+  name: guid(name, 'deploy')
+  scope: resourceGroup(subscriptionId, resourceGroupName)
+  params: {
```
wsmelton, line 62: The name of the Key Vault is not being passed to the nested module.

modules/security/keyvault/main.bicep
dciborow: I think this works? But let's review it.

We start with...
```bicep
param secretName string = 'secret${uniqueString(resourceGroup().id)}'
@secure()
param secretValue string = ''
```

So we have...
```bicep
if (secretName != '' && secretValue != '')
if (secret${uniqueString(resourceGroup().id)} != '' && '' != '')
if ( true && false )
if ( false ) // Results in section being skipped
```

Now if the user provides a value for secretValue, and leaves the default for secretName...
```bicep
if (secretName != '' && secretValue != '')
if (secret${uniqueString(resourceGroup().id)} != '' && 'Something' != '')
if ( true && true )
if ( true ) // Results in section being run
```

Now, if the user provides a secretValue, but mistakenly provides an empty string for secretName....
```bicep
if (secretName != '' && secretValue != '')
if ('' != '' && 'Something' != '')
if ( false && true )
if ( false ) // Results in section being skipped
```


I believe this works like I'm expecting?

modules/security/keyvault/main.bicep
dciborow: This one is tricky, and a weird edge case. It looks good to have the default as 'resourceGroup().location', but this resulted in issues when I tried to consume our modules in my Azure Marketplace Applications.

There are 3 linter rules that we have to comply with for the Marketplace.
https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/linter-rule-no-hardcoded-location
https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/linter-rule-no-loc-expr-outside-params
https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/linter-rule-explicit-values-for-loc-params

Let me chat around and get some more input, and see if everyone is aligned with `param location string` or `param location string = resourceGroup().location`, in which case we need to get the Marketplace team to adjust their arm-ttk checks.

dciborow:
@wsmelton , thanks for the feedback. I will be OoO next week, but am going to try and resolve as much of this as I can before that.

modules/security/keyvault/main.bicep
dciborow: fixed.

modules/security/keyvault/main.bicep
dciborow: I poked this line, and moved the conditions into their own variable so I can more easily reuse the flag. But we should still be aligned that this works like im expecting.

wsmelton: This is just me but I don't like modules auto generating names that get used in other automation workflows.

I would set the default value to an empty string to make the user provide it.

modules/security/keyvault/main.bicep
dciborow: The array you proposed makes sense. Let me think about the best way to bring this in. Custom types would make this better, but I think I can do a good job without.

modules/security/keyvault/main.bicep
dciborow: I was pushing this module a little further than the rest and was aiming to make it as generic and universal as possible. This is the first module I have gone this far with, so good spot here for some extra feedback and review.

My goal here was that this module should be able to create a new key vault in any sub/rg, or use an existing one, and then add a secret to it.

Without this part, I can add a secret to an existing key vault in a different RG. But if the user provides me with a different sub/resource group, and also uses the "new" flag, even if they have permission to perform the action, it will still fail.

dciborow: @wsmelton , do you have an example in mind that would cause this to fail? I can try and generate a test case for us that covers this so we can make sure the module addresses the concern.

modules/security/keyvault/main.bicep
dciborow: > This is just me but I don't like modules auto generating names that get used in other automation workflows.
> 
> I would set the default value to an empty string to make the user provide it.

Since the secretName is only required when the secretValue is provided, I'm worried about if a user provides 'secretValue' but forgets to also provide 'secretName'.

Maybe I can start off just addressing this in the descriptions though, and we can see if it is a real problem or not for users.

dciborow: I changed the default value of the secretName to just ''. And I removed `secretName != ''` from the check. This way, if a user provides a secret, but forgets the name, the deployment should fail with something like invalid secret name, instead of succeeding without setting the secret.

modules/security/keyvault/main.bicep
wsmelton: To deploy Bicep you have to connect to a specific subscription. Why do you need to support cross deployments?

Anyone using this to create or update secrets will be in the same subscription and resource group of the Key Vault. There is no reason for them not to be the way the deployment command works.

wsmelton: The general pattern promoted by most is Key Vault for each application or at most Resource Group. I'm not sure anyone uses a central or shared KV because you only control access at the vault level.

dciborow: rolled back the rg/sub parameters and will think about this again when I have the use case in front of me.

modules/security/keyvault/main.bicep
dciborow: So, if you have similar to that, here is how you could use the module.

I think I am going to keep the module how it is, because I am having trouble cleaning implementing the conditional loops with the more complex objects.

I will include this as an example

```bicep
module keyVault 'br/public:security/keyvault:0.0.1' = {
  name: 'myKeyVault'
  params: {
    location: location
    name: name
  }
}

var rbacAssignments = [
    {
        principalIds: [ '0000-000-000-00000-0000' ]
        roles: [ 'Key Vault Secrest Officer' ]
    }
    {
        principalId: [ '00001-000-000-000-000000' ]
        roles: [ 'Key Vault Secrets Officer', 'Key Vault Secrets User' ]
    }
]

module rbacAssignments 'br/public:security/keyvault:0.0.1' = [ rbacAssignment in rbacAssignments: {
  name: 'rbac-assignment-guid(rbacAssignment.principalId, rbacAssignment.role)'
  params: {
    location: location
    name: keyVault.outputs.name
    newOrExisting: 'existing'
    rbacPolicies: rbacAssignment.principalIds
    roleAssignments: rbacAssignment.roles
  }
}]
```

modules/security/keyvault/main.bicep
dciborow: pulled it on from parameters

modules/security/keyvault/main.bicep
```diff
@@ -0,0 +1,107 @@
...
+param name string = take('${prefix}-${uniqueString(resourceGroup().id)}', 24)
+
+@description('The tenant ID where the Key Vault is deployed')
+param tenantId string = subscription().tenantId
+
+@description('Deploy Key Vault into existing Virtual Network. Enabling this setting also requires subnetID')
+param enableVNet bool = false
+
+@description('Subnet ID for the Key Vault')
+param subnetID string = enableVNet ? null ? ''
```
dciborow, line 19:
```suggestion
param subnetID string = enableVNet ? null : ''
```

modules/security/keyvault/main.bicep
```diff
@@ -0,0 +1,107 @@
...
+    tenantId: tenantId
+    subnetID: subnetID
+    enableVNet: enableVNet
+  }
+}
+
+module rbacRoleAssignments 'modules/roleAssignment.bicep' = [for rbacRole in roleAssignments: {
+  name: guid(keyVault.name, rbacRole)
+  params: {
+    keyVaultName: keyVault.name
```
dciborow, line 74:
```suggestion
    keyVaultName: keyVault.name
```
```suggestion
    keyVaultName: keyVault.name
```
```suggestion
    keyVaultName: keyVault.outputs.name
```

modules/security/keyvault/main.bicep
```diff
@@ -0,0 +1,107 @@
...
+    rbacRole: rbacRole
+  }
+}]
+
+var createSecret = secretValue != ''
+
+module secret 'modules/secrets.bicep' = if (createSecret) {
+  name: guid(keyVault.name, 'secrets')
+  params: {
+    keyVaultName: keyVault.name
```
dciborow, line 85:
```suggestion
    keyVaultName: keyVault.outputs.name
```

modules/security/keyvault/main.bicep
```diff
@@ -0,0 +1,107 @@
...
+    skuFamily: skuFamily
+    skuName: skuName
+    tenantId: tenantId
+    subnetID: subnetID
+    enableVNet: enableVNet
+  }
+}
+
+module rbacRoleAssignments 'modules/roleAssignment.bicep' = [for rbacRole in roleAssignments: {
+  name: guid(keyVault.name, rbacRole)
```
dciborow, line 72:
```suggestion
  name: guid(name, rbacRole)
```

modules/security/keyvault/main.bicep
```diff
@@ -0,0 +1,107 @@
...
+    keyVaultName: keyVault.outputs.name
+    rbacPolicies: rbacPolicies
+    rbacRole: rbacRole
+  }
+}]
+
+var createSecret = secretValue != ''
+
+module secret 'modules/secrets.bicep' = if (createSecret) {
+  name: guid(keyVault.name, 'secrets')
```
dciborow, line 83:
```suggestion
  name: guid(name, 'secrets')
```

modules/security/keyvault/main.bicep
```diff
@@ -0,0 +1,107 @@
...
+    secretName: secretName
+    secretValue: secretValue
+  }
+}
+
+@description('Key Vault Id')
+output id string = keyVault.outputs.id
+
+@description('Key Vault Name')
+output name string = keyVault.outputs.name
```
dciborow, line 95:
```suggestion
output name string = name
```

modules/security/keyvault/main.bicep
```diff
@@ -0,0 +1,107 @@
...
+output id string = keyVault.outputs.id
+
+@description('Key Vault Name')
+output name string = name
+
+@description('Key Vault Id')
+output id string = createSecret ? secret.outputs.id : ''
+
+@description('Key Vault Name')
+output secretName string = createSecret ? secret.outputs.name : ''
```
dciborow, line 101:
```suggestion
output secretName string = secretName
```

modules/security/keyvault/main.bicep
```diff
@@ -0,0 +1,107 @@
...
+param name string = take('${prefix}-${uniqueString(resourceGroup().id)}', 24)
+
+@description('The tenant ID where the Key Vault is deployed')
+param tenantId string = subscription().tenantId
+
+@description('Deploy Key Vault into existing Virtual Network. Enabling this setting also requires subnetID')
+param enableVNet bool = false
+
+@description('Subnet ID for the Key Vault')
+param subnetID string = enableVNet ? null : ''
```
dciborow, line 19:
```suggestion
param subnetID string = ''
```

modules/security/keyvault/version.json
```diff
@@ -0,0 +1,8 @@
+{
+  "$schema": "https://aka.ms/bicep-registry-module-version-file-schema#",
+  "version": "0.0",
```
shenglol, line 3: Do we want to start with version `0.0.1`?

dciborow:
```suggestion
  "version": "0.0.1",
```

modules/security/keyvault/version.json
```diff
@@ -0,0 +1,8 @@
+{
+  "$schema": "https://aka.ms/bicep-registry-module-version-file-schema#",
+  "version": "0.0.1",
```
dciborow, line 3:
```suggestion
  "version": "1.0",
```

shenglol (approved):

PR #276, opened 2023-03-21
feat(new): Unreal Cloud DDC

dciborow:
## Description
Closes #277 

This PR features nested templates in the `modules` directory which we are working to replace but are still pending the following PRs. The versions of those templates in this PR made be out of date and not reflect the latest feedback. As the following PRs are approved, we will update this template with the latest changes.

https://github.com/Azure/bicep-registry-modules/pull/254
https://github.com/Azure/bicep-registry-modules/pull/251
https://github.com/Azure/bicep-registry-modules/pull/252
https://github.com/Azure/bicep-registry-modules/pull/253

- [ ] #251 
- [ ] #252
- [ ] #253 
- [ ] #254 

## Adding a new module

- [ ] A proposal has been submitted and approved.
- [x] I have included "Closes #{module_proposal_issue_number}" in the PR description.
- [x] I have run `brm validate` locally to verify the module files.
- [x] I have run deployment tests locally to ensure the module is deployable.

PR #278, opened 2023-03-21
Update AKS Run Command

dciborow:
## Description

The following changes are required so that this module can be used with a Marketplace application within our new Unreal Cloud DDC module - https://github.com/Azure/bicep-registry-modules/pull/276

This introduces a **BREAKING_CHANGE** by requiring the location parameter be set, as required by Azure Marketplace per https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/linter-rule-explicit-values-for-loc-params

Aligning newOrExisting with the createUI per https://learn.microsoft.com/en-us/azure/azure-resource-manager/managed-applications/microsoft-common-serviceprincipalselector

- [x] I have read the [Updating an existing module](https://github.com/Azure/bicep-registry-modules/blob/main/CONTRIBUTING.md#updating-an-existing-module) section in the contributing guide and updated the `version.json` file properly:
  - [x] The PR contains breaking changes, and I have bumped the MAJOR version in `version.json`.
- [x] I have updated the examples in README with the latest module version number.


modules/deployment-scripts/aks-run-command/main.bicep
```diff
@@ -31,35 +32,39 @@ param commands array
...
-  'OnSuccess'
-  'OnExpiration'
-  'Always'
-])
+@allowed([ 'OnSuccess', 'OnExpiration', 'Always' ])
 @description('When the script resource is cleaned up')
 param cleanupPreference string = 'OnSuccess'
 
-resource aks 'Microsoft.ContainerService/managedClusters@2022-09-01' existing = {
+@description('Set to true when deploying template across tenants')
```
Gordonby, line 39: Can you expand on this scenario description further?

Gordonby (approved):
Looks like a good update.

modules/deployment-scripts/aks-run-command/main.bicep
dciborow: When going cross-tenant, within a Marketplace application, without this the Marketplace's SP tries to run the custom script step and we can't RBAC failures.

We have tried without but have found that it is needed.

PR #290, opened 2023-03-24
docs: Clarifying that Bicep samples replace the usual .json parameter file

gbeaud:
## Description

- Clarifying the fact that the Bicep code samples "act as parameter file" 
- Clarifying that if the user wants to use the .json parameters file, then they need to declare the parameters at the top of the Bicep main file

### DIFF
--- a/modules/lz/sub-vending/README.md
+++ b/modules/lz/sub-vending/README.md
@@ -199,7 +199,7 @@ module sub003 'br/public:lz/sub-vending:1.2.2' = {
 
 ### Example JSON Parameter File
 
-These parameters would need to be declared in the Bicep file using them (currently not present in the above provided Bicep samples). Note that it is not necessary to use JSON files here since in the proposed design pattern, the Bicep file would be the source of parameters.
+These parameters would need to be declared in the Bicep file using them (currently not present in the above provided Bicep samples). Note that it is not necessary to use JSON parameter files here since in the proposed design pattern, the Bicep file would be the source of parameters.
 
 ```json
 {
--- a/modules/lz/sub-vending/README.md
+++ b/modules/lz/sub-vending/README.md
@@ -10,7 +10,7 @@ The landing zone Bicep modules are designed to accelerate deployment of the indi
 
 > See the different types of landing zones in the Azure Landing Zones documentation here: [What is an Azure landing zone? - Platform vs. application landing zones](https://learn.microsoft.com/azure/cloud-adoption-framework/ready/landing-zone/#platform-vs-application-landing-zones)
 
-The modules are designed to be instantiated many times, once for each desired landing zone. Since the entire logic is captured in the registry's module, you may use the below proposed Bicep code samples to be the source of parameters instead of the usual parameters.json file, and have in this case one Bicep file per landing zone. If you decide instead to use a parameters.json (as the one proposed below), you will need to declare parameters in the Bicep file so it can fetch them from the JSON file.
+The modules are designed to be instantiated many times, once for each desired landing zone. Since the logic is captured in the registry's module, you may use the below proposed Bicep code samples to be the source of parameters instead of the usual `parameters.json` file, and have in this case one Bicep file per landing zone. If you decide instead to use a `parameters.json` (as the one proposed below), you will need to declare parameters in the Bicep file so it can accept them as parameter inputs from the parameters JSON file.
 
 This is currently split logically into the following capabilities:
 
--- a/modules/lz/sub-vending/README.md
+++ b/modules/lz/sub-vending/README.md
@@ -10,7 +10,7 @@ The landing zone Bicep modules are designed to accelerate deployment of the indi
 
 > See the different types of landing zones in the Azure Landing Zones documentation here: [What is an Azure landing zone? - Platform vs. application landing zones](https://learn.microsoft.com/azure/cloud-adoption-framework/ready/landing-zone/#platform-vs-application-landing-zones)
 
-The modules are designed to be instantiated many times, once for each desired landing zone.
+The modules are designed to be instantiated many times, once for each desired landing zone. Since the entire logic is captured in the registry's module, you may use the below proposed Bicep code samples to be the source of parameters instead of the usual parameters.json file, and have in this case one Bicep file per landing zone. If you decide instead to use a parameters.json (as the one proposed below), you will need to declare parameters in the Bicep file so it can fetch them from the JSON file.
 
 This is currently split logically into the following capabilities:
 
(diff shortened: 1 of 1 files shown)


modules/lz/sub-vending/README.md
```diff
@@ -199,6 +199,8 @@ module sub003 'br/public:lz/sub-vending:1.2.2' = {
 
 ### Example JSON Parameter File
 
+These parameters would need to be declared in the Bicep file using them (currently not present in the above provided Bicep samples). Note that it is not necessary to use JSON files here since in the proposed design pattern, the Bicep file would be the source of parameters.
```
jtracey93, line 202:
```suggestion
These parameters would need to be declared in the Bicep file using them (currently not present in the above provided Bicep samples). Note that it is not necessary to use JSON parameter files here since in the proposed design pattern, the Bicep file would be the source of parameters.
```

modules/lz/sub-vending/README.md
```diff
@@ -10,7 +10,7 @@ The landing zone Bicep modules are designed to accelerate deployment of the indi
 
 > See the different types of landing zones in the Azure Landing Zones documentation here: [What is an Azure landing zone? - Platform vs. application landing zones](https://learn.microsoft.com/azure/cloud-adoption-framework/ready/landing-zone/#platform-vs-application-landing-zones)
 
-The modules are designed to be instantiated many times, once for each desired landing zone.
+The modules are designed to be instantiated many times, once for each desired landing zone. Since the entire logic is captured in the registry's module, you may use the below proposed Bicep code samples to be the source of parameters instead of the usual parameters.json file, and have in this case one Bicep file per landing zone. If you decide instead to use a parameters.json (as the one proposed below), you will need to declare parameters in the Bicep file so it can fetch them from the JSON file.
```
jtracey93, line 13:
```suggestion
The modules are designed to be instantiated many times, once for each desired landing zone. Since the logic is captured in the registry's module, you may use the below proposed Bicep code samples to be the source of parameters instead of the usual `parameters.json` file, and have in this case one Bicep file per landing zone. If you decide instead to use a `parameters.json` (as the one proposed below), you will need to declare parameters in the Bicep file so it can accept them as parameter inputs from the parameters JSON file.
```

jtracey93 (changes requested):

jtracey93:
@gbeaud a few changes please. I think the test is failing as the version in the metadata section of `main.json` has likely changed (which is annoying) 

@shenglol is there any way in the test to expose what the diff/change is to help troubleshoot? Also it may be worth updating the test triggers to not trigger on doc updates? (I'm happy to put a PR in for that, if you agree)

dciborow:
> @gbeaud a few changes please. I think the test is failing as the version in the metadata section of `main.json` has likely changed (which is annoying)
> 
> @shenglol is there any way in the test to expose what the diff/change is to help troubleshoot? Also it may be worth updating the test triggers to not trigger on doc updates? (I'm happy to put a PR in for that, if you agree)

@shenglol, I think we should consider adding a path filter that excludes something like `*.md`. There seems to be something funny going on here, but its unrelated to the changes being proposed.

modules/lz/sub-vending/README.md
```diff
@@ -199,6 +199,8 @@ module sub003 'br/public:lz/sub-vending:1.2.2' = {
 
 ### Example JSON Parameter File
 
+These parameters would need to be declared in the Bicep file using them (currently not present in the above provided Bicep samples). Note that it is not necessary to use JSON parameter files here since in the proposed design pattern, the Bicep file would be the source of parameters.
```
dciborow, line 202: I understand the other examples show how to hardcode the values directly into the module, which avoids the new for a parameters file. But I still find myself using a parameters.json pretty frequently. Very particularly when deploying a bicep file from VS Code.

I think it's great you have included the parameters.json example, but I think it would be best to just also include here an example of the .bicep file that uses it.

In the example, the parameters can be greatly reduced (compared to main.bicep) and be something like this (replacing '...' with the rest of the params, which FYI GitHub Copilot does a great job helping to auto-complete)

```bicep
param subscriptionDisplayName string
param subscriptionAliasName string
param subscriptionBillingScope string
@allowed([ 'DevTest', 'Production' ])
param subscriptionWorkload string = 'Production'
param existingSubscriptionId string = ''
param subscriptionManagementGroupAssociationEnabled bool = true
...

module sub005 'br/public:lz/sub-vending:1.2.2' = {
  name: 'sub005'
  params: {
    subscriptionDisplayName: subscriptionDisplayName
    subscriptionAliasName: subscriptionAliasName
    subscriptionBillingScope: subscriptionBillingScope
    subscriptionWorkload: subscriptionWorkload 
    existingSubscriptionId: existingSubscriptionId 
    subscriptionManagementGroupAssociationEnabled: subscriptionManagementGroupAssociationEnabled
    ...
  }
}

output subscriptionId string = sub005.outputs.subscriptionId
output subscriptionResourceId string = sub005.outputs.subscriptionResourceId
```

dciborow:
```suggestion
These parameters need to be declared in the Bicep file using them as shown in this example. Note that it is not necessary to use JSON parameter files here since in the proposed design pattern, the Bicep file would be the source of parameters.

```bicep
param subscriptionDisplayName string
param subscriptionAliasName string
param subscriptionBillingScope string
@allowed([ 'DevTest', 'Production' ])
param subscriptionWorkload string = 'Production'
param existingSubscriptionId string = ''
param subscriptionManagementGroupAssociationEnabled bool = true
...

module sub005 'br/public:lz/sub-vending:1.2.2' = {
  name: 'sub005'
  params: {
    subscriptionDisplayName: subscriptionDisplayName
    subscriptionAliasName: subscriptionAliasName
    subscriptionBillingScope: subscriptionBillingScope
    subscriptionWorkload: subscriptionWorkload 
    existingSubscriptionId: existingSubscriptionId 
    subscriptionManagementGroupAssociationEnabled: subscriptionManagementGroupAssociationEnabled
    ...
  }
}

output subscriptionId string = sub005.outputs.subscriptionId
output subscriptionResourceId string = sub005.outputs.subscriptionResourceId
```
```

dciborow:
> @gbeaud a few changes please. I think the test is failing as the version in the metadata section of `main.json` has likely changed (which is annoying)
> 
> @shenglol is there any way in the test to expose what the diff/change is to help troubleshoot? Also it may be worth updating the test triggers to not trigger on doc updates? (I'm happy to put a PR in for that, if you agree)

@jtracey93, to reproduce what is going on, you can open up the branch, and run the "brm generate" command in the directory. It looks like the output of "brm generate" may have changed. Now, none of this has anything to do with the changes to the README, but it may be worth still looking into and understanding what's going on.

ChrisSidebotham:
@jtracey93 - Is this still in progress 🤷

PR #292, opened 2023-03-24
Skip module checks if only README has changed

dciborow:
## Description

This PR updates the changed-module detection, and no longer considers a module updated if only the README.md has changed. This prevents all the pipelines from kicking off when we have only made documentation updates.


shenglol (approved):
:shipit:

PR #305, opened 2023-04-05
ci: Add PR Title Prefix Check and Auto Update README.md versions

### LINKED ISSUE #293: [Brm Feature] auto update version number in README
Currently, when we make an update to a module, we must manually update every example in the README manually. When we forget, users can get confused and start off using outdated versions of the module and may experience issues which have already been fixed in the latest release. 

When we run `brm generate`, it should update all of the version numbers in the README automatically.

This may be complicated by the fact we don't include the patch version number in the version.json. We may want to consider including the full version number here.
Or, if we are only going to include the Major.Minor in the version.json, a 2nd option would be to publish two modules to the registery. One with the existing versioning "Major.Minor.Patch", and a 2nd with just the "Major.Minor", where this version would be overwritten after every patch update.
Then, in the README.md, all of the examples can be defined by the MAJOR.MINOR, and users would not have to worry about updating their bicep code for patch updates. This is similar to how GitHub actions are versioned, where a user can choose between using @v1, @v1.5 or @v1.5.3.

dciborow:
Closes #293

## Description

https://github.com/amannn/action-semantic-pull-request

The modules in the repository already follow a Semantic Versioning, which is manually maintained. This GitHub action checks that PRs include a prefix. This acts as a stepping stone to automating the version incrementing, this action requires each PR have a semantic prefix. Examples include: 
- fix: Correct typo
- feat: Add support for Node.js 18
- refactor!: Drop support for Node.js 12
- feat(ui): Add Button component

The full list of [recommended prefixes](https://www.conventionalcommits.org/en/v1.0.0/):
fix:, feat:, build:, chore:, ci:, docs:, style:, refactor:, perf:, test:

For versioning
1. fix: a commit of the type fix patches a bug in your codebase (this correlates with [PATCH](http://semver.org/#summary) in Semantic Versioning).
2. feat: a commit of the type feat introduces a new feature to the codebase (this correlates with [MINOR](http://semver.org/#summary) in Semantic Versioning).
3. BREAKING CHANGE: a commit that has a footer BREAKING CHANGE:, or appends a ! after the type/scope, introduces a breaking API change (correlating with [MAJOR](http://semver.org/#summary) in Semantic Versioning). A BREAKING CHANGE can be part of commits of any type.

The Readme auto-gen can be seen working over here.
https://github.com/dciborow/bicep-registry-modules/pull/31


.github/workflows/push-auto-generate.yml
```diff
@@ -53,6 +55,27 @@ jobs:
...
+        shell: bash
+        env:
+          SAMPLEFOLDER_PATH: ${{ needs.get-module-to-validate.outputs.module_dir }}
+          OLD_VERSION: ${{ steps.tag_version.outputs.previous_version }}
+          VERSION: ${{ steps.tag_version.outputs.new_version }}
+        run: |
+          cd $SAMPLEFOLDER_PATH
+
+          sed -ri "s/$OLD_VERSION/$VERSION/" "README.md" || exit 1
+
```
shenglol, line 78: Sorry, I renamed the file, so you might need to merge main and resolve the conflict.

.github/workflows/semantic-pr-check.yml
```diff
@@ -0,0 +1,17 @@
+name: "Semantic PR Check"
```
shenglol, line 1: Could you change the file name to `on-pull-request-check-pr-title`? The convention I'm trying to set up for workflow files is to have the trigger be a file name prefix.

.github/workflows/push-auto-generate.yml
dciborow: rolling the changes to this file back for the moment, they need more testing.

.github/workflows/semantic-pr-check.yml
dciborow: I like this convention and will try and keep it in mind.

shenglol (approved):
:shipit:

PR #311, opened 2023-04-07
Move module table to a GitHub page

shenglol:
As a workaround for the company-wise policy that blocks auto PR creation, the PR updated the workflow that publishes the module index data to also publish a [GitHub page](https://azure.github.io/bicep-registry-modules/) to show available modules and their versions. The module table in the root readme file is removed. Below is a screenshot:

image


.github/workflows/publish-module-index.yml
```diff
@@ -0,0 +1,106 @@
...
+        with:
+          client-id: ${{ secrets.PUBLISH_CLIENT_ID }}
+          tenant-id: ${{ secrets.PUBLISH_TENANT_ID }}
+          subscription-id: ${{ secrets.PUBLISH_SUBSCRIPTION_ID }}
+
+      - name: Upload to blob storage
+        uses: azure/CLI@v1
+        with:
+          inlineScript: |
+            az storage blob upload --account-name biceplivedatasaprod --container-name bicep-cdn-live-data-container --name module-index --file moduleIndex.json --auth-mode login --overwrite
```
jeskew, line 50: Looks like this step is not new to this PR (just moved from the `update-modules-metadata` workflow), but should we be concerned about a GH action overwriting a manifest file in a bucket named biceplivedatasa**prod**?

docs/jekyll/index.md
jeskew, line 1: Should this file be `.gitignore`d on `main` since it's generated on push?

.github/workflows/publish-module-index.yml
shenglol: It will create a new blob. The old blob name `bicepModulesMetadata` was a bit confusing because the file does not contain any metadata, but it's more like a module catalog. I've updated `aka.ms/bicepModulesMetadata` (used in Bicep [here](https://github.com/Azure/bicep/blob/b886719ca38c0b9f7e583919cb59d56a13eeeb36/src/Bicep.LangServer/Completions/PublicRegistryModuleMetadataProvider.cs#L26)) to point to the new blob https://live-data.bicep.azure.com/module-index. Will remove the old blob and update Bicep to use a new aka link `aka.ms/br-modules-index` later.

docs/jekyll/index.md
shenglol: I kept the file for local testing, but it makes sense to remove it.

.github/workflows/publish-module-index.yml
jeskew: I was more concerned about the fact that a GH action is reaching into something labeled "prod," but since it wasn't introduced in this PR, no need to solve it here.

jeskew (approved):

PR #359, opened 2023-05-04
feat(new): create private endpoint module

### LINKED ISSUE #258: [Module Proposal]: Private Endpoint Library

Yes

### Why is the module needed?

I can foresee that we will end up with many modules duplicating the same boilerplate code, that is required for common integrations, such as Private Endpoints for Azure Resources that support Private Link Service.

e.g: https://github.com/Azure/bicep-registry-modules/blob/main/modules/compute/container-registry/.bicep/nested_privateEndpoint.bicep

We should look to modularise this and reduce the duplication. Not currently many base modules in play, but just thinking ahead before we do end up with many that might beed to be refactored. We currently have a large collection of private modules (that we are looking to move away from) and we did use this approach and it was helpful in avoiding duplication and reducing overhead when we did, from time to time, need to tweak the pattern for private endpoint support in modules.

### Module path

lib/private-endpoint

### Describe the module

Essentially shift the common Private Endpoint code into its own library module and then modules that require it simply reference it from mcr.microsoft.com.

Advantage here would be that when we have many modules down the road using the same Private Endpoint code, we only have to modify the code in one place if enhancement is required and then update the Bicep registry module reference when modifying the individual modules down the road (or that bit could even be automated).

thomasriley:
## Description

Create a library module that can be consumed from other Bicep Modules for managing Private Endpoints

This closes #258.

## Adding a new module

- [ ] I have run deployment tests locally to ensure the module is deployable.


thomasriley:
I have opened two draft PRs to prepare and visualise how this module will be used.

AKV - a module that does not already support PEP - #361
ACR - a module that already nested this code, and will instead use the BR module - #362

Will hold off on creating any more PRs for now, but once this PR is merged I will continue with PRs on other modules.

thllxb:
I think it makes more sense to move this module to `modules/network/`

modules/lib/private-endpoint/main.bicep
```diff
@@ -0,0 +1,65 @@
+@description('Location for all Private Endpoint(s).')
+param location string
+
+@description('Array of Private Endpoint(s) to create.')
+param privateEndpoints array = []
```
thllxb, line 5: I think it make more sense to just create one pep in a module. It makes the code alot cleaner and easier to maintain.
Users can always use `for` loop on the module.

thllxb, line 5: It would be better if we can use User-defined data types which is now available with bicep 0.17.1

thomasriley: Intention for this module is to cut down on duplicate code across all the other base modules, so moving that `for` logic into each module deviates away from that goal.

thomasriley: I have refactored the module in this direction now.

modules/network/private-endpoint/main.bicep
```diff
@@ -0,0 +1,78 @@
...
+    customNetworkInterfaceName: !empty(customNetworkInterfaceName) ? customNetworkInterfaceName : null
+  }
+}
+
+resource privateDnsZoneGroup 'Microsoft.Network/privateEndpoints/privateDnsZoneGroups@2022-05-01' = if (!empty(privateDnsZones)) {
+  name: 'default'
+  parent: privateEndpoint
+  properties: {
+    privateDnsZoneConfigs: [for privateDnsZone in privateDnsZones: {
+      name: contains(privateDnsZone, 'name') ? privateDnsZone.name : 'default'
```
thllxb, line 63: this line can be `privateDnsZone.?name ?? 'default'`

modules/network/private-endpoint/main.bicep
```diff
@@ -0,0 +1,78 @@
...
+      }
+    ] : null
+    subnet: {
+      id: subnetId
+    }
+    customNetworkInterfaceName: !empty(customNetworkInterfaceName) ? customNetworkInterfaceName : null
+  }
+}
+
+resource privateDnsZoneGroup 'Microsoft.Network/privateEndpoints/privateDnsZoneGroups@2022-05-01' = if (!empty(privateDnsZones)) {
```
thllxb, line 58: this resource can use a for loop, like `[ for zone in privateDnsZones: { `

thllxb: nvm, the loop is in the resource and that is correct.

modules/network/private-endpoint/main.bicep
```diff
@@ -0,0 +1,78 @@
...
+param privateDnsZones array = []
+
+@description('Optional. Flag to enable manual approval for the Private Endpoint.')
+param manualApprovalEnabled bool = false
+
+@description('Optional. The name of the custom Network Interface to associate with the Private Endpoint.')
+param customNetworkInterfaceName string = ''
+
+@description('Optional. Tags to assign too the Private Endpoint.')
+param tags object = {}
```
thllxb, line 26: `param tags {*:string} = {}` is better

modules/network/private-endpoint/main.bicep
```diff
@@ -0,0 +1,78 @@
+@description('Required. The name of the Private Endpoint.')
+param name string
+
+@description('Required. Location for the Private Endpoint.')
+param location string
```
thllxb, line 5: it can use a default as `resourceGroup().location`, as per https://github.com/Azure/bicep-registry-modules/blob/main/CONTRIBUTING.md#parameters

modules/network/private-endpoint/main.bicep
```diff
@@ -0,0 +1,78 @@
+@description('Required. The name of the Private Endpoint.')
+param name string
```
thllxb, line 2:
```suggestion
@minLength(2)
@maxLength(64)
param name string
```

modules/network/private-endpoint/main.bicep
thllxb:
```suggestion
param tags {*:string} = {} 
```

modules/network/private-endpoint/main.bicep
thllxb:
```suggestion
      name: privateDnsZone.?name ?? 'default'
```

modules/network/private-endpoint/main.bicep
```diff
@@ -0,0 +1,78 @@
...
+param manualApprovalEnabled bool = false
+
+@description('Optional. The name of the custom Network Interface to associate with the Private Endpoint.')
+param customNetworkInterfaceName string = ''
+
+@description('Optional. Tags to assign too the Private Endpoint.')
+param tags object = {}
+
+resource privateEndpoint 'Microsoft.Network/privateEndpoints@2022-05-01' = {
+  name: '${name}-${uniqueString(name, subnetId, privateLinkServiceId)}'
```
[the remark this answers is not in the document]
thllxb:
```suggestion
  name: name
```

skydiver1980 (approved):

ChrisSidebotham:
Dear @thomasriley , Azure Verified Modules ([AVM](https://aka.ms/AVM)) will become the single Microsoft standard for Bicep modules. Transitioning to AVM, requests for new features or bug fixes will need to be submitted against the appropriate AVM module via [AVM Module Issues](https://aka.ms/AVM/Bicep/ModuleIssue). 

For bugs or new features with existing modules in the `/modules` directory of this repository, we encourage you to first try the AVM equivalent module if available and if the module is lacking any features, you can file an [AVM Module Issue](https://aka.ms/AVM/Bicep/ModuleIssue). If the module is not available yet, you can file a new [AVM Module Proposal](https://aka.ms/AVM/ModuleProposal).

See this [informational notice](https://github.com/Azure/bicep-registry-modules?tab=readme-ov-file#%EF%B8%8F-upcoming-changes-%EF%B8%8F) for more details on the upcoming changes.

Thank you for your understanding!

PR #402, opened 2023-06-06
chore: support prefix for LAW

vlahane:
support for providing prefix for LAW module


vlahane:
hello  @dciborow pls review, thanks!

modules/storage/log-analytics-workspace/main.bicep
```diff
@@ -1,7 +1,10 @@
+@description('Prefix of Log Analytics Workspace Resource Name.This param is ignored when name is provided.')
+param prefix string = 'law'
```
Gordonby, line 2: Could you use `log` as the default since it's the recommended abbreviation?

ref : https://learn.microsoft.com/en-us/azure/cloud-adoption-framework/ready/azure-best-practices/resource-abbreviations#management-and-governance

Gordonby:
There's probably the bigger discussion of parameter consistency here too. Should we be providing an separate prefix param in addition to the name parameter?

Personally i wouldn't add the additional parameter surface area, and would built it into the name parameter

`param name string = take('log-${uniqueString(resourceGroup().id, subscription().id)}', 63)`

🗣️ discuss! 

@alex-frankel

modules/storage/log-analytics-workspace/README.md
```diff
@@ -35,7 +36,7 @@ This module deploys Log Analytics workspace and optionally available integration
 param name string = 'my-logAnalyticsWorkspace-01'
 param location = 'eastus'
 
-module logAnalyticsWorkspace 'br/public:storage/log-analytics-workspace:1.0.1' = {
+module logAnalyticsWorkspace 'br/public:storage/log-analytics-workspace:0.0.1' = {
```
dciborow, line 39:
```suggestion
module logAnalyticsWorkspace 'br/public:storage/log-analytics-workspace:1.0.2' = {
```

dciborow:
> There's probably the bigger discussion of parameter consistency here too. Should we be providing an separate param to prefix the name in addition to the name parameter?
> 
> Personally i wouldn't add the additional parameter surface area, and would built it into the name parameter
> 
> `param name string = take('log-${uniqueString(resourceGroup().id, subscription().id)}', 63)`
> 
> 🗣️ discuss!
> 
> @alex-frankel

@Gordonby , ill chat with you about why I use the prefix param, and try and showcase the feature.

@vlahane the test case should use the prefix as the input and not use the "name". (doing this for a bunch of test cases makes a good example)

dciborow:
I'm also waving the whiteflag on having location be required now that I'm not using the Azure Marketplace. You could update the default value to resourceGroup().location if you would like.

Gordonby:
> > There's probably the bigger discussion of parameter consistency here too. Should we be providing an separate param to prefix the name in addition to the name parameter?
> > Personally i wouldn't add the additional parameter surface area, and would built it into the name parameter
> @Gordonby , ill chat with you about why I use the prefix param, and try and showcase the feature.

Sure sure. Would be good to hear. @dciborow

dciborow (approved):

modules/storage/log-analytics-workspace/test/main.test.bicep
```diff
@@ -21,9 +21,13 @@ param tags object = {
 
 module test_01 '../main.bicep' = {
   name: '${uniqueName}-test01'
```
dciborow, line 23:
```suggestion
  prefix: 'test01'
```

modules/storage/log-analytics-workspace/test/main.test.bicep
```diff
@@ -21,9 +21,13 @@ param tags object = {
...
+}
+
+module test_02 '../main.bicep' = {
+  name: '${uniqueName}-test02'
   params: {
-    name: 'test01${uniqueName}'
-    location: location
+    name: 'test02-${uniqueName}'
```
dciborow, line 29:
```suggestion
    prefix: 'test02'
```

shenglol (approved):

Issue #399, opened 2023-06-02
[Module Proposal]: Azure Content Moderator


OmegaVVeapon:
### Have you checked this module does not already exist in the registry?

Yes

### Why is the module needed?

We require a base module to create Azure Content Moderator instances.
https://learn.microsoft.com/en-us/azure/cognitive-services/content-moderator/overview

### Module path

cognitiveservices/content-moderator

### Describe the module

Allows the provisioning and management of Azure Content Moderator, including pricing plan selection and support for VNET integration via private endpoints.

dciborow:
Will work on this one with @OmegaVVeapon  in parallel to #398

OmegaVVeapon:
I'm closing this.
It was already completed as part of https://github.com/Azure/bicep-registry-modules/pull/404.

Issue #398, opened 2023-06-02
[Module Proposal]: Azure OpenAI


OmegaVVeapon:
### Have you checked this module does not already exist in the registry?

Yes

### Why is the module needed?

We require a base module to create Azure OpenAI instances.
https://learn.microsoft.com/en-us/azure/cognitive-services/openai/overview

Like Speech Services, this is another cognitive service so it'll follow the path set by the module pending merge in https://github.com/Azure/bicep-registry-modules/issues/248

### Module path

cognitiveservices/openai

### Describe the module

Allows the provisioning and management of Azure OpenAI, including pricing plan selection and support for VNET integration via private endpoints.

dciborow:
I'm aligned with this one.

@OmegaVVeapon , is this something you could propose? I'm happy to support you on this and help.

OmegaVVeapon:
I'm closing this.
It was already completed as part of #404.

PR #404, opened 2023-06-08
feat: Cognitive Services module

### LINKED ISSUE #248: [Module Proposal]: Azure Speech Services

Yes

### Why is the module needed?

We require a base module for managing Azure Speech Service instances and one is not currently available in the public registry. It will follow the same concept of the Azure Container Registry module that was previously created by @thomasriley via https://github.com/Azure/bicep-registry-modules/issues/166

The intent of the module is to focus specifically on the Speech feature of Azure Cognitive Services: other features such as Vision, OpenAI, etc. are outside of our requirement scope.

### Module path

cognitiveservices/speech

### Describe the module

Allows the provisioning and management of Azure Speech Services, including pricing plan selection and support for VNET integration via private endpoints.

OmegaVVeapon:
## Description

Closes #248, #398 and #399

## Adding a new module


modules/cognitiveservices/multi/main.bicep
```diff
@@ -0,0 +1,172 @@
+@description('The location into which your Azure resources should be deployed.')
+param location string = resourceGroup().location
+
+@description('The kind of Cognitive Service to create. See: https://learn.microsoft.com/en-us/azure/cognitive-services/create-account-bicep for available kinds.')
+param kind string
```
dciborow, line 5: can we switch this term to "prefix" to follow the other templates, and provide a default value (ill link the website that has the mapping of default names)

OmegaVVeapon: mmm... not sure that feels right...

This `kind` here maps to the kind of Cognitive Service we'll create. It's more significant than just naming.

The `prefix` defined in https://github.com/Azure/bicep-registry-modules/blob/main/CONTRIBUTING.md#mainbicep serves a different purpose.

If anything I think I could _add_ `prefix` `param` to this but I wouldn't switch it for `kind`.

shenglol:
Could we make the module name more descriptive? To me it's hard to tell what resources will be deployed by a `multi` module. 

Also, I wonder if it makes sense to put the AI/ML related modules in new module folder `ai`. @alex-frankel @dciborow Thoughts?

OmegaVVeapon:
> Could we make the module name more descriptive? To me it's hard to tell what resources will be deployed by a `multi` module.
> 
> Also, I wonder if it makes sense to put the AI/ML related modules in new module folder `ai`. @alex-frankel @dciborow Thoughts?

I'm ready glad you brought this up.
So, this `multi` name was really just a placeholder I used because I couldn't think of a good name for a module that can handle any `kind` of Cognitive Service (there's like 15 or so?) so I'm more than open to suggestions.

`/ai/cognitiveservices` sounds alright to me.

shenglol:
Sounds good

modules/ai/cognitiveservices/main.bicep
```diff
@@ -0,0 +1,175 @@
+@description('The kind of Cognitive Service to create. See: https://learn.microsoft.com/en-us/azure/cognitive-services/create-account-bicep for available kinds.')
+param kind string
```
dciborow, line 2: can we add 'Allowed' annotation here with the list of "kinds".
(totally optional, but perhaps a good place to create a user-defined type)

```suggestion
@allowed(['CognitiveServices', 'ComputerVision', 'CustomVision.Prediction', 'CustomVision.Training', 'Face', 'FormRecognizer', 'SpeechServices', 'LUIS', 'QnAMaker', 'TextAnalytics', 'TextTranslation', 'AnomalyDetector', 'ContentModerator', 'Personalizer', 'OpenAI'])
param kind string
```

dciborow, line 2: i like to have a default value wherever we can.  Would a default value of 'CognitiveServices' here work okay?

```suggestion
param kind string = 'CognitiveServices'
```

modules/ai/cognitiveservices/main.bicep
```diff
@@ -0,0 +1,175 @@
...
+param customSubDomainName string = ''
+
+@description('The Public Network Access setting of the Cognitive Service. When false, only requests from Private Endpoints can access it.')
+param publicNetworkAccess bool = true
+
+@description('A list of private endpoints to connect to the Cognitive Service.')
+param privateEndpoints array = []
+
+@description('The name of the SKU.')
+param skuName string = 'F0'
```
dciborow, line 31: looks like another place to potentially use allowed annotation/user-defined type.

modules/ai/cognitiveservices/main.bicep
OmegaVVeapon: Mmmm... I think so.
Since it's the only `kind` that supports multiple services, it does seem like the reasonable choice here.

modules/ai/cognitiveservices/main.bicep
```diff
@@ -0,0 +1,175 @@
...
+    migrationToken: migrationToken
+    networkAcls: networkAcls
+    publicNetworkAccess: publicNetworkAccess ? 'Enabled' : 'Disabled'
+    restore: restore
+    restrictOutboundNetworkAccess: restrictOutboundNetworkAccess
+    userOwnedStorage: userOwnedStorage == [] ? null : userOwnedStorage
+  }
+}
+
+resource cognitiveServiceDeployment 'Microsoft.CognitiveServices/accounts/deployments@2022-12-01' = [for (deployment, index) in deployments: {
```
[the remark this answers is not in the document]
OmegaVVeapon: Ah, good catch. Updated.

shenglol (approved):

PR #408, opened 2023-06-09
chore: Recommend prettier extension in vscode and set format on save

StephenWeatherford:


.vscode/settings.json
```diff
@@ -3,7 +3,7 @@
   "files.autoSave": "onFocusChange",
   "git.autofetch": true,
   "githubPullRequests.pullBranch": "always",
-  "[json]": {
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
-  }
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "editor.formatOnSave": true,
```
shenglol, line 7: I'm open to enabling this for the JSON/JS files, but let's disable it for the Markdown files, because it will change how the tables are formatted and will cause failures when running `brm generate`.

StephenWeatherford: Everything under modules is currently ignored by prettier due to .prettierignore.  But went ahead and restricted the changes to js,json,jsonc

PR #420, opened 2023-06-20
feat: Publish module index on publish module

StephenWeatherford:
@shenglol I haven't actually tested this, just modeled it after publish-module-index in on-push-main. Does it makes sense for me to set this up and test in a forked branch first?

### DIFF
--- a/.github/workflows/publish-module.yml
+++ b/.github/workflows/publish-module.yml
@@ -71,3 +71,11 @@ jobs:
 
       - name: Publish module
         run: bicep publish "modules/${{ steps.parse-tag.outputs.module_path }}/main.json" --target "br:${{ secrets.PUBLISH_REGISTRY_SERVER }}/public/bicep/${{ steps.parse-tag.outputs.module_path }}:${{ steps.parse-tag.outputs.version }}" --documentationUri "${{ steps.parse-tag.outputs.documentation_uri }}" --force
+
+  publish-module-index:
+    needs: main
+    uses: ./.github/workflows/publish-module-index.yml
+    secrets:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-module.yml
```diff
@@ -71,3 +71,11 @@ jobs:
...
       - name: Publish module
         run: bicep publish "modules/${{ steps.parse-tag.outputs.module_path }}/main.json" --target "br:${{ secrets.PUBLISH_REGISTRY_SERVER }}/public/bicep/${{ steps.parse-tag.outputs.module_path }}:${{ steps.parse-tag.outputs.version }}" --documentationUri "${{ steps.parse-tag.outputs.documentation_uri }}" --force
+
+  publish-module-index:
+    needs: main
+    uses: ./.github/workflows/publish-module-index.yml
+    secrets:
+      PUBLISH_CLIENT_ID: ${{ secrets.PUBLISH_CLIENT_ID }}
+      PUBLISH_TENANT_ID: ${{ secrets.PUBLISH_TENANT_ID }}
+      PUBLISH_SUBSCRIPTION_ID: ${{ secrets.PUBLISH_SUBSCRIPTION_ID }}
```
shenglol, line 81: Unfortunately, we cannot publish module index right after publishing the module because there's a delay in synchronizing the modules to MCR from the ACR to which the module is published.

shenglol: One workaround is to add a job to poll the MCR API to see if the published tag is available before publishing module index.

StephenWeatherford: I could just set it up to run nightly, I guess.

shenglol: That works too.

PR #441, opened 2023-07-07
feat: Validate publish is successful

StephenWeatherford:
Tested with a version of the script that skipped the publish step.


.github/workflows/publish-module.yml
```diff
@@ -71,3 +71,56 @@ jobs:
 
       - name: Publish module
         run: bicep publish "modules/${{ steps.parse-tag.outputs.module_path }}/main.json" --target "br:${{ secrets.PUBLISH_REGISTRY_SERVER }}/public/bicep/${{ steps.parse-tag.outputs.module_path }}:${{ steps.parse-tag.outputs.version }}" --documentationUri "${{ steps.parse-tag.outputs.documentation_uri }}" --force
+
+      - name: Validate publish
+        run: |
+          Version="${{ steps.parse-tag.outputs.version }}"
+          ModulePath="${{ steps.parse-tag.outputs.module_path }}"
+
+          time_limit_seconds=30
```
shenglol, line 80: Sometimes it can take a few minutes for the module to show up in the MCR catalog. I think it would be safer to increase the time limit to 10 minutes.

shenglol (approved):

PR #443, opened 2023-07-10
ci: Disable PR merging with certain labels

shenglol:
We should prevent PRs from merging if they have one of the following labels:

image

### DIFF
--- a/.github/workflows/on-pull-request-check-labels.yml
+++ b/.github/workflows/on-pull-request-check-labels.yml
@@ -1,0 +1,32 @@
+name: Check Labels
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - labeled
+      - unlabeled
+      - synchronize
+
+jobs:
+  label-check:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: mheap/github-action-required-labels@v5
+        with:
+          mode: exactly
+          count: 0
+          labels: "Awaiting Module Path Assignment"
+          add_comment: true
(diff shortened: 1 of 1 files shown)


.github/workflows/on-pull-request-check-labels.yml
```diff
@@ -0,0 +1,32 @@
...
+          add_comment: true
+          message: "This PR is being prevented from merging because it is pending on module path assignment."
+
+      - uses: mheap/github-action-required-labels@v5
+        with:
+          mode: exactly
+          count: 0
+          labels: "Awaiting MCR Manifest Onboarding"
+          add_comment: true
+          message: "This PR is being prevented from merging because it is pending on MCR manifest update."
```
dciborow, line 32:
```suggestion
jobs:
  label-check:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        labels:
          - "Awaiting Module Path Assignment"
          - "Awaiting MCR Manifest Onboarding"
    permissions:
      issues: write
      pull-requests: write
    steps:
      - uses: mheap/github-action-required-labels@v5
        with:
          mode: exactly
          count: 0
          labels: ${{ matrix.labels }}
          add_comment: true
          message: "This PR is being prevented from merging because it is pending on label: ${{ matrix.labels }}."
```

You could use a matrix to foreach loop instead of repeating. (won't be offended if you decline)

dciborow (approved):

.github/workflows/on-pull-request-check-labels.yml
shenglol: Good point!

StephenWeatherford (approved):

PR #523, opened 2023-09-25
feat: avm/res/network/private-endpoint

eriqua:
## Description

Adding `network/private-endpoint` AVM Resource Module to the `avm/res` folder and corresponding module pipeline to `.github/workflows`


AlexanderSehr:
I'd suggest to recreate this one from main once the CI is merged

avm/res/network/private-endpoint/tests/e2e/defaults/main.test.bicep
```diff
@@ -0,0 +1,63 @@
...
+metadata name = 'Using only defaults'
+metadata description = 'This instance deploys the module with the minimum set of required parameters.'
+
+// ========== //
+// Parameters //
+// ========== //
+
+@description('Optional. The name of the resource group to deploy for testing purposes.')
+@maxLength(90)
+param resourceGroupName string = 'network.privateendpoints-${serviceShort}-rg'
```
AlexanderSehr, line 12:
```suggestion
param resourceGroupName string = 'network.privateendpoints-${namePrefx}-${serviceShort}-rg'
```
Just a reminder to add the `namePrefix` param reference here.

eriqua: Done with a different convention as discussed

avm/res/network/private-endpoint/main.bicep
```diff
@@ -0,0 +1,192 @@
+metadata name = 'Private Endpoints'
+metadata description = 'This module deploys a Private Endpoint.'
+metadata owner = 'Azure/module-maintainers'
+
+@description('Optional. Name of the private endpoint resource to create.')
+param name string = 'pep-${last(split(serviceResourceId, '/'))}'
```
jtracey93, line 6: Should this not have a default as a resource module?

But when it's used, optionally as an interface, it's should have this default name as it's not the primary resource?

AlexanderSehr: I added this today as per the conversation about CAF naming. You tell me @jtracey93 ;) It's now essentially applying the same default as if it were used by another resource with the exception that we also add the `groupId` there.

jtracey93: As this is being pushed as a resource module we should align to SNFR25 (https://azure.github.io/Azure-Verified-Modules/specs/shared/#id-snfr25---category-composition---resource-naming)

So technically this should not have a default for the name as it's a resource module. 

But if it's used in a cross reference, say for a key vault 😁, to implement private endpoints. Then in that calla of the module the name can be defaulted to as you show. 

That's what I intended from what I published for SNFR25 above.

@AlexanderSehr

AlexanderSehr: Fair, undone.

AlexanderSehr (approved):

PR #530, opened 2023-09-30
[CI Environment] Updated src template for test files used by `Set-AVMModule` utility

AlexanderSehr:
## Description

 Updated src template for test files used by `Set-AVMModule` utility. It should reference the namePrefix as per the latest change.


eriqua (approved):
🆗

PR #535, opened 2023-10-03
fix: Fixed readme utility bug when working with nullable types

AlexanderSehr:
## Description

Fixed readme utility bug when working with nullable types


jtracey93 (approved):
LGTM

PR #536, opened 2023-10-03
fix: avm-res-network-privateendpoints update

eriqua:
## Description

- Updating bicep template removing redundant code conditions for nullable properties.
- Updating test files with additional empty non-required parameters to cater for a test-specific limitation.


jtracey93 (approved):
LGTM

PR #538, opened 2023-10-06
feat: Alternative version pattern

AlexanderSehr:
## Description

Introduced alternative pattern for versioning. Moving out of draft once approved.

Context
- Using the original approach can be an issue when inner-sourcing the modules as they'd require a token replacement at all times. Without the token replacement, the deployment name for telemetry would be invalid
- This alternative approach is always valid and makes the extra variable redundant

Changes
- Removed version variable
- Introduced valid placeholder directly into telemetry deployment name
- Introduced 'absolute token' to logic that don't apply the default pre & suffix to a token. Otherwise we'd be where we started.

Ref
- [Publish test](https://github.com/AlexanderSehr/bicep-registry-modules/actions/runs/6430718629/job/17462428954)
- [Deploy test](https://github.com/AlexanderSehr/bicep-registry-modules/actions/runs/6429115545)


.github/actions/templates/avm-validateModuleDeployment/action.yml
```diff
@@ -93,7 +93,6 @@ runs:
             subscriptionId    = '${{ inputs.subscriptionId }}'
```
AlexanderSehr, line 93: Should be merged after #539

eriqua (approved):
:ok:

PR #539, opened 2023-10-06
feat: Adjust `generate-module-index-data.js` to support AVM 

jtracey93:
## Description

This PR updates `generate-module-index-data.js` to add support for the AVM modules.

This will allow the AVM modules to be discovered by the Bicep clients/IDEs using the Bicep IDE extensions.

@shenglol & @StephenWeatherford please review as JavaScript is not my daily driver 👍

### DIFF
--- a/scripts/github-actions/generate-module-index-data.js
+++ b/scripts/github-actions/generate-module-index-data.js
@@ -205,7 +205,7 @@ async function generateModuleIndexData({ require, github, context, core }) {
           core,
           path,
           modulePath,
-          (moduleType = "avm-res"),
+          (moduleType = "avm-ptn"),
           tag,
           context
         );
--- a/scripts/github-actions/generate-module-index-data.js
+++ b/scripts/github-actions/generate-module-index-data.js
@@ -16,7 +16,7 @@ async function getModuleDescription(
   tag,
   context
 ) {
-  const allowedModuleRoots = ["brm", "avm/res", "avm/res"];
+  const allowedModuleRoots = ["modules", "avm/res", "avm/res"];
 
   if (!allowedModuleRoots.includes(moduleRoot)) {
     throw new Error(
--- a/scripts/github-actions/generate-module-index-data.js
+++ b/scripts/github-actions/generate-module-index-data.js
@@ -12,15 +12,32 @@ async function getModuleDescription(
   core,
   path,
   modulePath,
+  moduleType = "brm",
   tag,
   context
 ) {
   // Retrieve the main.json file as it existed for the given tag
   const ref = `tags/${modulePath}/${tag}`;
   core.info(`  Retrieving main.json at ref ${ref}`);
-  const mainJsonPath = path
-    .join("modules", modulePath, "main.json")
-    .replace(/\\/g, "/");
+
+  let mainJsonPath;
+  if (moduleType === "brm") {
+    mainJsonPath = path
+      .join("modules", modulePath, "main.json")
+      .replace(/\\/g, "/");
+  } else if (moduleType === "avm-res") {
+    mainJsonPath = path
+      .join("avm/res", modulePath, "main.json")
+      .replace(/\\/g, "/");
+  } else if (moduleType === "avm-ptn") {
+    mainJsonPath = path
+      .join("avm/ptn", modulePath, "main.json")
+      .replace(/\\/g, "/");
+  } else {
+    throw new Error(
+      `Invalid module type provided to getModuleDescription function, permitted type are brm, avm-res, avm-ptn. The module type provided was: ${moduleType}`
+    );
+  }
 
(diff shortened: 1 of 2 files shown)


scripts/github-actions/generate-module-index-data.js
```diff
@@ -12,15 +12,32 @@ async function getModuleDescription(
   core,
   path,
   modulePath,
+  moduleType = "brm",
```
shenglol, line 15: I'd suggest changing `moduleType` to `moduleRoot` whose value can be `modules`, `avm/res` and `avm/ptn`. Then you can eliminate `if...else if...else`.

shenglol:
Did you test the changes locally with https://github.com/Azure/bicep-registry-modules/blob/main/scripts/github-actions/runlocally/run-generate-module-index.js?

jtracey93:
> Did you test the changes locally with https://github.com/Azure/bicep-registry-modules/blob/main/scripts/github-actions/runlocally/run-generate-module-index.js?

I did indeed. Although was getting unable to find some commits/tags on downloading the descriptions for all modules. With the file unchanged before I even started. 

But I tested the logic locally and it output the right data when I mocked it

shenglol:
> > Did you test the changes locally with https://github.com/Azure/bicep-registry-modules/blob/main/scripts/github-actions/runlocally/run-generate-module-index.js?
> 
> I did indeed. Although was getting unable to find some commits/tags on downloading the descriptions for all modules. With the file unchanged before I even started.
> 
> But I tested the logic locally and it output the right data when I mocked it

That's good. Thanks for testing it. I just realized that we should also add a filter in https://github.com/Azure/bicep-registry-modules/blob/main/scripts/github-actions/generate-module-index-md.js to exclude `avm` modules, since they are going to have its own module index site.

StephenWeatherford:
FYI Ari was looking into ways to add tests for js/ts files like this.

scripts/github-actions/generate-module-index-data.js
```diff
@@ -12,14 +12,24 @@ async function getModuleDescription(
   core,
   path,
   modulePath,
+  moduleRoot,
   tag,
   context
 ) {
+  const allowedModuleRoots = ["brm", "avm/res", "avm/res"];
```
shenglol, line 19:
```suggestion
  const allowedModuleRoots = ["modules", "avm/res", "avm/res"];
```

scripts/github-actions/generate-module-index-data.js
```diff
@@ -12,14 +12,24 @@ async function getModuleDescription(
...
+    );
+  }
   // Retrieve the main.json file as it existed for the given tag
   const ref = `tags/${modulePath}/${tag}`;
   core.info(`  Retrieving main.json at ref ${ref}`);
-  const mainJsonPath = path
-    .join("modules", modulePath, "main.json")
+
+  let mainJsonPath;
+  mainJsonPath = path
```
shenglol, line 31:
```suggestion
  const mainJsonPath = path
```

scripts/github-actions/generate-module-index-data.js
```diff
@@ -98,13 +117,114 @@ async function generateModuleIndexData({ require, github, context, core }) {
...
+  }
+
   core.info(`Writing moduleIndex.json`);
   await fs.writeFile(
     "moduleIndex.json",
     JSON.stringify(moduleIndexData, null, 2)
   );
 
-  core.info(`Processed ${moduleGroups.length} modules groups.`);
+  core.info(`Processed ${allModuleGroups.length} modules groups.`);
```
shenglol, line 226: It seems we only use `allModuleGroups` to log how many module groups are processed. I think we can just remove `allModuleGroups` and replace `allModulegroups.length` with `moduleGroups.length + moduleGroupsAvmRes.length + moduleGroupsAvmPtn.length`.

jtracey93: changed

scripts/github-actions/generate-module-index-data.js
```diff
@@ -98,13 +117,114 @@ async function generateModuleIndexData({ require, github, context, core }) {
...
+          properties,
+        });
+      } catch (error) {
+        core.setFailed(error);
+      }
+    }
+  }
+
+  // AVM Pattern Modules
+  for (const moduleName of moduleGroupsAvmPtn) {
```
shenglol, line 168: [non-blocking]: The two loops for processing AVM modules are almost identical (they are similar to brm loop as well). It would be good if we could extract the shared logic in a helper function to remove duplicates. This can be done in the future PR though.

shenglol (approved):

scripts/github-actions/generate-module-index-data.js
jtracey93: completely agree, will take a go at this in a new PR

PR #541, opened 2023-10-06
fix: Adjusted the `versionListUrl` for AVM to reference the published module name

AlexanderSehr:
## Description

Currently, the logic references the tags using the url `https://mcr.microsoft.com/v2/bicep/avm/res/network/private-endpoint/tags/list`

this should be `https://mcr.microsoft.com/v2/bicep/avm-res-network-privateendpoint/tags/list`


jtracey93 (approved):
LGTM

PR #542, opened 2023-10-07
feat: Adjust `generate-module-index-data.js` `moduleJsonPath` for AVM modules

jtracey93:
## Description

Adjust `generate-module-index-data.js` `moduleJsonPath` for AVM modules to remove double `avm/res/avm/res` or `avm/ptn/avm/ptn` paths to main.json

Testing of API manually works as expected: https://api.github.com/repos/Azure/bicep-registry-modules/contents/avm/res/network/private-endpoint/main.json?ref=tags/avm/res/network/private-endpoint/0.1.0

And have tested logic in output


AlexanderSehr (approved):

AlexanderSehr:
-> Will be followed up by a better fix

PR #547, opened 2023-10-07
feat: New module `kube-config/extension` 

AlexanderSehr:
## Description

- Migrated kube-extension module to AVM (this will lead to a release `0.1.0`)

| Pipelines |
| - |
| [avm.res.kubernetes-configuration.extension](https://github.com/AlexanderSehr/bicep-registry-modules/actions/workflows/avm.res.kubernetes-configuration.extension.yml) |


eriqua (changes requested):
I'd suggest two separate PRs, one for each module.
Probably a good practice in general, but in particular in this case, since `extension` is dependent on `flux-config`. We should first bump `flux config` and then directly reference the latest 0.1.1 from `extension`.

AlexanderSehr:
> I'd suggest two separate PRs, one for each module. Probably a good practice in general, but in particular in this case, since `extension` is dependent on `flux-config`. We should first bump `flux config` and then directly reference the latest 0.1.1 from `extension`.

Note taken and will do. Just for background: The reason I opted against it was that it is only a readme update at the end of the day (so there is not even a patch change for the functionality). However, a separate PR won't hurt

avm/res/kubernetes-configuration/extension/main.bicep
```diff
@@ -0,0 +1,112 @@
...
+      } : null
+      namespace: !empty(targetNamespace ?? '') ? {
+        targetNamespace: targetNamespace
+      } : null
+    }
+    version: version
+  }
+}
+
+module fluxConfiguration 'br/public:avm-res-kubernetesconfiguration-fluxconfiguration:0.1.0' = [for (fluxConfiguration, index) in (fluxConfigurations ?? []): {
```
eriqua, line 85:
```suggestion
module fluxConfiguration 'br/public:avm-res-kubernetesconfiguration-fluxconfiguration:0.1.1' = [for (fluxConfiguration, index) in (fluxConfigurations ?? []): {
```

ChrisSidebotham (commented):
🆗

avm/res/kubernetes-configuration/extension/tests/e2e/defaults/main.test.bicep
```diff
@@ -0,0 +1,63 @@
+targetScope = 'subscription'
+
```
eriqua, line 2:
```suggestion

metadata name = 'Using only defaults'
metadata description = '''
This instance deploys the module with the minimum set of required parameters.
> **Note:** The test currently implements additional non-required parameters to cater for a test-specific limitation.
'''
```

avm/res/kubernetes-configuration/extension/tests/e2e/max/main.test.bicep
```diff
@@ -0,0 +1,84 @@
+targetScope = 'subscription'
+
```
eriqua, line 2:
```suggestion

metadata name = 'Using large parameter set'
metadata description = 'This instance deploys the module with most of its features enabled.'

```

avm/res/kubernetes-configuration/extension/tests/e2e/waf-aligned/main.test.bicep
```diff
@@ -0,0 +1,84 @@
+targetScope = 'subscription'
+
```
eriqua, line 2: metadata name = 'WAF-aligned'
metadata description = 'This instance deploys the module in alignment with the best-pratices of the Azure Well-Architected Framework.'

eriqua, line 2:
```suggestion

metadata name = 'WAF-aligned'
metadata description = 'This instance deploys the module in alignment with the best-pratices of the Azure Well-Architected Framework.'

```

eriqua (approved):
:ok:

PR #562, opened 2023-10-14
fix: Updated metadata removal to work with templates using User-defined-types

AlexanderSehr:
## Description
Updated metadata removal to work with templates using User-defined-types

Background: Once you add UDT in your template, resources in an ARM template are not of type but hashtable. This did not work with the original version of the script. 

Successfully validated & tested with KVLT module pipeline [run](https://github.com/AlexanderSehr/bicep-registry-modules/actions/runs/6518560487).


eriqua (approved):
:ok:

PR #563, opened 2023-10-14
feat: Improved readme performance & enforced that all readme content but `## Notes` is always created anew

AlexanderSehr:
## Description

- Updated the script so that
  - ReadMes are always regenerated from the ground up 
  - Only the content of a '## Notes' sections is kept and added at the end of the readme file
  - Improved performance by only fetching cross-references one (instead during each template run)
  - Improved usage of variables in threads
  - Updated readme with new header

Re-ran on all current scripts and (as expected) did not trigger any unexpected changes to anything already merged.

| Pipeline |
| - |
| [avm.res.network.private-endpoint](https://github.com/AlexanderSehr/bicep-registry-modules/actions/workflows/avm.res.network.private-endpoint.yml) |

PR #568, opened 2023-10-16
feat: Added Batch account module

AlexanderSehr:
## Description

- Added Batch account module

| Pipeline |
| - |
| [avm.res.batch.batch-account](https://github.com/AlexanderSehr/bicep-registry-modules/actions/workflows/avm.res.batch.batch-account.yml) |


avm/res/batch/batch-account/tests/e2e/waf-aligned/main.test.bicep
```diff
@@ -0,0 +1,109 @@
+targetScope = 'subscription'
+
+metadata name = 'WAF-aligned'
+metadata description = 'This instance deploys the module in alignment with the best-pratices of the Well-Architectured-Framework.'
```
eriqua, line 4: Ah this typo! Any automation to be updated?

```suggestion
metadata description = 'This instance deploys the module in alignment with the best-practices of the Azure Well-Architected Framework.'
```

avm/utilities/pipelines/sharedScripts/Set-ModuleReadMe.ps1
```diff
@@ -1107,7 +1107,7 @@ function Set-UsageExamplesSection {
 
     # Process content
     $SectionContent = [System.Collections.ArrayList]@(
-        'The following module usage examples are retrieved from the content of the files hosted in the module''s `tests` folder.',
+        "The following section provides usage examples for the module, which were used to validate and deploy the module successfully. For a full reference, please review the module's test folder in its repository.",
```
eriqua, line 1110: 👻

.github/workflows/avm.res.batch.batch-account.yml
```diff
@@ -0,0 +1,81 @@
+name: "avm.res.batch.batch-account"
```
eriqua, line 1: Wait for #563

avm/utilities/pipelines/sharedScripts/Set-ModuleReadMe.ps1
AlexanderSehr: Not a ghost @eriqua. If we don't have this change in (before #546) is merged, readme's would become outdated again.

avm/res/batch/batch-account/tests/e2e/waf-aligned/main.test.bicep
AlexanderSehr: No, it's actually manual

avm/utilities/pipelines/sharedScripts/Set-ModuleReadMe.ps1
AlexanderSehr: gone it is

avm/res/batch/batch-account/main.bicep
```diff
@@ -0,0 +1,404 @@
...
+param diagnosticSettings diagnosticSettingType
+
+@description('Optional. The customer managed key definition.')
+param customerManagedKey customerManagedKeyType
+
+@description('Optional. The managed identity definition for this resource.')
+param managedIdentities managedIdentitiesType
+
+var formattedUserAssignedIdentities = reduce(map((managedIdentities.?userAssignedResourcesIds ?? []), (id) => { '${id}': {} }), {}, (cur, next) => union(cur, next)) // Converts the flat array to an object like { '${id1}': {}, '${id2}': {} }
+var identity = !empty(managedIdentities) ? {
```
eriqua, line 85:
```suggestion

var identity = !empty(managedIdentities) ? {
```

avm/res/batch/batch-account/main.bicep
```diff
@@ -0,0 +1,404 @@
...
+
+  @description('Optional. Version of the condition.')
+  conditionVersion: '2.0'?
+
+  @description('Optional. The Resource Id of the delegated managed identity resource.')
+  delegatedManagedIdentityResourceId: string?
+}[]?
+
+type privateEndpointType = {
+
```
eriqua, line 324:
```suggestion
```

avm/res/batch/batch-account/main.bicep
AlexanderSehr: If wonder if that's worth commenting as it requires me to re-render the template. But regardless - I'd be against adding an empty line here as I clustered them together on purpose (i.e., they belong together)

eriqua: Just thinking it would improve readability to separate each var. I won't block the merge on such a small detail though.

avm/res/batch/batch-account/main.bicep
```diff
@@ -0,0 +1,403 @@
...
+  'Disabled'
+])
+param publicNetworkAccess string = ''
+
+@allowed([
+  'Allow'
+  'Deny'
+])
+@description('Optional. The network profile default action for endpoint access. It is only applicable when publicNetworkAccess is not explicitly disabled.')
+param networkProfileDefaultAction string = 'Deny'
```
eriqua, line 50: What about a `networkProfile` object hosting this and the next parameter?

AlexanderSehr: In that case we'd need another User Defined Type. Not saying we shouldn't, but this would need to be updated in all modules with PE (+ specs) and I suggest to do this  in a separate issue + PR.

avm/res/batch/batch-account/main.bicep
AlexanderSehr: I'd appreciate it if we don't 🙏

avm/res/batch/batch-account/main.bicep
eriqua: Do we have a spec saying objects cannot be used unless typed with udt? I agree that would be the end goal but for the moment we made exceptions, e.g. for child resources, and not only. Other modules have networkAcls in an object, without udt (e.g. cognitive services). NetworkAcls and networkProfile are basically the same thing. So yes, let's agree on something.

AlexanderSehr: For the time being, I would only like to enforce UDT for what we prepared with extension resources. If somebody wants to go the extra mile and add any UDT on top of that, be my guest (e.g., @ChrisSidebotham). But blocking PRs for that seems wrong to me as we have more than enough work porting the CARML modules as is, yet alone optimizing them for UDTs.

eriqua: I'm not asking for UDT, I'm suggesting to use an object for networkProfile, to align to what done for networkAcls for other modules. Adding udt next will be easier if we already have an object.

AlexanderSehr: I see. Sorry I'm only working of the comment. I'll compare the two and check it out

AlexanderSehr: Turns out it's actually a bit more different that I thought. May I suggest that we push the change to the module's logic back?

eriqua: OK, but I'll raise one PR to fix this and bump to 0.2.0 before getting this ORPHANED if that is OK

avm/utilities/tools/Set-AVMModule.ps1
```diff
@@ -118,6 +116,7 @@ function Set-AVMModule {
             ################
             if (-not $using:SkipReadMe) {
                 Write-Output "Generating readme for [$resourceTypeIdentifier]"
+                . $using:ReadMeScriptFilePath
```
eriqua, line 119: Please resolve conflicts, probably due to readme updates

avm/res/batch/batch-account/tests/e2e/encr/main.test.bicep
```diff
@@ -0,0 +1,86 @@
...
+// ========== //
+// Parameters //
+// ========== //
+
+@description('Optional. The name of the resource group to deploy for testing purposes.')
+@maxLength(90)
+param resourceGroupName string = 'dep-${namePrefix}-batch.batchaccounts-${serviceShort}-rg'
+
+@description('Optional. The location to deploy resources to.')
+param location string = deployment().location
```
eriqua, line 15: It looks like we have Azure Batch accounts per region per subscription default limit set to 3. Should we change the deployment location for one of the tests?

Ref: https://learn.microsoft.com/en-us/azure/batch/batch-quota-limit#resource-quotas

avm/res/batch/batch-account/main.bicep
eriqua: Done in PR #592

avm/res/batch/batch-account/tests/e2e/encr/main.test.bicep
AlexanderSehr: Instead of implement a workaround for one resource alone I'm wondering if we should invest some time to implement a CI-powered location rotation. What do you thing @eriqua?

avm/utilities/tools/Set-AVMModule.ps1
AlexanderSehr: done

avm/res/batch/batch-account/tests/e2e/encr/main.test.bicep
AlexanderSehr: Updated & retested @eriqua

eriqua (approved):
:ok:

PR #576, opened 2023-10-19
ci: Include documentationUri in module index data

shenglol:
This is the first step to fix https://github.com/Azure/bicep/issues/12122.

### DIFF
--- a/scripts/github-actions/generate-module-index-data.js
+++ b/scripts/github-actions/generate-module-index-data.js
@@ -10,42 +10,19 @@ async function getSubdirNames(fs, dir) {
 async function getModuleDescription(
   github,
   core,
-  path,
-  modulePath,
-  moduleRoot,
-  tag,
+  mainJsonPath,
+  gitTag,
   context
 ) {
-  const allowedModuleRoots = ["modules", "avm/res", "avm/ptn"];
+  const gitTagRef = `tags/${gitTag}`;
 
-  if (!allowedModuleRoots.includes(moduleRoot)) {
-    throw new Error(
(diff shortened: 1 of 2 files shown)


scripts/github-actions/generate-module-index-data.js
```diff
@@ -68,157 +45,107 @@ async function getModuleDescription(
...
-            context
-          );
-          if (description) {
-            properties[tag] = { description };
-            modulesWithDescriptions[modulePath] = true;
+  for (const avmModuleRoot of ["avm/res", "avm/ptn"]) {
+    const avmModuleGroups = await getSubdirNames(fs, avmModuleRoot);
+
+    for (const moduleGroup of avmModuleGroups) {
+      const moduleGroupPath = `avm/res/${moduleGroup}`;
```
jtracey93, line 105: Think we need some logic here to handle the pattern modules here too.

As in the 2 separate loops we had before handling this had the following values for:

- Resource module ```const modulePath = `avm/res/${moduleGroup}/${moduleName}`;```
- Pattern module ```const modulePath = `avm/ptn/${moduleName}`;```

jtracey93 (changes requested):
Hey @shenglol, just one thing about pattern modules i think we need to handle, but looks good otherwise 👍

scripts/github-actions/generate-module-index-data.js
shenglol: Good catch...I meant to use `avmModuleGroup` here. Will fix it.

shenglol: Done.

jtracey93 (approved):
LGTM 🚀

PR #597, opened 2023-10-31
fix: Fixed small issue when detecting required parameters & enhanced allowed & default values

AlexanderSehr:
## Description

- Fixed small issue when detecting required parameters
- Enhanced allowed & default values

| Pipeline |
| - | 
| [avm.res.key-vault.vault](https://github.com/AlexanderSehr/bicep-registry-modules/actions/workflows/avm.res.key-vault.vault.yml) |


eriqua (approved):
:ok:

PR #622, opened 2023-11-13
feat: Added user defined type for Key Vault Access Policies

johnlokerse:
## Description

This pull request introduces the type `accessPoliciesType` so the author experience is enhanced when using this parameter to set access policies. I have added the `accessPoliciesType` to `key-vault/vault/main.bicep` and `key-vault/vault/access-policy`. 

Besides the addition of the type I have removed `formattedAccessPolicies` because the user-defined type has taken over the task to enforce required properties. This makes it simpler for the user to read and use the key vault main.bicep.

I have tested the modules by deploying to my Azure environment:

1. **With** access policies configured in bicepparam:

```bicepparam
using './main.bicep'

param name = 'br-kv-john-test-no-ap'
param enableRbacAuthorization = false
param accessPolicies = [
  {
    objectId: '<object-id-here>'
    permissions: {
      certificates: [
        'all'
      ]
      keys: [
        'all'
      ]
      secrets: [
        'all'
      ]
      storage: [
        'all'
      ]
    }
    tenantId: '<tenant-id-here>'
  }
]
```

**Succesful deployment**
image

2. **Without** access policies configured in bicepparam to showcase it is optional:

```bicepparam
using './main.bicep'

param name = 'br-kv-john-test-no-ap'
```

**Succesful deployment**
image

- [x] I have read the [Updating an existing module](https://github.com/Azure/bicep-registry-modules/blob/main/CONTRIBUTING.md#updating-an-existing-module) section in the contributing guide and updated the `version.json` file properly:
  - [ ] The PR contains backwards compatible bug fixes, and I have NOT bumped the MAJOR or MINOR version in `version.json`.
  - [x] The PR contains backwards compatible feature updates, and I have bumped the MINOR version in `version.json`.
  - [ ] The PR contains breaking changes, and I have bumped the MAJOR version in `version.json`.
- [ ] I have updated the examples in README with the latest module version number.


johnlokerse:
The pull request doesn't contain breaking changes, but according to `CONTRIBUTING.md` it is required to bump the major version when changing a type (`Changing the type of a parameter`). Does the major version need a bump in this case?

avm/res/key-vault/vault/access-policy/main.bicep
```diff
@@ -35,3 +28,31 @@ output name string = policies.name
 
 @description('The resource ID of the access policies assignment.')
 output resourceId string = policies.id
+
+// ================ //
+// Definitions      //
+// ================ //
+type accessPoliciesType = {
+  @description('Required. The tenant ID that is used for authenticating requests to the key vault.')
+  tenantId: string
```
AlexanderSehr, line 37: Hey @johnlokerse, thanks for the contribution. You're right that the UDT takes care of the formattedAccessPolicies with one exception (which may have been the original reason it was there in the first place), that is, the default for the tenantId `tenant().tenantId`.

It would be nice from a usage perspective if we could still keep this feature. However, this would require us to keep the loop.

johnlokerse: Ah yes, I get it. I will revert it back!

johnlokerse: I might have an idea. Will test some cases and report back later :-)

johnlokerse: I have reverted the code back. Had an idea of using the map function, and checking if `tenantId` exists but realized I was writing the same code but in a different way and it went way more complex than a simple for-loop. Also tried something with the string `replace` function.

All in all, a `tenantId` property and value check and a for-loop are needed and I think the current solution is the way to go.

I have replaced the `contains` for the safe-dereference operator.

avm/res/key-vault/vault/access-policy/main.bicep
```diff
@@ -35,3 +35,31 @@ output name string = policies.name
 
 @description('The resource ID of the access policies assignment.')
 output resourceId string = policies.id
+
+// ================ //
+// Definitions      //
+// ================ //
+type accessPoliciesType = {
+  @description('Required. The tenant ID that is used for authenticating requests to the key vault.')
```
AlexanderSehr, line 43:
```suggestion
  @description('Optional. The tenant ID that is used for authenticating requests to the key vault.')
```

avm/res/key-vault/vault/access-policy/main.bicep
```diff
@@ -35,3 +35,31 @@ output name string = policies.name
...
+  tenantId: string
```
AlexanderSehr, line 44:
```suggestion
  @description('Optional. The tenant ID that is used for authenticating requests to the key vault.')
  tenantId: string?
```

avm/res/key-vault/vault/main.bicep
```diff
@@ -423,3 +423,28 @@ type lockType = {
   @description('Optional. Specify the type of lock.')
   kind: ('CanNotDelete' | 'ReadOnly' | 'None')?
 }?
+
+type accessPoliciesType = {
+  @description('Required. The tenant ID that is used for authenticating requests to the key vault.')
+  tenantId: string
```
AlexanderSehr, line 429:
```suggestion
  @description('Optional. The tenant ID that is used for authenticating requests to the key vault.')
  tenantId: string?
```

johnlokerse: Fixed!

avm/res/key-vault/vault/access-policy/main.bicep
johnlokerse: Fixed!

AlexanderSehr (changes requested):
Hey @johnlokerse, I think the module looks good now. Nothing to complain about.

One ask though any chance you could setup the CI on your end so that we can see the pipeline runs successfully through all test cases?

This would require you to 
1. Own an Azure Subscription
2. Set the CI up by following the steps described [here](https://azure.github.io/Azure-Verified-Modules/contributing/bicep/bicep-contribution-flow/#3-configure-your-ci-environment)

If that's not an option for you, let me know. I guess then I'd need to clone your repo and test it on my own ;) 

PS: I don't see a reason why the deployment shouldn't work, the logic looks solid. But the Process demands it ;)

avm/res/key-vault/vault/access-policy/main.bicep
```diff
@@ -6,13 +6,13 @@ metadata owner = 'Azure/module-maintainers'
 param keyVaultName string
```
AlexanderSehr, line 6: Please update the module using the `avm/utilities/Set-AVMModule.ps1` utility. It will update the ReadMes & JSONs for you :)

johnlokerse: Done!

avm/res/key-vault/vault/README.md
```diff
@@ -886,6 +886,79 @@ All access policies to create.
...
+- Type: array
+- Allowed: `[all, backup, delete, get, list, purge, recover, restore, set]`
+
+### Parameter: `accessPolicies.permissions.storage`
+
+Optional. Permissions to storage accounts.
+
+- Required: No
+- Type: array
+- Allowed: `[all, backup, delete, deletesas, get, getsas, list, listsas, purge, recover, regeneratekey, restore, set, setsas, update]`
```
AlexanderSehr, line 952: Sorry if it starts to feel like bullying by know (which is not my intention), but I couldn't help but notice that your fork isn't using the latest changes to the generation script. 
We're working on catching such cases with static tests on our end, and I further suspect the utility won't change much going forward. But for the time being I'd need to ask you to
1. Fetch the lastest changes from upstream via that fancy button in your fork
1. Pull these changes to your local clone
1. Merge them back in the yout branch
1. Run the generation again

You will notice that these values will be shift into a vertical list, thought a bugfix is the more important change.

johnlokerse:
> Hey @johnlokerse, I think the module looks good now. Nothing to complain about.
> 
> One ask though any chance you could setup the CI on your end so that we can see the pipeline runs successfully through all test cases?
> 
> This would require you to
> 
> 1. Own an Azure Subscription
> 2. Set the CI up by following the steps described [here](https://azure.github.io/Azure-Verified-Modules/contributing/bicep/bicep-contribution-flow/#3-configure-your-ci-environment)
> 
> If that's not an option for you, let me know. I guess then I'd need to clone your repo and test it on my own ;)
> 
> PS: I don't see a reason why the deployment shouldn't work, the logic looks solid. But the Process demands it ;)

Green ✅ ! https://github.com/johnlokerse/bicep-registry-modules/actions/runs/6875676191

Took a while because I had some issues with my environment.

avm/res/key-vault/vault/README.md
johnlokerse: Haha, don't worry! Will take a look 👍

johnlokerse: I have updated my fork with the latest changes from the upstream and ran the generation script again, but there are no new changes. Even after deleting the readme's and rerunning the generation script, there are no new changes. The Set-AVMModule function is 1-to-1 on my remote and local feature branch and the main branch.

Am I missing a step?

AlexanderSehr: Hm set PR was merged last week. Let me take a shot at that. Will update you shortly

AlexanderSehr: Shot taken and issue identified. It's an area in the inner `Set-ModuleReadMe` script that for UDTs does not format the allowed & default values as it should (that is, like it's done for top-level parameters). I'll fix this in a separate PR. Sorry for that :)

avm/res/key-vault/vault/access-policy/main.bicep
```diff
@@ -6,13 +6,13 @@ metadata owner = 'Azure/module-maintainers'
...
 
 @description('Optional. An array of 0 to 16 identities that have access to the key vault. All identities in the array must use the same tenant ID as the key vault\'s tenant ID.')
-param accessPolicies array?
+param accessPolicies accessPoliciesType
 
 var formattedAccessPolicies = [for accessPolicy in (accessPolicies ?? []): {
-  applicationId: contains(accessPolicy, 'applicationId') ? accessPolicy.applicationId : ''
-  objectId: contains(accessPolicy, 'objectId') ? accessPolicy.objectId : ''
+  applicationId: accessPolicy.?applicationId ?? ''
+  objectId: accessPolicy.?objectId ?? ''
```
AlexanderSehr, line 13: This is an interesting one @johnlokerse - in the loop you re-use the original logic that treats the `objectId` as optional.
Yet in the UDT you defined it as required. Which is it? :D 

As per the [API Specs](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/keyvault/resource-manager/Microsoft.KeyVault/stable/2023-07-01/keyvault.json#L1063-L1091):
```JSON
"AccessPolicyEntry": {
  "properties": {
    "tenantId": {
      "type": "string",
      "format": "uuid",
      "description": "The Azure Active Directory tenant ID that should be used for authenticating requests to the key vault."
    },
    "objectId": {
      "type": "string",
      "description": "The object ID of a user, service principal or security group in the Azure Active Directory tenant for the vault. The object ID must be unique for the list of access policies."
    },
    "applicationId": {
      "type": "string",
      "format": "uuid",
      "description": " Application ID of the client making request on behalf of a principal"
    },
    "permissions": {
      "$ref": "#/definitions/Permissions",
      "description": "Permissions the identity has for keys, secrets and certificates."
    }
  },
  "description": "An identity that have access to the key vault. All identities in the array must use the same tenant ID as the key vault's tenant ID.",
  "required": [
    "tenantId",
    "objectId",
    "permissions"
  ],
  "type": "object"
}
```
I'd argue the loop must be updated to
```suggestion
  objectId: accessPolicy.objectId
```

AlexanderSehr: Note: I advice to re-run the `Set-AVMModule` utility as this will require an update of the `main.json` files

johnlokerse: Yes, correct 😄... Fixed and ran Set-AVMModule again.

AlexanderSehr (approved):
@johnlokerse thanks for your contribution. Approved it is

johnlokerse:
> @johnlokerse thanks for your contribution. Approved it is

Great! Thanks for the help and looking forward to do more contributions😄

PR #641, opened 2023-11-16
fix:  Removed incorrect nullable ref from kustomizations

AlexanderSehr:
## Description

- Removed incorrect nullable ref from kustomizations
- Regenerated files


eriqua (approved):
🆗

PR #645, opened 2023-11-17
ci: Fix 1mb limit issue for GitHub `getContent` API for module index generation

jtracey93:
## Description

Fix 1mb limit issue for GitHub `getContent` API for module index generation.

Also checked that the new use of `getBlob` supports sizes up to 1mb which will be sufficient for what we need as ARM max is 4mb today

## Testing

Have tested in my fork and a [copy of a trimmed down workflow](https://github.com/jtracey93/bicep-registry-modules/actions/runs/6906771219/job/18792602689#step:5:1) that prints out the `moduleIndex.json` ([gist here](https://gist.github.com/jtracey93/38d61cdd84ceb0403c63fb6e2bbb0fa6))


shenglol (approved):
:shipit:

PR #649, opened 2023-11-18
feat: new module `avm/res/event-grid/domain`

fabmas:
## Description
New event-grid/domain module migrated from CARML to AVM

[avm.res.event-grid.domain](https://github.com/fabmas/bicep-registry-modules/actions/workflows/avm.res.event-grid.domain.yml)


.github/workflows/avm.res.event-grid.domain.yml
```diff
@@ -0,0 +1,81 @@
+name: "avm.res.event-grid.domain"
```
AlexanderSehr, line 1: Hey @fabmas, please note that because this workflow does not exist in main, it won't show. My recommendation (though not perfect) for a general flow is:
1. Create the pipeline in the `main` branch
2. Then create your development branch based on main in which you develop the module 
3. Open your PR from this branch

The only but important caveat of this approach is that if you develop multiple modules at the same time, you have to remove those workflows that don't exist in upstream yet (i.e., the workflows of your other modules you work on) from this dev branch (but not `main`) after creating it.
If you do that, you'll have
- a `main` branch with all workflows
- a `module1` branch with only the `module1` workflow
- a `module2` branch with only the `module2` workflow
- ...

I hope that helps and made some sense. Ain't easy to explain without drawing a picture :D

avm/res/event-grid/domain/tests/e2e/defaults/main.test.bicep
```diff
@@ -0,0 +1,78 @@
...
+        tags: {
+          'hidden-title': 'This is visible in the resource name'
+          Environment: 'Non-Prod'
+          Role: 'DeploymentValidation'
+        }
+      }
+    ]    
+    managedIdentities: {
+      systemAssigned: true
+    }
```
AlexanderSehr, line 72: It appears all those parameters are optional and should hence not be part of the 'min' test (but instead max). I'd hence suggest to remove them and regenerate the readme.
```suggestion
```

avm/res/event-grid/domain/tests/e2e/defaults/dependencies.bicep
```diff
@@ -0,0 +1,60 @@
+@description('Optional. The location to deploy to.')
```
AlexanderSehr, line 1: Note: If most parameters from the `defaults` test file are removed, most if not all of these dependencies may be redundant too. You may actually be able to remove the whole file.

avm/res/event-grid/domain/tests/e2e/waf-aligned/main.test.bicep
```diff
@@ -0,0 +1,124 @@
...
+    managedIdentities: {
+      systemAssigned: true
+    }
+    roleAssignments: [
+      {
+        roleDefinitionIdOrName: 'Reader'
+        principalId: nestedDependencies.outputs.managedIdentityPrincipalId
+        principalType: 'ServicePrincipal'
+      }
+    ]
```
AlexanderSehr, line 110: I'd suggest to remove these as I'm fairly certain one won't need them to deploy the module in a WAF-aligned state. Similar to the PE test case we show ideally only the properties that are required for this test case to not give an incorrect impression (e.g., somebody could think they need a 'reader' role assignment to be WAF-aligned).
Note, if removed, this will require a regeneration of the readme AND may enable you to remove the MSI deployment from the test case's `dependencies.bicep` file.
```suggestion
```

avm/res/event-grid/domain/main.bicep
```diff
@@ -0,0 +1,349 @@
...
+  }[]?
+
+  @description('Optional. The name of logs that will be streamed. "allLogs" includes all possible logs for the resource. Set to \'\' to disable log collection.')
+  metricCategories: {
+    @description('Required. Name of a Diagnostic Metric category for a resource type this setting is applied to. Set to \'AllMetrics\' to collect all metrics.')
+    category: string
+  }[]?
+
+  @description('Optional. A string indicating whether the export to Log Analytics should use the default destination type, i.e. AzureDiagnostics, or use a destination type.')
+  logAnalyticsDestinationType: ('Dedicated' | 'AzureDiagnostics' | null)?
```
AlexanderSehr, line 333:
```suggestion
  logAnalyticsDestinationType: ('Dedicated' | 'AzureDiagnostics')?
```

avm/res/event-grid/domain/main.bicep
```diff
@@ -0,0 +1,349 @@
...
+
+type roleAssignmentType = {
+  @description('Required. The name of the role to assign. If it cannot be found you can specify the role definition ID instead.')
+  roleDefinitionIdOrName: string
+
+  @description('Required. The principal ID of the principal (user/group/identity) to assign the role to.')
+  principalId: string
+
+  @description('Optional. The principal type of the assigned principal ID.')
+  principalType: ('ServicePrincipal' | 'Group' | 'User' | 'ForeignGroup' | 'Device' | null)?
```
AlexanderSehr, line 230:
```suggestion
  principalType: ('ServicePrincipal' | 'Group' | 'User' | 'ForeignGroup' | 'Device')?
```

This should be caught by the recently added test cases. My recommendation:
1. In your fork, disable actions in the settings
2. Pull the latest changes from upstream in (wich will now NOT trigger all your pipelines ;) )
3. Re-enable the pipelines again
4. Re-run at least the static validation for this module on your DEV branch :)

avm/res/event-grid/domain/tests/e2e/defaults/main.test.bicep
fabmas: That was one of my doubts because I understand that "default" and "waf-aligned" tests should be compliant with all the PSRules. In this case if I avoid inserting this part, I receive exceptions in the PSRules

AlexanderSehr: Hey @fabmas, 
then I know the confusion you're referring to. What you're seeing is actually not PSRule complaining about those parameters (or rather their lack of), but outright failing, as there is a [know issue](https://github.com/microsoft/PSRule/issues/1640) in PSRule that is currently being addressed. It's the reason why you can see in many of our test files in AVM that it says something along the lines of 'Workaround for PSRule' where nullable parameters are introduced with a dummy null value to the test file.

E.g.

https://github.com/Azure/bicep-registry-modules/blob/3c5f62204b5ef47d2c396fe43588d355c4a712df/avm/res/batch/batch-account/tests/e2e/defaults/main.test.bicep#L48-L66

AlexanderSehr (approved):

PR #659, opened 2023-11-21
chore: Update CODEOWNERS file

matebarabas:
CODEOWNERS file updated with newly published modules and wildcard line to make sure members of the owners' teams are limited to the scope of their modules.


jtracey93 (approved):
LGTM

PR #673, opened 2023-11-26
feat: Added 'Publish from tag' pipeline for AVM

AlexanderSehr:
## Description

- Added 'Publish from tag' pipeline for AVM
- Updated description of current publishing script

| Pipeline |
| - |
| [Test-Run in fork](https://github.com/AlexanderSehr/bicep-registry-modules/actions/runs/6995186301/job/19029748855) |


.github/workflows/avm.platform.publish.tag.yml
```diff
@@ -0,0 +1,63 @@
+name: "avm.platform.publish.tag"
```
AlexanderSehr, line 1:
```suggestion
name: "avm.platform.publish-tag"
```

.github/workflows/avm.platform.publish.tag.yml
```diff
@@ -0,0 +1,63 @@
+name: "avm.platform.publish-tag"
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "The git tag of the module to publish."
```
eriqua, line 7: Would it be helpful to add an example?

.github/workflows/avm.platform.publish.tag.yml
```diff
@@ -0,0 +1,63 @@
...
+      - name: "Publish tagged module to public bicep registry"
+        uses: azure/powershell@v1
+        with:
+          azPSVersion: "latest"
+          inlineScript: |
+            # Grouping task logs
+            Write-Output '::group::Publish tagged module to public bicep registry'
+
+            # Load used functions
+            . (Join-Path $env:GITHUB_WORKSPACE 'avm' 'utilities' 'pipelines'  'publish' 'Publish-ModuleFromTagToPBR.ps1')
```
eriqua, line 51: Wondering about the location of this script. Should we keep it here or move it to the platform folder? Currently modifying the script would trigger all validation pipelines.

avm/utilities/pipelines/publish/Publish-ModuleFromTagToPBR.ps1
```diff
@@ -1,66 +1,62 @@
...
   # Load used functions
-  . (Join-Path $PSScriptRoot 'helper' 'Get-BRMRepositoryName.ps1')
   . (Join-Path $PSScriptRoot 'helper' 'Get-ModuleReadmeLink.ps1')
-  . (Join-Path (Split-Path $PSScriptRoot -Parent) 'sharedScripts' 'tokenReplacement' 'Convert-TokensInFileList.ps1')
-
-  # TODO: Diff in between tag & tag^-1 to find modules to publish?
 
   # 1. Find tag as per function input
-  $repositoryRoot = (Get-Item $PSScriptRoot).Parent.Parent.Parent.Parent.Parent
+  $repositoryRoot = (Get-Item $PSScriptRoot).Parent.Parent.Parent.Parent.Parent.FullName
```
eriqua, line 34: As per what discussed some time ago, I'd suggest a repositoryroot input parameter with default value, and passing the repo root from the pipeline, leveraging the corrwsponding gh env var

.github/workflows/avm.platform.publish.tag.yml
AlexanderSehr: Maybe, but where? In the title of the input? Or as a comment in the code? 
There is an example in the utility script, and I guess we should add this to the AVM docs too. There is in general a User story open to describe the publishing process where this could be a part of.

eriqua: In the description

.github/workflows/avm.platform.publish.tag.yml
AlexanderSehr: Well it's a script that is run only by a pipeline for publishing. If we'd move it anywhere else I suspect that it would be rather confusing / hard to find afterwards.

However - we could also exclude it from the trigger / find a smart way to exclude specific scripts? Stupid example: if they'd have 'manual' in the name, they would no trigger runs? 

It's a general challenge. For this script in particular it may not matter much as I doubt it will change much (if at all). But the topic itself is something we could brainstorm.

.github/workflows/avm.platform.publish.tag.yml
AlexanderSehr: That will make it very long. But ok, your call

avm/utilities/pipelines/publish/Publish-ModuleFromTagToPBR.ps1
AlexanderSehr: Added to both publish scripts. While I fundamentally think it's a good idea - I'd like to discuss this if we want to do this on scale. On the upside, it allows you to provide whatever path you want in case you want to load dependent functions in isoluation from a custom location.
The downside I'd like to bring up is that we're introducing more and more dependencies on the exact folder structure (outside of e.g. the publishing). I.e., we're using now absolute paths instead of relative ones. If we were to rename e.g. the pipelines folder or something, we'd need to now update a lot more places than before. Not saying we shouldn't do this necessarily, but I'd at least like to question if we're doing us a favor.

AlexanderSehr: But they're not. We're just passing the repository root. If we'd want to make them less dependent, we'd need to pass each function's path.

.github/workflows/avm.platform.publish.tag.yml
eriqua: I'm still inclined to have all platform scripts in the platform folder and to not increase trigger complexity. So sure let's discuss offline.

AlexanderSehr: Fair enough - to me it's not really worth discussing further. I don't agree, but don't care enough either. I moved it now into a `platform` folder to unblock this PR.

avm/utilities/pipelines/publish/Publish-ModuleFromTagToPBR.ps1
eriqua: That is true. It would however result in less effort through search&replace vs wondering how many levels to go up and down in the hierarchy starting from the current script.

AlexanderSehr: That is true.

eriqua:
@AlexanderSehr looks good to me, only prettier between you and merge

AlexanderSehr:
> @AlexanderSehr looks good to me, only prettier between you and merge

My nemesis

eriqua (approved):
:ok:

PR #676, opened 2023-11-28
chore: CODEOWNERS update for AVM

eriqua:
## Description

Uncomment relevant lines to allow corresponding AVM modules to be owned by module owners


matebarabas (approved):
While some of these teams are still yet to be created, their owner have already been notified/reminded, so these changes in the codeowners file are aligned.

PR #678, opened 2023-11-28
fix: Corrected optional PE parameter definition

AlexanderSehr:
## Description

- Updated RequestMessage parameter to be as optional as it should be

| Pipeline |
| - |
| [avm.res.network.private-endpoint](https://github.com/AlexanderSehr/bicep-registry-modules/actions/workflows/avm.res.network.private-endpoint.yml) |


segraef (approved):
lgtm

PR #694, opened 2023-12-03
feat: enable/disable workflows

rahalan:
workflow that can toggle workflows from enabled to disabled state and vice versa


.github/workflows/avm.platform.toggle-avm-workflows.yml
```diff
@@ -0,0 +1,48 @@
...
+        required: true
+        options:
+          - "enable"
+          - "disable"
+        default: "disable"
+      includePattern:
+        type: string
+        description: "RegEx which workflows are included"
+        required: false
+        default: "avm.res"
```
AlexanderSehr, line 18:
```suggestion
        default: "avm\.(?:res|ptn)"
```
Also note that `.` is a special char in regex ;)

.github/workflows/avm.platform.toggle-avm-workflows.yml
```diff
@@ -0,0 +1,48 @@
...
+          fetch-depth: 0
+      - env:
+          GH_TOKEN: ${{ github.token }}
+        name: ${{ inputs.workMode }} AVM workflows
+        shell: pwsh
+        run: |
+            $repo = "${{ github.repository_owner }}/${{ github.event.repository.name }}"
+            $workflows = gh workflow list --repo $repo --all --json 'name,state,id' | ConvertFrom-Json -Depth 100 | Where-Object {$_.name -match "${{ inputs.includePattern }}" -and (-not (($_.name -match "${{ inputs.excludePattern }}") -and "${{ inputs.excludePattern }}".Length -gt 0))}
+
+            foreach ($workflow in $workflows) {
```
AlexanderSehr, line 42:
```suggestion
            $workflows = gh workflow list --repo $repo --all --json 'name,state,id' | ConvertFrom-Json -Depth 100
            $relevantWorkflows = $workflows | Where-Object {
                $_.name -match "${{ inputs.includePattern }}" -and $_.name -notmatch "${{ inputs.excludePattern }}"
            }

            foreach ($workflow in $relevantWorkflows) {
```

.github/workflows/avm.platform.toggle-avm-workflows.yml
```diff
@@ -0,0 +1,48 @@
...
+        options:
+          - "enable"
+          - "disable"
+        default: "disable"
+      includePattern:
+        type: string
+        description: "RegEx which workflows are included"
+        required: false
+        default: "avm.res"
+      excludePattern:
```
AlexanderSehr, line 19: General question. Why not only a pattern? Once can filter to a specific set just with a single regex

.github/workflows/avm.platform.toggle-avm-workflows.yml
AlexanderSehr, line 1: Overall I still can't help but wonder if the logic shouldn't be in a PowerShell script - called by this action (& without GH tokens in it). Then we'd have the best of both worlds.

However, I can also offer to refactor this once the PR was merged in.

.github/workflows/avm.platform.toggle-avm-workflows.yml
rahalan: That was asked in our call, to also have an exclusion pattern

.github/workflows/avm.platform.toggle-avm-workflows.yml
eriqua, line 1: A required check is failing on PR and would block it once approved. The check is complaining about yml formatting, for which you may want to run Prettify. Normally is just indentation.

rahalan: Will do.

.github/workflows/avm.platform.toggle-avm-workflows.yml
AlexanderSehr: Ah well, maybe there is a use case somewhere even though I don't see it right now :D Will resolve

.github/workflows/avm.platform.toggle-avm-workflows.yml
AlexanderSehr: Adding another tested suggestion

AlexanderSehr: Note however, this won't be the only thing prettier complains about. E.g. it really doesn't like single quotes (which is very annoying)

eriqua: I'm normally on the same advice. In this case I agree with Rainer that the script is so limited that it may sound like an overkill. However for consistency sure, we can do that. In that case we need to agree on a proper folder structure to host platform pipeline related scripts. I'd suggest cresting a new avm/utilities/platform so that we won't erroneously trigger all module pipelines whenever we update anything in that folder

AlexanderSehr (approved):

.github/workflows/avm.platform.toggle-avm-workflows.yml
AlexanderSehr: Approving on @eriqua's behalf as the original reason is resolved

PR #699, opened 2023-12-04
feat: avm/res/network/trafficmanagerprofile

lsnoddy:
## Description

Migrate bicep module for trafficmanager from CARML

[avm.res.network.trafficmanagerprofile](https://github.com/lsnoddy/bicep-registry-modules/actions/workflows/avm.res.network.trafficmanagerprofile.yml)


avm/res/network/trafficmanagerprofile/tests/e2e/defaults/main.test.bicep
```diff
@@ -0,0 +1,47 @@
...
+// =================
+resource resourceGroup 'Microsoft.Resources/resourceGroups@2021-04-01' = {
+  name: resourceGroupName
+  location: location
+}
+
+// ============== //
+// Test Execution //
+// ============== //
+var resourceName = '${namePrefix}${serviceShort}001'
```
ChrisSidebotham, line 37: Should this logic not be in the resource deployment directly?

avm/res/network/trafficmanagerprofile/tests/e2e/defaults/main.test.bicep
```diff
@@ -0,0 +1,47 @@
...
+@batchSize(1)
+module testDeployment '../../../main.bicep' = [for iteration in [ 'init', 'idem' ]: {
+  scope: resourceGroup
+  name: '${uniqueString(deployment().name, location)}-test-${serviceShort}-${iteration}'
+  params: {
+    name: resourceName
+    relativeName: resourceName
```
ChrisSidebotham, line 44: Would it be sensible to default this to be the same value as name if null?

avm/res/network/trafficmanagerprofile/main.bicep
```diff
@@ -0,0 +1,246 @@
...
+  'Priority'
+  'Weighted'
+  'Geographic'
+  'MultiValue'
+  'Subnet'
+])
+param trafficRoutingMethod string = 'Performance'
+
+@description('Required. The relative DNS name provided by this Traffic Manager profile. This value is combined with the DNS domain name used by Azure Traffic Manager to form the fully-qualified domain name (FQDN) of the profile.')
+param relativeName string
```
ChrisSidebotham, line 28: Could we make this nullable and then coalesce to name in the resource deployment logic? 

```suggestion
@description('Optional. The relative DNS name provided by this Traffic Manager profile. This value is combined with the DNS domain name used by Azure Traffic Manager to form the fully-qualified domain name (FQDN) of the profile.')
param relativeName string?
```

avm/res/network/trafficmanagerprofile/main.bicep
```diff
@@ -0,0 +1,246 @@
...
+
+resource trafficManagerProfile 'Microsoft.Network/trafficmanagerprofiles@2018-08-01' = {
+  name: name
+  tags: tags
+  location: location
+  properties: {
+    profileStatus: profileStatus
+    trafficRoutingMethod: trafficRoutingMethod
+    dnsConfig: {
+      relativeName: relativeName
```
ChrisSidebotham, line 107: With regards to the above comment on the param:

```suggestion
      relativeName: relativeName ?? name
```

ChrisSidebotham (changes requested):
Please can you also update the CODEOWNERS file as defined [here](https://azure.github.io/Azure-Verified-Modules/specs/shared/#:~:text=bicep%20parent%20team.-,codeowners,-file "CODEOWNERS")

avm/res/network/trafficmanagerprofile/tests/e2e/defaults/main.test.bicep
lsnoddy: I believe it was declared as a variable because it's used twice in name and relative name. But, do you think it would be easier to read if the logic was added to both name and relative name and delete the variable?

avm/res/network/trafficmanagerprofile/main.bicep
lsnoddy: Agreed and implemented.

avm/res/network/trafficmanagerprofile/tests/e2e/defaults/main.test.bicep
lsnoddy: Agreed, made relativeName nullable, so omitting in default and waf test cases and including it in the resource deployment for the max test case. Added a '-rn' on the logic for relativeName in the test case to test the param with a different name than the name param.

ChrisSidebotham (approved):
Nice work 🤌

PR #710, opened 2023-12-08
feat: Reduced publish validation vebosity

AlexanderSehr:
## Description

Given that recent publish actions took around 20 minutes to complete, the amount of output in the log was unnecessary huge ([ref](https://github.com/Azure/bicep-registry-modules/actions/runs/7127028581/job/19406388825)).

I hence removed the default `verbose` flag, updated the time to wait from 5 seconds to 60 seconds and added a retry output.

For example
```
WARNING: Warning: Module [avm/res/service-bus/namespacee] is not in the MCR catalog. Retrying in [60] seconds [1/60]   
WARNING: Warning: Module [avm/res/service-bus/namespacee] is not in the MCR catalog. Retrying in [60] seconds [2/60]   
```


ChrisSidebotham (approved):
Very Nice Sir Alex 🤌

PR #713, opened 2023-12-09
feat: New Module `avm/res/api-management/service`

krbar:
## Description

New Module `avm/res/api-management/service`, migrated from CARML.

## Adding a new module

## Pipeline references

| Pipeline |
| - |
| [avm.res.api-management.service](https://github.com/krbar/bicep-registry-modules/actions/workflows/avm.res.api-management.service.yml) |


avm/res/api-management/service/identity-provider/main.bicep
```diff
@@ -0,0 +1,84 @@
+metadata name = 'API Management Service Identity Providers'
+metadata description = 'This module deploys an API Management Service Identity Provider.'
+metadata owner = 'Azure/module-maintainers'
+
+@description('Conditional. The name of the parent API Management service. Required if the template is used in a standalone deployment.')
+param apiManagementServiceName string
+
+@description('Optional. Used to enable the deployment of the identityProviders child resource.')
+param enableIdentityProviders bool = false
```
AlexanderSehr, line 9: The heck? :D Shouldn't that condition be in its parent, if at all?

avm/res/api-management/service/portalsetting/main.bicep
```diff
@@ -0,0 +1,36 @@
...
+param name string
+
+@description('Optional. Portal setting properties.')
+param properties object = {}
+
+resource service 'Microsoft.ApiManagement/service@2021-08-01' existing = {
+  name: apiManagementServiceName
+}
+
+resource portalSetting 'Microsoft.ApiManagement/service/portalsettings@2021-08-01' = if (!empty(properties)) {
```
AlexanderSehr, line 23: I guess this condition should be part of the parent deployment?

avm/res/api-management/service/tests/e2e/waf-aligned/main.test.bicep
```diff
@@ -0,0 +1,201 @@
...
+        subscriptionRequired: false
+      }
+    ]
+    subscriptions: [
+      {
+        name: 'testArmSubscriptionAllApis'
+        scope: '/apis'
+      }
+    ]
+    managedIdentities: {
```
AlexanderSehr, line 189: Requird for WAF-alignment? :D If not - can be removed alongside the dependency (if I'm not mistaken)

avm/res/api-management/service/main.bicep
```diff
@@ -0,0 +1,526 @@
...
+  @description('Required. The principal ID of the principal (user/group/identity) to assign the role to.')
+  principalId: string
+
+  @description('Optional. The principal type of the assigned principal ID.')
+  principalType: ('ServicePrincipal' | 'Group' | 'User' | 'ForeignGroup' | 'Device')?
+
+  @description('Optional. The description of the role assignment.')
+  description: string?
+
+  @description('Optional. The conditions on the role assignment. This limits the resources it can be assigned to. e.g.: @Resource[Microsoft.Storage/storageAccounts/blobServices/containers:ContainerName] StringEqualsIgnoreCase "foo_storage_container"')
```
AlexanderSehr, line 480:
```suggestion
  @description('Optional. The conditions on the role assignment. This limits the resources it can be assigned to. e.g.: @Resource[Microsoft.Storage/storageAccounts/blobServices/containers:ContainerName] StringEqualsIgnoreCase "foo_storage_container".')
```

avm/res/api-management/service/portalsetting/main.bicep
krbar: Moved it there and made the properties param of the child module required.

avm/res/api-management/service/identity-provider/main.bicep
krbar: In fact, an interesting implementation. I removed this param and created a condition in the parent.

avm/res/api-management/service/tests/e2e/waf-aligned/main.test.bicep
krbar: Removed user-assigned and the dependency, the result of the test will be known in the morning... :)

krbar:
@AlexanderSehr There is an error in the PSRule validation for the waf-aligned https://github.com/krbar/bicep-registry-modules/actions/runs/7161001234. It looks like the script cannot deal with the parameter `policies`, whose property `value` in fact contains some untypical characters. 

`Error: Failed to expand bicep source '/home/runner/work/bicep-registry-modules/bicep-registry-modules/avm/res/api-management/service/tests/e2e/waf-aligned/main.test.bicep'. Exception calling "GetBicepResources" with "2" argument(s): "Unable to expand resources because the source file '/home/runner/work/bicep-registry-modules/bicep-registry-modules/avm/res/api-management/service/tests/e2e/waf-aligned/main.test.bicep' was not valid. An error occurred evaluating expression '[length(parameters('policies'))]' line 1350. Object reference not set to an instance of an object."`

Anything I or someone else can do here?

avm/res/api-management/service/tests/e2e/waf-aligned/main.test.bicep
AlexanderSehr: ok, but is the property required in the first place is what I'm wondering :D Given that it is optional.

avm/res/api-management/service/identity-provider/main.bicep
AlexanderSehr: For some odd reason that didn't work which lets me to believe that the tast case is incorrect / some properties that should be mandatory are not. The only case that I could image where this is not the case is if the `identityProviders` parameter is used also for another module deployment - which again would be very odd.

AlexanderSehr:
> @AlexanderSehr There is an error in the PSRule validation for the waf-aligned https://github.com/krbar/bicep-registry-modules/actions/runs/7161001234. It looks like the script cannot deal with the parameter `policies`, whose property `value` in fact contains some untypical characters.
> 
> `Error: Failed to expand bicep source '/home/runner/work/bicep-registry-modules/bicep-registry-modules/avm/res/api-management/service/tests/e2e/waf-aligned/main.test.bicep'. Exception calling "GetBicepResources" with "2" argument(s): "Unable to expand resources because the source file '/home/runner/work/bicep-registry-modules/bicep-registry-modules/avm/res/api-management/service/tests/e2e/waf-aligned/main.test.bicep' was not valid. An error occurred evaluating expression '[length(parameters('policies'))]' line 1350. Object reference not set to an instance of an object."`
> 
> Anything I or someone else can do here?

Does not seem that odd to me - it's just happends to be XML. Why PSRule would fail I don't really understand. @BernieWhite would you happen to have an idea?

So that you don't have to search, the parameter in the [`waf-aligned/main.test.bicep` file](https://github.com/krbar/bicep-registry-modules/blob/c7397505336673a143857f2f77fb9666e6443efb/avm/res/api-management/service/tests/e2e/waf-aligned/main.test.bicep#L133-L138) is
```Bicep
policies: [
  {
    format: 'xml'
    value: '<policies> <inbound> <rate-limit-by-key calls=\'250\' renewal-period=\'60\' counter-key=\'@(context.Request.IpAddress)\' /> </inbound> <backend> <forward-request /> </backend> <outbound> </outbound> </policies>'
  }
]
```

The reference in the [`main.bicep` file](https://github.com/krbar/bicep-registry-modules/blob/users/krbar/apiMgmtModule/avm/res/api-management/service/main.bicep#L340-L347) is
```Bicep
@description('Optional. Policies.')
param policies array = []

(...)

module service_policies 'policy/main.bicep' = [for (policy, index) in policies: {
  name: '${uniqueString(deployment().name, location)}-Apim-Policy-${index}'
  params: {
    apiManagementServiceName: service.name
    value: policy.value
    format: contains(policy, 'format') ? policy.format : 'xml'
  }
}]
```

The [error message](https://github.com/krbar/bicep-registry-modules/actions/runs/7161001234/job/19496002229#step:4:224) is
```
Error: Failed to expand bicep source '/home/runner/work/bicep-registry-modules/bicep-registry-modules/avm/res/api-management/service/tests/e2e/waf-aligned/main.test.bicep'. Exception calling "GetBicepResources" with "2" argument(s): "Unable to expand resources because the source file '/home/runner/work/bicep-registry-modules/bicep-registry-modules/avm/res/api-management/service/tests/e2e/waf-aligned/main.test.bicep' was not valid. An error occurred evaluating expression '[length(parameters('policies'))]' line 1350. Object reference not set to an instance of an object."
```

Can't say that anything stands out to me.

BernieWhite:
@AlexanderSehr Agreed, seems fine. Although we just fixed a dependency issue in the latest prerelease. So you could try that. Happy to have a look at it though.

avm/res/api-management/service/identity-provider/main.bicep
krbar: This parameter with the default value `false` actually resulted in all IdentityProviders tests being deactivated (which was perhaps even the intention: implement IdentityProviders, but left them out of the test due to the strong integration with AAD). 
But I found out that deploying Identity providers also works with fake values, so I've re-enabled them in the test cases (see [02bfcec](https://github.com/Azure/bicep-registry-modules/pull/713/commits/02bfcecdceadfd3c914f724beb041dba65972cca))

avm/res/api-management/service/tests/e2e/waf-aligned/main.test.bicep
AlexanderSehr: We actually had a similar rule for CognitiveServices and opted to exclude it. I'd suggest the same.

https://github.com/Azure/bicep-registry-modules/blob/8ab1e4d6a26b8ab4d38df0954134c667c57b0ee4/avm/utilities/pipelines/staticValidation/psrule/.ps-rule/min-suppress.Rule.yaml#L14-L15

avm/res/api-management/service/main.bicep
```diff
@@ -0,0 +1,525 @@
...
+    apiRevisionDescription: api.?apiRevisionDescription
+    apiType: api.?apiType
+    apiVersion: api.?apiVersion
+    apiVersionDescription: api.?apiVersionDescription
+    apiVersionSetId: api.?apiVersionSetId
+    authenticationSettings: api.?authenticationSettings
+    format: api.?format ?? 'openapi'
+    isCurrent: api.?isCurrent
+    protocols: api.?protocols
+    policies: api.?policies
```
BernieWhite, line 219: @AlexanderSehr The issue is related to a null being passed into a copy loop within `api/main.bicep` which loops for policies.

Need to check on the behaviour of `length` with null.

BernieWhite: @krbar @AlexanderSehr Ok. The error is not related to the `policies` parameter in `main.bicep` for the APIM service. It is related to a nested module under `api/main.bicep` which also has a `policies` parameter.

```bicep
param policies array = []
```

As a result, the `for` loop for configuring policies under the API is using a `null` in the `length` function of the expanded template which is failing.

From what I can tell similarly in ARM `length` would fail on a `null` argument. As a result, I think this issue is desired. But I admit we can improve the reporting message in PSRule, which I'll log a card for.

But happy for further feedback/ suggestions or to be proven wrong.

avm/res/api-management/service/tests/e2e/waf-aligned/main.test.bicep
BernieWhite: Wouldn't a managed identity of some kind be required for using key vault which is where all secrets would be stored?

avm/res/api-management/service/main.bicep
krbar: @BernieWhite you are right, I came to the same conclusion and updated the implementation of this parameter. Now PSRule shows the results properly and the template also deploys without issues. Thank you for investigating it!

avm/res/api-management/service/tests/e2e/waf-aligned/main.test.bicep
AlexanderSehr: If referencing secrets from there absolutely, yes. In this deployment I however cannot see any key vault references. 

I guess the `clientSecret` could be such a candidate, where a secret could be pulled. But if a user where not to deploy anything that requires a secret the question would still remain why they should deploy a MI.

Overall my take on this and similar policies is: If a service accesses anything it should always use a managed identity. But if it doesn't, it should not have an identity for the sake of having one. If that makes sense :D

krbar:
@AlexanderSehr I think the module's issues are sorted out, please have a look if we can merge the module or if still some updated are required. Thanks!

AlexanderSehr:
> @AlexanderSehr I think the module's issues are sorted out, please have a look if we can merge the module or if still some updated are required. Thanks!

Just one comment left to get convinced on ;)

AlexanderSehr (approved):

AlexanderSehr (approved):

PR #718, opened 2023-12-11
fix: Remove `managedBy`

segraef:
[avm.res.resources.resource-group](https://github.com/segraef/bicep-registry-modules/actions/workflows/avm.res.resources.resource-group.yml)

Fix for https://github.com/Azure/ResourceModules/pull/4351


ChrisSidebotham (commented):
I am happy with this and will pend for @AlexanderSehr approval after our discussion this morning

coolhome:
I'm not sure if it's a good idea to just omit the property. This leaves early adaptors unable to use this module until they delete the resource group since we all been creating them with empty strings. 

I don't know if producers of Managed Application use AVM but I would imagine its useful.

segraef:
> I'm not sure if it's a good idea to just omit the property. This leaves early adaptors unable to use this module until they delete the resource group since we all been creating them with empty strings.
> 
> I don't know if producers of Managed Application use AVM but I would imagine its useful.

They still can use `0.1`. I'm a bit hesitant to put in more logic than necessary.
But if we agree to add your suggested [logic](https://github.com/Azure/ResourceModules/blob/ec9a93c8da2728085e2e3ee48ecd903a65fa9b3c/modules/resources/resource-group/main.bicep) to support both cases (rg and managed rg) then I'd be also fine.

Let's see what the team says.

AlexanderSehr (approved):

AlexanderSehr:
> > I'm not sure if it's a good idea to just omit the property. This leaves early adaptors unable to use this module until they delete the resource group since we all been creating them with empty strings.
> > 
> > I don't know if producers of Managed Application use AVM but I would imagine its useful.
> 
> They still can use `0.1`. I'm a bit hesitant to put in more logic than necessary.
> But if we agree to add your suggested [logic](https://github.com/Azure/ResourceModules/blob/ec9a93c8da2728085e2e3ee48ecd903a65fa9b3c/modules/resources/resource-group/main.bicep) to support both cases (rg and managed rg) then I'd be also fine.
> 
> Let's see what the team says, @AlexanderSehr, @ChrisSidebotham.

It's an issue but one that only gets worse with time. And while the module isn't 1.0 yet it might be better to make that call now rather than later 😶

PR #700, opened 2023-12-05
feat: New Module Web/Serverfarms CARML Migration

tsc-buddy:
## Description

CARML migration for Microsoft.Web/serverfarms resource module.

- [x] Tested deployments locally
- [x] Tested deployments via Workflow (just pulled the latest PSRule updates also)
- [x] Stepped through the CARML migration Issue guide.

Note if you are testing deployments to an MCAPS internal subscription you may face quota issues in the default region used by the workflow template (West Europe, I believe).

@ChrisSidebotham , let us know if there is anything I've missed. If its all good, I shall move onto the other Microsoft.Web modules. :)

Workflow Badge Status: Failing only on the publishing task.

[avm.res.web.serverfarm](https://github.com/tsc-buddy/bicep-registry-modules/actions/workflows/avm.res.web.serverfarm.yml)

PR #724, opened 2023-12-12
fix: updating location for capacity constraints

ChrisSidebotham:
## Description

Following PR #700. Capacity in WestEurope is running low for compute. Qouta limits have been imposed on non priority subscriptions ultimately causing the deployment test to fail. Deployment Test has now been changed to North Europe. And the subscription Quota should be checked before merging.

| Pipeline | 
| -------- | 
| [avm.res.web.serverfarm](https://github.com/Azure/bicep-registry-modules/actions/workflows/avm.res.web.serverfarm.yml) |

### DIFF
--- a/avm/res/web/serverfarm/README.md
+++ b/avm/res/web/serverfarm/README.md
@@ -54,7 +54,7 @@ module serverfarm 'br/public:avm/res/web/serverfarm:<version>' = {
       tier: 'Premium'
     }
     // Non-required parameters
-    location: '<location>'
+    location: 'eastus'
   }
 }
 ```
@@ -86,7 +86,7 @@ module serverfarm 'br/public:avm/res/web/serverfarm:<version>' = {
     },
     // Non-required parameters
     "location": {
-      "value": "<location>"
+      "value": "eastus"
     }
   }
 }
@@ -133,7 +133,7 @@ module serverfarm 'br/public:avm/res/web/serverfarm:<version>' = {
       }
     ]
     kind: 'App'
-    location: '<location>'
+    location: 'eastus'
     lock: {
       kind: 'CanNotDelete'
       name: 'lock'
@@ -212,7 +212,7 @@ module serverfarm 'br/public:avm/res/web/serverfarm:<version>' = {
       "value": "App"
     },
     "location": {
-      "value": "<location>"
+      "value": "eastus"
     },
     "lock": {
       "value": {
@@ -297,7 +297,7 @@ module serverfarm 'br/public:avm/res/web/serverfarm:<version>' = {
       }
     ]
     kind: 'App'
-    location: '<location>'
+    location: 'eastus'
     lock: {
       kind: 'CanNotDelete'
       name: 'lock'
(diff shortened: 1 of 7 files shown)


avm/res/web/serverfarm/main.bicep
```diff
@@ -110,7 +110,7 @@ resource avmTelemetry 'Microsoft.Resources/deployments@2023-07-01' = if (enableT
 resource appServicePlan 'Microsoft.Web/serverfarms@2022-09-01' = {
   name: name
   kind: kind
-  location: location
+  location: 'eastus'
```
eriqua, line 113: Is it required to be hardcoded in the template?

ChrisSidebotham: After discussion with Alex it is due to the parameter being over written in the pipeline - this current change just needs to be to be a per module change and therefore must be hardcoded to the location value

ChrisSidebotham: Updated! fixed `main.bicep`

avm/res/web/serverfarm/tests/e2e/max/dependencies.bicep
```diff
@@ -6,7 +6,7 @@ param managedIdentityName string
 
 resource managedIdentity 'Microsoft.ManagedIdentity/userAssignedIdentities@2018-11-30' = {
   name: managedIdentityName
-  location: location
+  location: 'eastus'
```
AlexanderSehr, line 9: why? Does it have to be in the same region as the server farm?

avm/res/web/serverfarm/README.md
```diff
@@ -54,7 +54,7 @@ module serverfarm 'br/public:avm/res/web/serverfarm:<version>' = {
       tier: 'Premium'
     }
     // Non-required parameters
-    location: '<location>'
+    location: 'eastus'
```
AlexanderSehr, line 57: Needs updating

avm/res/web/serverfarm/tests/e2e/max/dependencies.bicep
ChrisSidebotham: Surely keeping all of the resources together would be best for now so we can unblock the CI Env?

avm/res/web/serverfarm/README.md
ChrisSidebotham: This is generated from the test files what needs updating here?

AlexanderSehr: Just testing you!

avm/res/web/serverfarm/tests/e2e/max/dependencies.bicep
AlexanderSehr: in theory yes - but if we start also changing the dependency files, cleaning this up later will become a lot harder (also harder to troubleshoot if we overlook a location).

AlexanderSehr: In that case I'd rather suggest to hardcode it in the invocating deployment - i.e. in the `main.test.bicep`. Then the workaround is limited to one location

AlexanderSehr (changes requested):

avm/res/web/serverfarm/tests/e2e/max/dependencies.bicep
ChrisSidebotham: updated

AlexanderSehr (approved):

PR #727, opened 2023-12-13
feat: Added matrix name to template increasing readability

AlexanderSehr:
## Description

Added matrix name to template increasing readability

Before:
image

After:
image


.github/workflows/avm.res.insights.action-group.yml
```diff
@@ -71,7 +71,7 @@ jobs:
   #   Call reusable workflow   #
   ##############################
   call-workflow-passing-data:
-    name: "Module"
+    name: "Run"
```
eriqua, line 74: Shouldn't we have this change in all caller workflows?

eriqua (changes requested):
Great improvement on readability, both for the compressed test case and replacing `Module` with `Run`. 

The only update I'd challenge is the matrix header. Can we keep `PSRule` and `Deploy` instead of `job_test_psrule` and `job_module_deploy`?

.github/workflows/avm.res.insights.action-group.yml
AlexanderSehr: Yes

AlexanderSehr:
> Great improvement on readability, both for the compressed test case and replacing `Module` with `Run`.
> 
> The only udate I'd challenge is the matrix header. Can we keep `PSRule` and `Deploy` instead of `job_test_psrule` and `job_module_deploy`?

@eriqua Unfortunately not. The change was not intended and seems to be automatically done once you add a name to the matrix job. The name is then used for the jobs while the Id is used for the matrix header. I did not find a way to avoid this 😕

eriqua:
> > Great improvement on readability, both for the compressed test case and replacing `Module` with `Run`.
> > The only udate I'd challenge is the matrix header. Can we keep `PSRule` and `Deploy` instead of `job_test_psrule` and `job_module_deploy`?
> 
> @eriqua Unfortunately not. The change was not intended and seems to be automatically done once you add a name to the matrix job. The name is then used for the jobs while the 

Heartbreaking

.github/workflows/avm.res.insights.action-group.yml
AlexanderSehr: done

eriqua (approved):
:ok:

PR #732, opened 2023-12-15
fix: `avm/res/web/serverfarm` 

ChrisSidebotham:
## Description

Updated description to enable publishing

| Pipeline | 
| -------- | 
| [avm.res.web.serverfarm](https://github.com/Azure/bicep-registry-modules/actions/workflows/avm.res.web.serverfarm.yml) |


eriqua (approved):
🆗

PR #749, opened 2023-12-27
feat: New Module `compute/disk-encryption-set`

krbar:
## Description

New Module `compute/disk-encryption-set`, migrated from CARML.

## Adding a new module

## Pipeline references

| Pipeline |
| - |
| [avm.res.compute.disk-encryption-set](https://github.com/krbar/bicep-registry-modules/actions/workflows/avm.res.compute.disk-encryption-set.yml) |


avm/res/compute/disk-encryption-set/tests/e2e/waf-aligned/main.test.bicep
```diff
@@ -0,0 +1,72 @@
...
+  params: {
+    name: '${namePrefix}${serviceShort}001'
+    location: location
+    keyName: nestedDependencies.outputs.keyName
+    keyVaultResourceId: nestedDependencies.outputs.keyVaultResourceId
+    managedIdentities: {
+      userAssignedResourceIds: [
+        nestedDependencies.outputs.managedIdentityResourceId
+      ]
+    }
```
AlexanderSehr, line 65: Required?

avm/res/compute/disk-encryption-set/tests/e2e/accessPolicies/main.test.bicep
AlexanderSehr, line 1: I take it this is originating from CARML. Would you see any value in the test? The access policies are set on the KeyVault which is out of scope for the of the module - hence the test looks 1:1 the same. Also actually no AccessPolicies are set in the KeyVault which begs the question if the test actually works (i.e., not just deploys). I'd recomend to remove it.

avm/res/compute/disk-encryption-set/modules/nested_keyVaultPermissions.bicep
```diff
@@ -0,0 +1,69 @@
...
+
+resource keyVault 'Microsoft.KeyVault/vaults@2021-10-01' existing = {
+  name: last(split(keyVaultResourceId, '/'))!
+
+  resource key 'keys@2021-10-01' existing = {
+    name: keyName
+  }
+}
+
+module userAssignedIdentity 'nested_managedIdentityReference.bicep' = {
```
AlexanderSehr, line 24: Can this not just be an existing resource? Straight in this file?

avm/res/compute/disk-encryption-set/tests/e2e/accessPolicies/main.test.bicep
```diff
@@ -0,0 +1,90 @@
+targetScope = 'subscription'
+
+metadata name = 'Using Key Vault with the Access Policy permission model'
+metadata description = 'This instance uses a Key Vault with the Access Policy permission model.'
```
AlexanderSehr, line 4:
```suggestion
metadata description = 'This instance uses a Key Vault with the Access Policy permission model. If no permissions are set, the module attempts to add the permissions for you.'
```

avm/res/compute/disk-encryption-set/main.bicep
```diff
@@ -0,0 +1,217 @@
+metadata name = 'Disk Encryption Sets'
+metadata description = 'This module deploys a Disk Encryption Set.'
```
AlexanderSehr, line 2:
```suggestion
metadata description = 'This module deploys a Disk Encryption Set. The module will attempt to set permissions on the provided Key Vault for any used user-assigned identity.'
```

avm/res/compute/disk-encryption-set/modules/nested_keyVaultPermissions.bicep
```diff
@@ -0,0 +1,69 @@
...
+@description('Required. The resourceID of the User Assigned Identity to assign permissions to.')
+param userAssignedIdentityResourceId string
+
+@description('Optional. Resource location.')
+param location string = resourceGroup().location
+
+@description('Required. Resource ID of the KeyVault containing the key or secret.')
+param keyVaultResourceId string
+
+@description('Required. Key URL (with version) pointing to a key or secret in KeyVault.')
```
AlexanderSehr, line 13: I'd suggest to add a parameter for the image version (similar to the interface for CMK) so that we can fetch the image version for the user if not provided

avm/res/compute/disk-encryption-set/main.bicep
```diff
@@ -0,0 +1,217 @@
...
+}
+
+// Note: This is only enabled for user-assigned identities as the service's system-assigned identity isn't available during its initial deployment
+module keyVaultPermissions 'modules/nested_keyVaultPermissions.bicep' = [for (userAssignedIdentityResourceId, index) in (managedIdentities.?userAssignedResourceIds ?? []): {
+  name: '${uniqueString(deployment().name, location)}-DiskEncrSet-KVPermissions-${index}'
+  params: {
+    keyName: keyName
+    keyVaultResourceId: keyVaultResourceId
+    userAssignedIdentityResourceId: userAssignedIdentityResourceId
+    rbacAuthorizationEnabled: keyVault.properties.enableRbacAuthorization
```
AlexanderSehr, line 105: Redundant parameter as we have the key vault reference in the referenced module anyways

avm/res/compute/disk-encryption-set/modules/nested_keyVaultPermissions.bicep
```diff
@@ -0,0 +1,69 @@
...
+    location: location
+  }
+  scope: resourceGroup(split(userAssignedIdentityResourceId, '/')[2], split(userAssignedIdentityResourceId, '/')[4])
+}
+
+// =============== //
+// Role Assignment //
+// =============== //
+
+resource keyVaultKeyRBAC 'Microsoft.Authorization/roleAssignments@2022-04-01' = if (rbacAuthorizationEnabled == true) {
```
AlexanderSehr, line 37:
```suggestion
resource keyVaultKeyRBAC 'Microsoft.Authorization/roleAssignments@2022-04-01' = if (keyVault.properties.enableRbacAuthorization == true) {
```

avm/res/compute/disk-encryption-set/modules/nested_keyVaultPermissions.bicep
```diff
@@ -0,0 +1,69 @@
...
+    roleDefinitionId: subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '12338af0-0e69-4776-bea7-57ae8d297424') // Key Vault Crypto User
+    principalType: 'ServicePrincipal'
+  }
+}
+
+// ============= //
+// Access Policy //
+// ============= //
+
+module keyVaultAccessPolicies '../../../key-vault/vault/access-policy/main.bicep' = if (rbacAuthorizationEnabled != true) {
```
AlexanderSehr, line 51:
```suggestion
module keyVaultAccessPolicies '../../../key-vault/vault/access-policy/main.bicep' = if (keyVault.properties.enableRbacAuthorization != true) {
```

avm/res/compute/disk-encryption-set/modules/nested_keyVaultPermissions.bicep
```diff
@@ -0,0 +1,69 @@
+@description('Required. A boolean to specify whether or not the used Key Vault has RBAC authentication enabled or not.')
+param rbacAuthorizationEnabled bool = true
+
```
AlexanderSehr, line 3:
```suggestion
```

avm/res/compute/disk-encryption-set/modules/nested_keyVaultPermissions.bicep
```diff
@@ -0,0 +1,69 @@
...
+param location string = resourceGroup().location
+
+@description('Required. Resource ID of the KeyVault containing the key or secret.')
+param keyVaultResourceId string
+
+@description('Required. Key URL (with version) pointing to a key or secret in KeyVault.')
+param keyName string
+
+resource keyVault 'Microsoft.KeyVault/vaults@2021-10-01' existing = {
+  name: last(split(keyVaultResourceId, '/'))!
```
AlexanderSehr, line 17: May require a scope statement in case the key vault is in a different RG than the disk encryption set

avm/res/compute/disk-encryption-set/modules/nested_keyVaultPermissions.bicep
krbar: Unfortunately, this doesn't work and we probably need to keep the `rbacAuthorizationEnabled` parameter. 
The reason:
`This expression is being used in the if-condition expression, which requires a value that can be calculated at the start of the deployment.`

avm/res/compute/disk-encryption-set/modules/nested_keyVaultPermissions.bicep
krbar: Same as below: 
Unfortunately, this doesn't work and we probably need to keep the `rbacAuthorizationEnabled` parameter. 
The reason:
`This expression is being used in the if-condition expression, which requires a value that can be calculated at the start of the deployment.`

avm/res/compute/disk-encryption-set/main.bicep
```diff
@@ -0,0 +1,217 @@
...
+  @description('Required. The principal ID of the principal (user/group/identity) to assign the role to.')
+  principalId: string
+
+  @description('Optional. The principal type of the assigned principal ID.')
+  principalType: ('ServicePrincipal' | 'Group' | 'User' | 'ForeignGroup' | 'Device')?
+
+  @description('Optional. The description of the role assignment.')
+  description: string?
+
+  @description('Optional. The conditions on the role assignment. This limits the resources it can be assigned to. e.g.: @Resource[Microsoft.Storage/storageAccounts/blobServices/containers:ContainerName] StringEqualsIgnoreCase "foo_storage_container"')
```
AlexanderSehr, line 209:
```suggestion
  @description('Optional. The conditions on the role assignment. This limits the resources it can be assigned to. e.g.: @Resource[Microsoft.Storage/storageAccounts/blobServices/containers:ContainerName] StringEqualsIgnoreCase "foo_storage_container".')
```

avm/res/compute/disk-encryption-set/modules/nested_keyVaultPermissions.bicep
krbar: As soon as I add the scope, the deployment of RBAC complains about its scope:
`A resource's scope must match the scope of the Bicep file for it to be deployable. You must use modules to deploy resources to a different scope.` 
I will try to catch you to discuss it.

AlexanderSehr: Touche. In that case, the scope would need to be set upon referencing this template in the `disk-encryption-set/main.bicep` file.

avm/res/compute/disk-encryption-set/modules/nested_keyVaultPermissions.bicep
krbar: Yes, it can. Implemented:
- https://github.com/Azure/bicep-registry-modules/pull/749/commits/2e20017846bb85638d18183b71051fbe7f286dc2
- https://github.com/Azure/bicep-registry-modules/pull/749/commits/d61f2a4698e61d655f3055c7e49966920db4f0c5

avm/res/compute/disk-encryption-set/modules/nested_keyVaultPermissions.bicep
```diff
@@ -0,0 +1,66 @@
...
+    roleDefinitionId: subscriptionResourceId('Microsoft.Authorization/roleDefinitions', '12338af0-0e69-4776-bea7-57ae8d297424') // Key Vault Crypto User
+    principalType: 'ServicePrincipal'
+  }
+}
+
+// ============= //
+// Access Policy //
+// ============= //
+
+module keyVaultAccessPolicies '../../../key-vault/vault/access-policy/main.bicep' = if (rbacAuthorizationEnabled != true) {
```
krbar, line 48: @AlexanderSehr should we actually reference this module from the public registry?

avm/res/compute/disk-encryption-set/modules/nested_keyVaultPermissions.bicep
krbar: I think it's already set properly there:

`scope: resourceGroup(split(keyVaultResourceId, '/')[2], split(keyVaultResourceId, '/')[4])`

avm/res/compute/disk-encryption-set/modules/nested_keyVaultPermissions.bicep
AlexanderSehr: No, not at this time. While we fortunately kept the door open for that possibility, it's not yet a priority to publish child-modules.

That said, you'd need to do what @rahalan did in the VM PR where he referenced the protected item child module of the RSV module. That is, to add a copy of the accessPolicy `main.bicep` as e.g., a `key-vault.vault.access-policies.bicep` (or whatever you want to call it) to a subfolder `modules` in the `disk-encryption-set` module. So that would look like:

- disk-encryption-set
  - modules
    - `key-vault.vault.access-policies.bicep` (contains the copied code) 
  - `main.bicep` (references `key-vault.vault.access-policies.bicep`)

It's the same option contributors have for, for example, the Private Endpoint module. You don't 'have to' reference the published module but could instead have your own copy if you want.

avm/res/compute/disk-encryption-set/modules/nested_keyVaultPermissions.bicep
AlexanderSehr: True, but just to double check:

- The `main.bicep` sets the scope for this file here to possibly a different resource group (which is as intended)
- And then we set yet another scope in the below `userAssignedIdentity` reference as the MSI could (of course) be in yet another resource group.

The only thing that strikes me as a bit odd (but maybe that's just the way it is) is, that one 'existing' resource allows to set a scope (userAssignedIdentity) and one doesn't (keyVault) :D

AlexanderSehr (changes requested):

avm/res/compute/disk-encryption-set/modules/nested_keyVaultPermissions.bicep
krbar: done: https://github.com/Azure/bicep-registry-modules/pull/749/commits/d1f1ae4e43de6ae80e82c001123fbfe7df8a7c5d

avm/res/compute/disk-encryption-set/modules/nested_keyVaultPermissions.bicep
krbar: It is already implemented in the `main.bicep`. This nested bicep module only sets the permissions and we don't need the key version here as the permissions are set on the Key Vault level (access policy model) or on key level (RBAC model), but not for the particular key version,

avm/res/compute/disk-encryption-set/main.bicep
krbar: See my other (resolved) comments: 
We actually need to keep the `rbacAuthorizationEnabled` parameter. If we attempt to calculate this value at run time, we get the error:
`This expression is being used in the if-condition expression, which requires a value that can be calculated at the start of the deployment.`

avm/res/compute/disk-encryption-set/modules/nested_keyVaultPermissions.bicep
krbar: But I have updated the description of this parameter, as it was referring to an URL: https://github.com/Azure/bicep-registry-modules/pull/749/commits/ccb04d56f6138cd91c6bea79ff8aa47f80844c2f

avm/res/compute/disk-encryption-set/main.bicep
AlexanderSehr: How unfortunate. Alright, thanks

avm/res/compute/disk-encryption-set/modules/nested_keyVaultPermissions.bicep
AlexanderSehr: That certainly helps. Thanks

AlexanderSehr (approved):

PR #760, opened 2024-01-02
fix: Enable UDT schema tests to differentiate modules that uses the `sys` namespace explicitely

AlexanderSehr:
## Description

Some modules (like Compute Gallery) must use metadata annotations using the `sys` namespace (e.g., @sys.description('...'). Up until now, when testing the implementation against the schema, Pester would complain about those cases. This was fixed by detecting if the `sys` namespace was used and if so, updating the expected schema.

Before:
image

After:
image


ChrisSidebotham (approved):
Looking Good Mr Sehr!

PR #767, opened 2024-01-04
feat: New Module `avm/res/network/firewall-policy`

krbar:
## Description

New Module `avm/res/network/firewall-policy`, migrated from CARML.

## Adding a new module

## Pipeline references

| Pipeline |
| - |
| [avm.res.network.firewall-policy](https://github.com/krbar/bicep-registry-modules/actions/workflows/avm.res.network.firewall-policy.yml) |


avm/res/network/firewall-policy/main.bicep
```diff
@@ -0,0 +1,212 @@
...
+@description('Optional. Enable/Disable usage telemetry for module.')
+param enableTelemetry bool = true
+
+@description('Optional. Rule collection groups.')
+param ruleCollectionGroups array?
+
+var formattedUserAssignedIdentities = reduce(map((managedIdentities.?userAssignedResourceIds ?? []), (id) => { '${id}': {} }), {}, (cur, next) => union(cur, next)) // Converts the flat array to an object like { '${id1}': {}, '${id2}': {} }
+
+var identity = !empty(managedIdentities) ? {
+  type: !empty(managedIdentities.?userAssignedResourceIds ?? {}) ? 'UserAssigned' : null
```
AlexanderSehr, line 102:
```suggestion
  type: 'UserAssigned'
```

I guess, this will be good enough

avm/res/network/firewall-policy/tests/e2e/max/main.test.bicep
```diff
@@ -0,0 +1,91 @@
...
+  name: resourceGroupName
+  location: location
+}
+
+// ============== //
+// Test Execution //
+// ============== //
+
+@batchSize(1)
+module testDeployment '../../../main.bicep' = [for iteration in [ 'init', 'idem' ]: {
```
AlexanderSehr, line 39: Should contain an MSI test I pressume

krbar: I agree, however the Managed Identity is part of the TLS Inspection, which is a Premium feature. 
We could switch the the `max` test case to Premium tier. 
Actually it would be even better to have a separate test case with the Premium features, but the TLS Inspection requires a certificate with particular properties, which is not that easy to generate 'on-the-fly' for test purposes.
Also for now, as a part of the module migration I would propose to just test the MSI using the `max` test case and Premium tier. And the full test of all Premium features like TLS Inspection or IDPS (intrusion detection and prevention system) could be implemented later. What do you think?

krbar: Added: https://github.com/Azure/bicep-registry-modules/pull/767/commits/afdce4a052437c35896c01953d5742aff4b12c27

ChrisSidebotham: Good Effort @krbar - This is also similar for Service Bus we used the Max test to test the premium features. I would state I would only expect to see a separate test if there is entirely different parameter set functionality between the SKU's e.g

ChrisSidebotham (approved):
Looks good and thanks for taking the last changes for the read me :D

PR #775, opened 2024-01-05
fix: Move platform scripts

eriqua:
## Description

Move platform scripts out of the pipeline folder.
This is the easiest solution to avoid triggering all module pipelines in case changes are made to scripts only used by platform pipelines. 
The alternative would be to explicitly exclude platform folder from triggers in all module workflows

[avm.platform.toggle-avm-workflows](https://github.com/eriqua/bicep-registry-modules/actions/workflows/avm.platform.toggle-avm-workflows.yml)


.github/workflows/avm.platform.publish-tag.yml
```diff
@@ -48,7 +48,7 @@ jobs:
             Write-Output '::group::Publish tagged module to public bicep registry'
 
             # Load used functions
-            . (Join-Path $env:GITHUB_WORKSPACE 'avm' 'utilities' 'pipelines' 'platform' 'Publish-ModuleFromTagToPBR.ps1')
+            . (Join-Path $env:GITHUB_WORKSPACE 'avm' 'utilities' 'platform' 'Publish-ModuleFromTagToPBR.ps1')
```
eriqua, line 51: Solution under discussion

eriqua: Ping @ChrisSidebotham and @AlexanderSehr here for feedback. We have 2 platform PRs incoming

AlexanderSehr: ok - jap we still need to have that discussions / brainstorming ;)

ChrisSidebotham: I am for this change as discussed offline, happy to have a quick trio sync if we can squeeze it in

ChrisSidebotham: @eriqua - How about using an exclusion for the folder path within the original file structure?

e.g
` - !*/platform/*`

ChrisSidebotham:
@eriqua - can we add this to the loop for the next discussion?

ChrisSidebotham:
Coverting this to draft until a decision/conflict is made... @eriqua @AlexanderSehr

ChrisSidebotham:
@eriqua - Will we be picking this up again soon?

eriqua:
Closing in favor of #1294

PR #778, opened 2024-01-08
feat: Added IAM/RBAC to OperationalInsights Workspace (LAW) Table

AlexanderSehr:
## Description

- Added IAM/RBAC to OperationalInsights Workspace (LAW) Table

| Pipeline |
| - |
| [avm.res.operational-insights.workspace](https://github.com/AlexanderSehr/bicep-registry-modules/actions/workflows/avm.res.operational-insights.workspace.yml) |


krbar (commented):
Looks good to me, but should we also increase the module version?

AlexanderSehr:
> Looks good to me, but should we also increase the module version?

Good catch. Yes we can do that.

krbar (approved):

PR #782, opened 2024-01-10
feat: Region Rotator for Az Deployment Tests

ChrisSidebotham:
## Description

This PR is to support the capability of rotating regions at random during module deployment. due to availbility constraints we no longer wish to target a single region for all module testing. this PR brings the addition of a powershell script to filter regions based on the following:

- Recommended Regions from Microsoft 
- Paired Regions
- Exclusion of regions with availability constraints

## Testing

| Pipelines | 
| --------- | 
| [avm.res.automation.automation-account](https://github.com/ChrisSidebotham/bicep-registry-modules/actions/workflows/avm.res.automation.automation-account.yml) (Not Related)|
| [avm.res.key-vault.vault](https://github.com/ChrisSidebotham/bicep-registry-modules/actions/workflows/avm.res.key-vault.vault.yml)|
| [avm.res.network.dns-zone](https://github.com/ChrisSidebotham/bicep-registry-modules/actions/workflows/avm.res.network.dns-zone.yml) |
|[avm.res.compute.ssh-public-key](https://github.com/ChrisSidebotham/bicep-registry-modules/actions/workflows/avm.res.compute.ssh-public-key.yml)|
|[avm.res.web.serverfarm](https://github.com/ChrisSidebotham/bicep-registry-modules/actions/workflows/avm.res.web.serverfarm.yml)|
|[avm.res.logic.workflow](https://github.com/ChrisSidebotham/bicep-registry-modules/actions/workflows/avm.res.logic.workflow.yml)|
|[avm.res.data-factory.factory](https://github.com/ChrisSidebotham/bicep-registry-modules/actions/workflows/avm.res.data-factory.factory.yml)|

PR #803, opened 2024-01-12
fix: Revert Location Rotator and point to eastus until rotater is further refined

ChrisSidebotham:
## Description

As per current issues with region reliability. reverting #782 in favour of defaulting to `eastus`


ChrisSidebotham:
closing in favour of #804

PR #817, opened 2024-01-16
fix: Module `avm/res/web/site` array index is out of bounds issue

krbar:
## Description

Fixing the `The language expression property array index '2' is out of bounds..` issue in the module `avm/res/web/site`. The issue occurs in a specific scenario: when `appSettingsKeyValuePairs` parameter is not empty, but at the same time the `appInsightResourceId` or `storageAccountResourceId` is not provided.

## Pipeline references

| Pipeline |
| - |
| [avm.res.web.site](https://github.com/krbar/bicep-registry-modules/actions/workflows/avm.res.web.site.yml) |


ChrisSidebotham (approved):
🤌

PR #829, opened 2024-01-18
feat: Region Selector [WIP]

ChrisSidebotham:
## Description

This Pull Request is to introduce the region selector script used to rotate regional deployment as per the discussions within the core team. @AlexanderSehr FYI

### DIFF
--- a/avm/regions.json
+++ b/avm/regions.json
@@ -7,7 +7,7 @@
   "microsoft.cdn/profiles": [
     "eastus"
   ],
-  "microsoft.web/staticSite": [
+  "Microsoft.Web/staticSites": [
     "eastus2"
   ],
   "microsoft.compute/*": [
--- a/avm/regions.json
+++ b/avm/regions.json
@@ -11,7 +11,7 @@
     "eastus2"
   ],
   "microsoft.compute/*": [
-    "westus3"
+    "eastus"
   ]
 }
 
--- a/avm/regions.json
+++ b/avm/regions.json
@@ -1,0 +1,17 @@
+{
+  "default": [
+    "eastus",
+    "australiaeast",
+    "uksouth"
+  ],
+  "microsoft.cdn/profiles": [
+    "eastus"
+  ],
+  "microsoft.web/staticSite": [
+    "eastus2"
+  ],
+  "microsoft.compute/*": [
+    "westus3"
+  ]
+}
+
--- a/avm/utilities/pipelines/e2eValidation/regionSelector/Get-AzDeploymentRegion.ps1
+++ b/avm/utilities/pipelines/e2eValidation/regionSelector/Get-AzDeploymentRegion.ps1
@@ -50,7 +50,7 @@ function Get-AzDeploymentRegion {
   $controlledRegionList = Get-Content -Path $RepoRoot/avm/regions.json -Raw | ConvertFrom-Json
 
   # Validate if wildcard exists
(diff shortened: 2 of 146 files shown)


avm/utilities/pipelines/e2eValidation/regionSelector/Get-AzDeploymentRegion.ps1
```diff
@@ -5,96 +5,111 @@ Set the deployment region for the module deployment
...
 
-.PARAMETER excludedRegions
-Optional. The regions to exclude from the recommended regions list.
-
 .PARAMETER usePairedRegionsOnly
 Optional. If set, only paired regions will be returned.
 
-.PARAMETER useKnownRegionsOnly
+.PARAMETER moduleRoot
 Optional. If set, only known regions will be returned.
```
AlexanderSehr, line 12: The description seems off

AlexanderSehr (changes requested):

avm/regions.json
```diff
@@ -0,0 +1,17 @@
+{
```
AlexanderSehr, line 1: I wonder if this PR should also undo the tempLocations introduced for some module tests or if we should do that in a dedicated PR

AlexanderSehr (changes requested):

avm/regions.json
eriqua: I'd suggest to include the change in this PR

eriqua: Anyway, as per discussion, this json file should not be needed. We should retrieve the available locations for each resource via PS cmdlet. What we can keep is the logic to exclude specific locations known to have capacity issues. Please correct me if I'm missing something

avm/utilities/pipelines/e2eValidation/regionSelector/Get-AzDeploymentRegion.ps1
eriqua, line 1: The script name should probably be updated. It is not anymore the deployment location that we are going to retrieve, but the resource location. The deployment location should be deterministic for all deployments.

ChrisSidebotham: This PR is just in draft for me to align with Alex's function but good spot

avm/regions.json
ChrisSidebotham: I thought the agreed approach is still to use the json. especially for where we need to do SKU level exceptions for certain regions then we have a centrally controlled file for the module owner.

avm/utilities/pipelines/e2eValidation/regionSelector/Get-AzDeploymentRegion.ps1
ChrisSidebotham: WIP

avm/regions.json
eriqua: Not sure I got the same, let's sync offline on that

ChrisSidebotham:
Pending Merge #838

avm/regions.json
AlexanderSehr: I agree with @eriqua - BUT - think we agreed to keep the script to be able to still outright exclude locations. Together with the other location PR I think we should have a call sometime to finalize this.

ChrisSidebotham: Lets get this booked in

ChrisSidebotham:
closing in favour of #844

PR #840, opened 2024-01-21
fix: avm\res\databricks\workspace update AzureDatabricks objId

eriqua:
## Description

[avm.res.databricks.workspace](https://github.com/eriqua/bicep-registry-modules/actions/workflows/avm.res.databricks.workspace.yml)


ChrisSidebotham (approved):
Approving as module is orphaned

PR #862, opened 2024-01-26
fix: Global Resources in Region Selection

ChrisSidebotham:
## Description 

Global Resources was being filtered out due to a missing condition as 'Global' is not recognised by Get-AzLocation

## Tests

[avm.res.resource-graph.query](https://github.com/Azure/bicep-registry-modules/actions/workflows/avm.res.resource-graph.query.yml)


ChrisSidebotham:
CLosing in favour of #863

PR #908, opened 2024-02-01
feat: Add --with-source flag to bicep publish command

jtracey93:
This pull request primarily focuses on the enhancement of the module publishing process in the GitHub workflows and PowerShell scripts. The main change is the addition of the `--with-source` flag in the `bicep publish` command to include the source code while publishing the module.

Changes in GitHub workflows:

* [`.github/workflows/publish-module.yml`](diffhunk://#diff-dce5c6e5d79da2cb499477c887887e01776b06787df34431f7d1563f43a19efdL73-R73): Added `--with-source` flag in the `bicep publish` command under the `Publish module` job. This flag ensures that the source code is included when the module is published.

Changes in PowerShell scripts:

* [`avm/utilities/pipelines/platform/Publish-ModuleFromTagToPBR.ps1`](diffhunk://#diff-d2aff421712fb6d67688399143065aaa7ef6c3c613990095524c18a73df427d0R59): Included `--with-source` flag in the `bicep publish` command in the `Publish-ModuleFromTagToPBR` function. This change ensures that the source code is included when the module is published from a tag to PBR.
* [`avm/utilities/pipelines/publish/Publish-ModuleFromPathToPBR.ps1`](diffhunk://#diff-f6e98419158599eea6bbbeb80f2563633795674762eb88b4e2cc772a71532c7dR100): Added `--with-source` flag in the `bicep publish` command in the `Publish-ModuleFromPathToPBR` function. This ensures that the source code is included when the module is published from a path to PBR.

This PR cannot be merged until `v0.25` is released of bicep is released (hence draft initially)


ChrisSidebotham:
@jtracey93 - is the do not merge status still applicable?

.github/workflows/publish-module.yml
```diff
@@ -70,7 +70,7 @@ jobs:
           bicep --version
 
       - name: Publish module
-        run: bicep publish "modules/${{ steps.parse-tag.outputs.module_path }}/main.json" --target "br:${{ secrets.PUBLISH_REGISTRY_SERVER }}/public/bicep/${{ steps.parse-tag.outputs.module_path }}:${{ steps.parse-tag.outputs.version }}" --documentationUri "${{ steps.parse-tag.outputs.documentation_uri }}" --force
+        run: bicep publish "modules/${{ steps.parse-tag.outputs.module_path }}/main.json" --target "br:${{ secrets.PUBLISH_REGISTRY_SERVER }}/public/bicep/${{ steps.parse-tag.outputs.module_path }}:${{ steps.parse-tag.outputs.version }}" --documentationUri "${{ steps.parse-tag.outputs.documentation_uri }}" --with-source --force
```
eriqua, line 73: This pipeline is not used by AVM, but only by the `modules` folder. 
If we agreed to proceed separately we can revert this change

jtracey93: This will still work for BRM modules once they pass their tests to be published so can be left here.

StephenWeatherford (approved):
Looks good to me.

.github/workflows/publish-module.yml
AlexanderSehr: Please don't self-resolve comments. Makes re-reviews a lot harder 😉

To the topic at hand: Fair point, if they pass, they pass.

jtracey93: Fair, this can stay here though, there's no harm to add this the BRM modules publishing flow. They issues with those modules is not related to the publishing part, so this gets it in for when/if they are fixed for diagnostic settings etc.

eriqua: Sounds good. As per latest discussions, this workflow will likely never be run anyway.

eriqua (approved):
:ok:

PR #921, opened 2024-02-04
fix: remove ORPHANED.md from 2 recently parented modules

eriqua:
## Description

Remove ORPHANED.md from 2 recently parented modules


avm/res/document-db/database-account/README.md
```diff
@@ -1,10 +1,5 @@
 # DocumentDB Database Accounts `[Microsoft.DocumentDB/databaseAccounts]`
 
-> ⚠️THIS MODULE IS CURRENTLY ORPHANED.⚠️
```
eriqua, line 3: Blocked until owner not in right team

AlexanderSehr: Not sure I understand the comment :D

eriqua: Currently the corresponding module owners team is only having Kris as a member, who initially migrated the module.

AlexanderSehr: That's actuall sets an interesting precedence for the future. 'Should' we block this on the code level (here) or rather on the Process Level (i.e., on the AVM issue when somebody takes ownership?). I'd vote for the latter as this will become very tedious to review on our end (unless there is a good way doing so via Pester). Also, if we block it here, e.g. the plan to re-publish all modules will be blocked at least for these 2 modules now with the static tests failing in main.

eriqua: I've blocked it so that we can discuss those points and don't forget

eriqua: Or actually even better, I'm assigning @bryansan-msft for approval given he's the module owner 😎

ghost (approved):
Hello @eriqua, everything looks fine for me. Thanks for your help

PR #1041, opened 2024-02-19
feat: Module `avm/res/network/public-ip-prefix` API version update

krbar:
## Description

API version update of the module `avm/res/network/public-ip-prefix`

## Pipeline references

| Pipeline |
| - |
| [avm.res.network.public-ip-prefix](https://github.com/krbar/bicep-registry-modules/actions/workflows/avm.res.network.public-ip-prefix.yml) |


ChrisSidebotham (approved):
lgtm

PR #1052, opened 2024-02-21
fix: storage-account module - fixing Block Blob Storage creation

krbar:
## Description

- fixing Block Blob Storage creation
- added a test case testing Block Blob Storage creation

Resolves #1012 

| Pipeline |
| - |
| [avm.res.storage.storage-account](https://github.com/krbar/bicep-registry-modules/actions/workflows/avm.res.storage.storage-account.yml) |

## Updating an existing module

- [x] This is a bug fix:
  - [x] Someone has opened a bug report issue, and I have included "Closes #1012" in the PR description.
  - [ ] The bug was found by the module author, and no one has opened an issue to report it yet.
- [ ] I have run `brm validate` locally to verify the module files.
- [x] I have run deployment tests locally to ensure the module is deployable.
- [x] I have read the [Updating an existing module](https://github.com/Azure/bicep-registry-modules/blob/main/CONTRIBUTING.md#updating-an-existing-module) section in the contributing guide and updated the `version.json` file properly:
  - [x] The PR contains backwards compatible bug fixes, and I have NOT bumped the MAJOR or MINOR version in `version.json`.
  - [ ] The PR contains backwards compatible feature updates, and I have bumped the MINOR version in `version.json`.
  - [ ] The PR contains breaking changes, and I have bumped the MAJOR version in `version.json`.
- [ ] I have updated the examples in README with the latest module version number.

PR #1053, opened 2024-02-22
feat: New Module `avm/res/purview/account`

krbar:
## Description

New Module `avm/res/purview/account`, migrated from CARML.

## Adding a new module

## Pipeline references

| Pipeline |
| - |
| [avm.res.purview.account](https://github.com/krbar/bicep-registry-modules/actions/workflows/avm.res.purview.account.yml) |


avm/res/purview/account/tests/e2e/defaults/main.test.bicep
```diff
@@ -0,0 +1,46 @@
...
+
+// ========== //
+// Parameters //
+// ========== //
+@description('Optional. The name of the resource group to deploy for testing purposes.')
+@maxLength(90)
+param resourceGroupName string = 'dep-${namePrefix}-purview-${serviceShort}-rg'
+
+@description('Optional. The location to deploy resources to.')
+param resourceLocation string = 'eastus' // Only available in selected locations: eastus, eastus2, southcentralus, westcentralus, westus, westus2, westus3
```
AlexanderSehr, line 14: @krbar this value would be overwritten by the CI. If you' d want to hardcode it, it would need to be hardcoded in the deployment below.

In any case, like the storage PR #1052, this is another one where the general locations would run into trouble @eriqua & @ChrisSidebotham. I guess that means the PR would be blocked?

avm/res/purview/account/main.bicep
```diff
@@ -0,0 +1,416 @@
...
+    tags: privateEndpoint.?tags ?? tags
+    manualPrivateLinkServiceConnections: privateEndpoint.?manualPrivateLinkServiceConnections
+    customDnsConfigs: privateEndpoint.?customDnsConfigs
+    ipConfigurations: privateEndpoint.?ipConfigurations
+    applicationSecurityGroupResourceIds: privateEndpoint.?applicationSecurityGroupResourceIds
+    customNetworkInterfaceName: privateEndpoint.?customNetworkInterfaceName
+  }
+}]
+
+module portal_privateEndpoints 'br/public:avm/res/network/private-endpoint:0.4.0' = [for (privateEndpoint, index) in (portalPrivateEndpoints ?? []): {
```
AlexanderSehr, line 169: Does somebody know why we have like a dozen PE deployments in here? 😄

avm/res/purview/account/tests/e2e/defaults/main.test.bicep
AlexanderSehr: Same for the other test files of course.

avm/res/purview/account/tests/e2e/max/main.test.bicep
```diff
@@ -0,0 +1,187 @@
...
+
+@description('Optional. The location to deploy resources to.')
+param resourceLocation string = 'eastus' // Only available in selected locations: eastus, eastus2, southcentralus, westcentralus, westus, westus2, westus3
+
+@description('Optional. A short identifier for the kind of deployment. Should be kept short to not run into resource-name length-constraints.')
+param serviceShort string = 'pvamax'
+
+@description('Optional. A token to inject into the name of each resource.')
+param namePrefix string = '#_namePrefix_#'
+// =========== //
```
AlexanderSehr, line 21:
```suggestion

// =========== //
```

avm/res/purview/account/tests/e2e/waf-aligned/main.test.bicep
```diff
@@ -0,0 +1,135 @@
...
+
+@description('Optional. The location to deploy resources to.')
+param resourceLocation string = 'eastus' // Only available in selected locations: eastus, eastus2, southcentralus, westcentralus, westus, westus2, westus3
+
+@description('Optional. A short identifier for the kind of deployment. Should be kept short to not run into resource-name length-constraints.')
+param serviceShort string = 'pvawaf'
+
+@description('Optional. A token to inject into the name of each resource.')
+param namePrefix string = '#_namePrefix_#'
+// =========== //
```
AlexanderSehr, line 21:
```suggestion

// =========== //
```

avm/res/purview/account/tests/e2e/waf-aligned/main.test.bicep
```diff
@@ -0,0 +1,135 @@
...
+      'hidden-title': 'This is visible in the resource name'
+      Environment: 'Non-Prod'
+      Role: 'DeploymentValidation'
+    }
+    managedIdentities: {
+      userAssignedResourceIds: [
+        nestedDependencies.outputs.managedIdentityResourceId
+      ]
+    }
+    managedResourceGroupName: '${namePrefix}${serviceShort}001-managed-rg'
```
AlexanderSehr, line 79: Required for WAF alignment?

avm/res/purview/account/main.bicep
krbar: I was surprised too, it's a beast 😅

avm/res/purview/account/tests/e2e/defaults/main.test.bicep
krbar: Good catch, thanks. The hardcoded location comes from CARML. I don't insist on hardcoding it if we can ensure in other ways that only defined (supported) regions are used. We would probably need something like "allowedRegions" that can be defined per resource type, either as a parameter, or determined dynamically.

For now, I'll move the hardcoded location down to the deployment, but I'm open for better ideas.

eriqua: Also in this case I'd suggest to double check if the default location list covers this

AlexanderSehr: The default location list is: 'eastus', 'uksouth', 'northeurope', 'eastasia'

In other words, the deployment will fail for 
- `uksouth`
- `northeurope`
- `eastasia`

and work if `eastus` is selected
cc: @eriqa

avm/res/purview/account/tests/e2e/waf-aligned/main.test.bicep
krbar: Removed the managed identity, `waf-aligned` still looking good.
Commit: https://github.com/Azure/bicep-registry-modules/pull/1053/commits/ab0063d0b800f46b55c41d6255e5cc827a4f8fe4

avm/res/purview/account/tests/e2e/defaults/main.test.bicep
krbar: In the mean time much more regions are supported:
`eastus`, `westeurope`, `southeastasia`, `canadacentral`, `canadaeast`, `swedencentral`, `southcentralus`, `brazilsouth`, `centralindia`, `uksouth`, `francecentral`, `koreacentral`, `uaenorth`, `japaneast`, `switzerlandnorth`, `westus`, `southafricanorth`, `germanywestcentral`, `qatarcentral`, `westus3`, `australiaeast`, `northeurope`, `westcentralus`, `westus2`, `eastus2`

However `eastasia` is not on the list.

krbar: > @krbar this value would be overwritten by the CI. If you' d want to hardcode it, it would need to be hardcoded in the deployment below.

Placing a hardcoded location directly in the deployment creates a Bicep warning and an issue while generating the README. Therefore I've created a new parameter `fixedResourceLocation` to overcome these issues. I still hope we can remove this hardcoded location completely.

avm/res/purview/account/tests/e2e/defaults/main.test.bicep
```diff
@@ -0,0 +1,48 @@
...
+// ========== //
+@description('Optional. The name of the resource group to deploy for testing purposes.')
+@maxLength(90)
+param resourceGroupName string = 'dep-${namePrefix}-purview-${serviceShort}-rg'
+
+@description('Optional. The location to deploy resources to.')
+param resourceLocation string = deployment().location
+
+@description('Optional. The location to deploy resources to.')
+param fixedResourceLocation string = 'eastus' // Not available all locations used for testing, so we need a fixed location
```
AlexanderSehr, line 17: I suggest using the approach we temporarily applied in the past using a variable like this

```bicep
#disable-next-line no-hardcoded-location // services used for this module are only available in major regions
var tempLocation = 'eastus'
```

instead of introducing another parameter that won't be used as a parameter. We undid the change back then in this [PR](https://github.com/Azure/bicep-registry-modules/pull/880/files).

AlexanderSehr: Same for the other tests

avm/res/purview/account/tests/e2e/defaults/main.test.bicep
AlexanderSehr: See comment I added

avm/res/purview/account/main.bicep
```diff
@@ -0,0 +1,416 @@
...
+  name: '${uniqueString(deployment().name, location)}-Portal-PrivateEndpoint-${index}'
+  params: {
+    privateLinkServiceConnections: [
```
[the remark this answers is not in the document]
AlexanderSehr: The [schema update](https://azure.github.io/Azure-Verified-Modules/specs/shared/interfaces/#private-endpoints) was merged. Would be great if we could apply it here.

avm/res/purview/account/tests/e2e/defaults/main.test.bicep
eriqua: The region rotation is anyway selecting the location among the default locations which are also available for the service. If eastasia is not in the list of available locations for purview, it won't be picked. Did we test the module with the default behavior?

avm/res/purview/account/main.bicep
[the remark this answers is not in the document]
krbar: Updated: https://github.com/Azure/bicep-registry-modules/pull/1053/commits/2c854d21e6f56004eb037dfce841f39d2db3a2d2

avm/res/purview/account/tests/e2e/defaults/main.test.bicep
krbar: No, this is new to me. I can give it a try.

krbar: It looks like the region selection logic works fine for the Purview. I ran the pipeline several times in a row and there were no attempts to deploy in `eastasia`. I removed the workaround.

avm/res/purview/account/tests/e2e/defaults/main.test.bicep
krbar: I removed the location workaround completely, see the other comment.

avm/res/purview/account/main.bicep
```diff
@@ -189,29 +212,40 @@ module portal_privateEndpoints 'br/public:avm/res/network/private-endpoint:0.4.0
...
-    privateLinkServiceConnections: [
+    name: privateEndpoint.?name ?? 'pep-${last(split(account.id, '/'))}-${privateEndpoint.service}-${index}'
+    privateLinkServiceConnections: privateEndpoint.?manualPrivateLinkServiceConnections != true ? [
       {
-        name: name
+        name: privateEndpoint.?privateLinkServiceConnectionName ?? '${last(split(account.id, '/'))}-${privateEndpoint.service}-${index}'
         properties: {
           privateLinkServiceId: account.properties.managedResources.storageAccount
           groupIds: [
-            privateEndpoint.?service ?? 'blob'
```
AlexanderSehr: Hey @krbar was there a reason you removed the default's all over? Can this maybe only be `blob` for purview?

krbar: @AlexanderSehr Updated to Private Endpoints variant 1, with the defaults

AlexanderSehr: I see where you're comming from. Just leaning towards variant one because the way the resources are named heavily implies they're created for a very specific service (e.g. `blob`)

AlexanderSehr: Thanks for recovering :)

AlexanderSehr (approved):

PR #1080, opened 2024-02-26
fix: Used Network-Manager token in test file

AlexanderSehr:
## Description

- Fixed used Network-Manager token in test file

## Pipeline Reference

| Pipeline |
| - |
| [avm.res.network.network-manager](https://github.com/AlexanderSehr/bicep-registry-modules/actions/workflows/avm.res.network.network-manager.yml) |


eriqua (approved):
:ok:

PR #1140, opened 2024-03-05
ci: Exclude archived BRM modules from module index data

shenglol:
Updated the script for generating module index data to exclude archived BRM modules.

### DIFF
--- a/scripts/github-actions/generate-module-index-data.js
+++ b/scripts/github-actions/generate-module-index-data.js
@@ -79,6 +79,7 @@ async function getModuleDescription(
  */
 async function generateModuleIndexData({ require, github, context, core }) {
   const fs = require("fs").promises;
+  const { existsSync } = require("fs");
   const axios = require("axios").default;
   const moduleIndexData = [];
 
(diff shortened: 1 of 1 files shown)


scripts/github-actions/generate-module-index-data.js
```diff
@@ -91,6 +92,12 @@ async function generateModuleIndexData({ require, github, context, core }) {
 
     for (const moduleName of moduleNames) {
       const modulePath = `${moduleGroupPath}/${moduleName}`;
+      const archivedFilePath = `${modulePath}/ARCHIVED.md`;
+
+      if (existsSync(archivedFilePath)) {
+        continue;
+      }
+
```
shenglol, line 100: Tested the logic locally.

matebarabas: Ok, so as discussed, this jumps out from the for cycle and continues with the next one, in case the ARCHIVED.md file exists - which is exactly what we need to not include those modules in the index and IntelliSense JSON that were already archived. Thanks a lot!

matebarabas (approved):

PR #1224, opened 2024-03-09
feat: Added UDT for Managed identity and Deployment Scripts

johnlokerse:
## Description

This pull request introduces two user-defined types for the following modules:
* Deployment scripts -> (secure) environment variables (tagging module owner: @sebassem)
* User Assigned Managed Identity -> federated identity credentials (tagging module owner: @elanzel)
* Removed an unused line of code in `.github/actions/templates/avm-validateModuleDeployment/action.yml`

Extra context regarding `@secure()` decorator on user-defined type. When the Bicep template is built the type is referred as a secureString so the parameter inherits the secureString type and the parameter is secure during deployment:

Untitled

Also double checked if the outputs were working and tested it via this Bicep template:

```bicep
/* Bicepparam
using './testsecureoutput.bicep'

param parTestingParameter = {
  secureList: [
    {
      name: 'test'
      secureValue: 'test123'
    }
  ]
}
*/

param parTestingParameter environmentVariableType

output outTestBool bool = parTestingParameter != null
output outTestArray array = parTestingParameter != null ? parTestingParameter!.secureList : []
output outTestArrayV2 array = parTestingParameter!.secureList ?? []
output outTestString string = first(parTestingParameter.secureList)!.secureValue!

@secure()
type environmentVariableType = {
  secureList: {
    name: string
    secureValue: string?
    value: string?
  }[]
}?
```

## Pipeline Reference

[avm.res.managed-identity.user-assigned-identity](https://github.com/johnlokerse/bicep-registry-modules/actions/workflows/avm.res.managed-identity.user-assigned-identity.yml)

[avm.res.resources.deployment-script](https://github.com/johnlokerse/bicep-registry-modules/actions/workflows/avm.res.resources.deployment-script.yml)

## Type of Change

  - [x] Update to documentation

## Checklist


johnlokerse:
@AlexanderSehr FYI - had some issues with the randomised resource location selector on the deployment of the user-assigned managed identity. It deployed in `eastasia`, but that region does not support federated credentials. Maybe being able to override the location would be handy.

AlexanderSehr:
> @AlexanderSehr FYI - had some issues with the randomised resource location selector on the deployment of the user-assigned managed identity. It deployed in `eastasia`, but that region does not support federated credentials. Maybe being able to override the location would be handy

Hey @johnlokerse,
that's indeed unfortunate as the location selector tries to only select regions where the resource exists, yet fails if certain features don't exist. 
I worked around it in some other module recently by introducing an additional parameter to enforce a certain location 

https://github.com/Azure/bicep-registry-modules/blob/a964d0c128afd8ccfe1d2ecbb3428868214cdf17/avm/res/purview/account/tests/e2e/max/main.test.bicep#L24-26

cc: @eriqua, @ChrisSidebotham fyi

avm/res/resources/deployment-script/main.bicep
```diff
@@ -195,7 +183,7 @@ resource deploymentScript 'Microsoft.Resources/deploymentScripts@2023-08-01' = {
     containerSettings: !empty(containerSettings) ? containerSettings : null
     storageAccountSettings: !empty(storageAccountResourceId) ? storageAccountSettings : null
     arguments: arguments
-    environmentVariables: !empty(environmentVariables) ? environmentVariables.secureList : []
+    environmentVariables: environmentVariables != null ? environmentVariables!.secureList : []
```
AlexanderSehr, line 186: I think this could be simplified to

```suggestion
    environmentVariables: environmentVariables.?secureList ?? []
```
But no guarantees without testing 😄

AlexanderSehr (dismissed):

avm/res/resources/deployment-script/main.bicep
johnlokerse: I tried the coalesce way before, but didn't work for some reason.. Just tried it again and for some reason I got it working 😄. Maybe we should make it even simpler by removing `secureList`? There's no for it because of the secure decoration on the UDT.

AlexanderSehr: Hey @johnlokerse,
if you can make it so that the `@secure` works with arrays, then I'd be all for it. It's non-support for array was the reason we used an object in the first place (in this and other modules)

johnlokerse: Yes, that's true... forget about it. Will leave it as-is for now 😄 Auto-completion with the UDT makes it a lot better already.

sebassem (approved):
LGTM :)

AlexanderSehr (approved):

PR #1256, opened 2024-03-12
feat: Add more VM extensions and build a dependency graph - `avm/res/compute/virtual-machine`

### LINKED ISSUE #1067: [AVM Question/Feedback]: VM extension

### Description

Dear AVM maintainers, 

While deploying the Azure Virtual Machine, it is required to perform a domain join (AD). In the extension, you can of course set the restarted setting to false, as such:

```
extensionDomainJoinConfig: {
      enabled: true
      settings: {
        name: '' 
        user: '
        restart: 'false' // setting
        options: 3
        ouPath:
      }
    }
```

But the virtual machine needs to be restarted anyway and this is particularly helpful when doing post deployment task. The only thing that I'm looking for, is that after the domain join has happened, the custom script extension sometime fails if you combine them (because of the restart). Is it possible to do a "dependsOn"?

Another extension that I was missing, was the GuestConfiguration extension. Are there any plans on adding it to the VM?

Thanks in advance.

rahalan:
Successful run (except nVidia due to capacity restrains - it still works, I tested it in AVM subcription): [avm.res.compute.virtual-machine](https://github.com/rahalan/bicep-registry-modules/actions/workflows/avm.res.compute.virtual-machine.yml)

New extensions:
- nVidia graphics driver
- add VM to hostpool
- Guest Configuration

Because some extensions cannot be installed at the same time, they are all now serialized (one extension at a time)


avm/res/compute/virtual-machine/tests/e2e/linux.max/dependencies.bicep
```diff
@@ -214,12 +217,28 @@ resource keyVault 'Microsoft.KeyVault/vaults@2022-07-01' = {
   }
 }
 
+resource backupServiceKeyVaultPermissions 'Microsoft.Authorization/roleAssignments@2022-04-01' = {
+  name: guid('${keyVault.name}-${location}-268f6a53-9f68-4a38-ae47-166f730d86af-KeyVault-KeyVaultAdministrator-RoleAssignment')
+  scope: keyVault
+  properties: {
+    principalId: '268f6a53-9f68-4a38-ae47-166f730d86af' // Backup Management Service Enterprise Application Object Id (Note: this is tenant specific)
```
AlexanderSehr, line 224: Is that the AVM-specific one? Same question for the other tests.

AlexanderSehr (changes requested):

avm/res/compute/virtual-machine/main.bicep
```diff
@@ -174,6 +174,31 @@ param extensionCustomScriptConfig object = {
...
+  enabled: false
+}
+
+@description('Optional. The configuration for the [Host Pool Registration] extension. Must at least contain the ["enabled": true] property to be executed. Needs a managed identy.')
+param extensionHostPoolRegistration object = {
+  enabled: false
+}
+
+@description('Optional. Required if host pool registiration extension is used. Name of the host pool.')
+param hostPoolName string = ''
```
AlexanderSehr, line 188: Why not make this part of the host pool registration object?  Name question for the other property

AlexanderSehr (changes requested):

avm/res/compute/virtual-machine/main.bicep
```diff
@@ -98,14 +98,14 @@ param proximityPlacementGroupResourceId string = ''
...
 param availabilitySetResourceId string = ''
 
-@description('Optional. If set to 1, 2 or 3, the availability zone for all VMs is hardcoded to that value. If zero, then availability zones is not used. Cannot be used in combination with availability set nor scale set.')
+@description('Required. If set to 1, 2 or 3, the availability zone for all VMs is hardcoded to that value. If zero, then availability zones is not used. Cannot be used in combination with availability set nor scale set.')
 @allowed([
-  0
-  1
-  2
-  3
+  '0'
```
AlexanderSehr, line 103: I take it numbers are not supported? I wonder if we should keep the input across all modules as an integer and just convert the value to a string. Something to discuss in the team, I guess

avm/res/compute/virtual-machine/main.bicep
```diff
@@ -618,19 +820,125 @@ module vm_azureDiskEncryptionExtension 'extension/main.bicep' = if (extensionAzu
...
+     ? extensionHostPoolRegistration.typeHandlerVersion 
+     : '2.77'
+    autoUpgradeMinorVersion: contains(extensionHostPoolRegistration, 'autoUpgradeMinorVersion') 
+     ? extensionHostPoolRegistration.autoUpgradeMinorVersion 
+     : true
+    enableAutomaticUpgrade: contains(extensionHostPoolRegistration, 'enableAutomaticUpgrade') 
+     ? extensionHostPoolRegistration.enableAutomaticUpgrade 
+     : false
+    settings: {
+      modulesUrl: 'https://wvdportalstorageblob.blob.${environment().suffixes.storage}/galleryartifacts/Configuration_09-08-2022.zip'
```
AlexanderSehr, line 885: I wonder if that value should be configurable. It would appear the file name could change over time? Even the storage account has technically an outdated name (wvd vs avd).
Same question for the configuration function

AlexanderSehr (changes requested):

avm/res/compute/virtual-machine/tests/e2e/windows.hostpool/dependencies.bicep
```diff
@@ -0,0 +1,113 @@
...
+      '${managedIdentity.id}': {}
+    }
+  }
+  dependsOn: [
+    msiHPDesktopVirtualizationVirtualMachineContributorRoleAssignment
+  ]
+  properties: {
+    azPowerShellVersion: '10.0'
+    arguments: '-HostPoolName "${hostPool.name}" -HostPoolResourceGroupName "${resourceGroup().name}" -SubscriptionId "${subscription().subscriptionId}"'
+    scriptContent: '''
```
AlexanderSehr, line 82: I'd recommend to move this into a script in the same way we did for the virtual machine images tests. Take a look, maybe it speaks to you

AlexanderSehr (changes requested):

avm/res/compute/virtual-machine/main.bicep
rahalan: Chris mentioned, that the array in the Bicep reference for availability zones, is a string array, so I changed it to that

avm/res/compute/virtual-machine/main.bicep
rahalan: done

avm/res/compute/virtual-machine/main.bicep
rahalan: Moved it also into the host pool registration object

avm/res/compute/virtual-machine/tests/e2e/windows.hostpool/dependencies.bicep
rahalan: done

avm/res/compute/virtual-machine/main.bicep
AlexanderSehr: I'll bring it up in the team :) to test the waters

avm/res/compute/virtual-machine/main.bicep
```diff
@@ -133,7 +133,7 @@ param extensionDomainJoinConfig object = {
   enabled: false
 }
 
-@description('Optional. The configuration for the [AAD Join] extension. Must at least contain the ["enabled": true] property to be executed.')
+@description('Optional. The configuration for the [AAD Join] extension. Must at least contain the ["enabled": true] property to be executed. To enroll in Intune, add the setting mdmId: "0000000a-0000-0000-c000-000000000000".')
```
AlexanderSehr, line 136:
```suggestion
@description('Optional. The configuration for the [AAD Join] extension. Must at least contain the ["enabled": true] property to be executed. To enroll in Intune, add the setting mdmId: "<GUID>".')
```

Does it have to be that explicit ID? If not we could keep it generic. OR - if it must be the `00000` we should, I guess remove the letters from the ID, right?
```suggestion
@description('Optional. The configuration for the [AAD Join] extension. Must at least contain the ["enabled": true] property to be executed. To enroll in Intune, add the setting mdmId: "00000000-0000-0000-0000-000000000000".')
```

avm/res/compute/virtual-machine/main.bicep
AlexanderSehr: What about the aforementioned configuration function?

avm/utilities/e2e-template-assets/scripts/Get-HostPoolRegistrationKey.ps1
```diff
@@ -0,0 +1,50 @@
...
+  [string] $HostPoolName,
+
+  [Parameter(Mandatory = $true)]
+  [string] $HostPoolResourceGroupName,
+
+  [Parameter(Mandatory = $true)]
+  [string] $SubscriptionId
+)
+
+if ($installedModules.Name -notcontains "Az.DesktopVirtualization") {
```
AlexanderSehr, line 33: Where is `$installedModules` coming from? It appears the variable has not been initialized

avm/res/compute/virtual-machine/main.bicep
```diff
@@ -618,19 +813,125 @@ module vm_azureDiskEncryptionExtension 'extension/main.bicep' = if (extensionAzu
...
+     ? extensionGuestConfigurationExtension.protectedSettings 
+     : {}
+    tags: extensionGuestConfigurationExtension.?tags ?? tags
+  }
+  dependsOn: [
+    vm_hostPoolRegistrationExtension
+  ]
+}
+
+resource AzureWindowsBaseline 'Microsoft.GuestConfiguration/guestConfigurationAssignments@2020-06-25' = if (!empty(guestConfiguration)) {
```
AlexanderSehr, line 926: Question: Does that need a dependency on `vm_azureGuestConfigurationExtension` as per your description of the parameter `guestConfiguration`.

avm/res/compute/virtual-machine/tests/e2e/windows.hostpool/main.test.bicep
```diff
@@ -0,0 +1,113 @@
+targetScope = 'subscription'
+
+metadata name = 'Using only defaults for Windows'
+metadata description = 'This instance deploys the module with the minimum set of required parameters.'
```
AlexanderSehr, line 4: These seem incorrect for the test

avm/res/compute/virtual-machine/main.bicep
rahalan: It has to be the exact sated string

avm/res/compute/virtual-machine/main.bicep
rahalan: Now it needs to be defined by the consumer of the module. In the test they can see how it works and maybe change the URL, if needed

avm/res/compute/virtual-machine/tests/e2e/windows.hostpool/main.test.bicep
rahalan: Fixed

avm/res/compute/virtual-machine/main.bicep
rahalan: correct. fixed

avm/utilities/e2e-template-assets/scripts/Get-HostPoolRegistrationKey.ps1
rahalan: fixed

avm/res/compute/virtual-machine/main.bicep
```diff
@@ -174,6 +174,24 @@ param extensionCustomScriptConfig object = {
...
+  enabled: false
+}
+
+@description('Optional. The configuration for the [Host Pool Registration] extension. Must at least contain the ["enabled": true] property to be executed. Needs a managed identy.')
+param extensionHostPoolRegistration object = {
+  enabled: false
+}
+
+@description('Optional. The configuration for the [Guest Configuration] extension. Must at least contain the ["enabled": true] property to be executed. Needs a managed identy.')
+param extensionGuestConfigurationExtension object = {
```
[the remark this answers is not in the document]
rahalan: Good question. I have no idea. Do we have a spec that could give us guidance here?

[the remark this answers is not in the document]
AlexanderSehr: Not as far as I know, but there are a few options here:
1. Make the whole object `secure()`
2. Move the protection settings out and only make them `secure()`
3. Create a UDT for the extension resource in which you specifiy the protected settings as `secure()`. 
The last would likely be the 'best' solution, yet the one that requires the most work. All 3 would work though. @eriqua, @ChrisSidebotham any thoughts/suggestions from your side?

avm/res/compute/virtual-machine/tests/e2e/windows.hostpool/main.test.bicep
```diff
@@ -0,0 +1,113 @@
+targetScope = 'subscription'
+
+metadata name = 'Using a host pool to register the VM'
+metadata description = 'This instance deploys the module and register it in a host pool.'
```
AlexanderSehr, line 4:
```suggestion
metadata description = 'This instance deploys the module and registers it in a host pool.'
```

AlexanderSehr: Note, requires refreshing the readme too

avm/res/compute/virtual-machine/main.bicep
AlexanderSehr: But isn't it still hardcoded? I can see the following in the template:
```bicep
configurationFunction: 'Configuration.ps1\\AddSessionHost'
```

I'd suggest to update to something like 
```bicep
configurationFunction: extensionHostPoolRegistration.?configurationFunction ?? 'Configuration.ps1\\AddSessionHost'
```

rahalan: done

avm/res/compute/virtual-machine/main.bicep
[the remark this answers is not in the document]
rahalan: I went with option 2 for now, to be consistent with the other extension

avm/res/compute/virtual-machine/main.bicep
AlexanderSehr: Alright, as discussed please change this back to the original int.

rahalan: Done

AlexanderSehr (approved):

PR #1276, opened 2024-03-13
fix: Corrected incorrect token regex 

AlexanderSehr:
## Description

Fixed fail-safe regex (added missing escapes). Until now, `--------` would cause an issue, even though we're only interested in `-..--..-`.

Output for Azure-Firewall with default behavior:
image

Output for Azure Firewall with disabled token replacement (to validate the fail-safe works):
image

## Pipeline Reference

| Pipeline |
| -------- |
[avm.res.network.azure-firewall](https://github.com/AlexanderSehr/bicep-registry-modules/actions/workflows/avm.res.network.azure-firewall.yml)        
[avm.res.network.bastion-host](https://github.com/AlexanderSehr/bicep-registry-modules/actions/workflows/avm.res.network.bastion-host.yml)
[avm.res.network.virtual-network](https://github.com/AlexanderSehr/bicep-registry-modules/actions/workflows/avm.res.network.virtual-network.yml)

## Type of Change

  - [ ] Update to documentation


avm/utilities/pipelines/publish/Publish-ModuleFromPathToPBR.ps1
```diff
@@ -77,11 +77,11 @@ function Publish-ModuleFromPathToPBR {
   $null = Convert-TokensInFileList @tokenConfiguration
 
   # Double-check that tokens are correctly replaced
-  $templateContent = Get-Content -Path $moduleBicepFilePath
+  $templateContent = bicep build $moduleBicepFilePath --stdout
```
eriqua, line 80: Do we want to include minor changes to modules missing publishing in this PR? Or should that happen separately?

AlexanderSehr: Hey @eriqua, it's a good question. We can - or - we re-try to use the publish from tag workflow first. Thoughts? Either is fine with me.

eriqua: Unfortunately that wouldn't work. The code to publish the tag is checked out from the tag itself, so this fix wouldn't be included.

AlexanderSehr: on it

AlexanderSehr: done

avm/res/network/azure-firewall/main.bicep
```diff
@@ -14,7 +14,7 @@ param name string
 param azureSkuTier string = 'Standard'
 
 @description('Conditional. Shared services Virtual Network resource ID. The virtual network ID containing AzureFirewallSubnet. If a Public IP is not provided, then the Public IP that is created as part of this module will be applied with the subnet provided in this variable. Required if `virtualHubId` is empty.')
-param virtualNetworkId string = ''
+param virtualNetworkResourceId string = ''
```
eriqua, line 17: Although I agree with the change, updating an input parameter is a breaking change. For azure fw, since the module was not published, I guess we can avoid to care. But for bastion and vnet this would mean a minor version update.

AlexanderSehr: Hence I updated the version number...

avm/res/network/bastion-host/main.bicep
```diff
@@ -8,8 +8,8 @@ param name string
 @description('Optional. Location for all resources.')
 param location string = resourceGroup().location
 
-@description('Required. Shared services Virtual Network resource identifier.')
-param vNetId string
+@description('Required. Shared services Virtual Network resource Id.')
+param vNetResourceId string
```
eriqua, line 12: At this point what about aligning with what in az fw 
```suggestion
param virtualNetworkResourceId string
```

Will not block the PR on this, just seemed strange not to mention, since we're updating the same parameter used by the same setting implementation

AlexanderSehr: done

eriqua (approved):
🆗

PR #1293, opened 2024-03-15
feat: Update the issue workflow to @ mentioning the module team on a failed workflow run

rahalan:
If the workflow is not a module one, than the AVM core team gets mentioned


avm/utilities/pipelines/platform/Set-AvmGithubIssueForWorkflow.ps1
```diff
@@ -82,7 +82,15 @@ function Set-AvmGithubIssueForWorkflow {
 
       if ($issues.title -notcontains $issueName) {
         if ($PSCmdlet.ShouldProcess("Issue [$issueName]", 'Create')) {
-          gh issue create --title "$issueName" --body "$failedrun" --label 'Type: AVM :a: :v: :m:,Type: Bug :bug:' --repo $Repo
+          $issueUrl = gh issue create --title "$issueName" --body "$failedrun" --label 'Type: AVM :a: :v: :m:,Type: Bug :bug:' --repo $Repo
+          $notifyTeam = "Azure/avm-core-team-technical-bicep"
+
+          if ($workflowRun.workflowName -match "avm.(?:res|ptn)") {
+            $notifyTeam = "Azure/$($workflowRun.workflowName.Replace(".","-"))-module-owners-bicep"
```
eriqua, line 89: What if the module is orphaned?

rahalan: @eriqua I added a check, if the module has an owner. In that case, the module owner team is mentioned, otherwise the avm technical team

eriqua: Sounds good, unless we want to add a specific message if the module is orphaned, just to set expectations. That would be in general for module feature requests and bugs in addition to failed pipelines. In any case, I wouldn't block the PR on that. This is already an improvement. 
I'd just ask if possible to wait for the triggers PR to be merged first, so that this PR will not trigger all module workflows again.

eriqua: Since we discussed about this offline today, tagging @matebarabas in case we want to add a specific notice/tag in case the module is orphaned

rahalan:
Will be replaced with a new PR

PR #1311, opened 2024-03-17
fix: Skip deployment if PSRule WAF reliability validation fails

eriqua:
## Description

Update reusable workflow validation template to correctly skip deployment validation if the newly introduced PSRule WAF reliability check fails

## Pipeline Reference

| Pipeline |
| -------- |
| [avm.res.network.bastion-host](https://github.com/Azure/bicep-registry-modules/actions/workflows/avm.res.network.bastion-host.yml) (As intended, skips deployment after PSRule reliability check failure) |

## Type of Change

  - [ ] Update to documentation

## Checklist

### DIFF
--- a/.github/workflows/avm.template.module.yml
+++ b/.github/workflows/avm.template.module.yml
@@ -101,7 +101,7 @@ jobs:
     runs-on: ubuntu-latest
     if: |
       !cancelled() &&
-      (fromJson(inputs.workflowInput)).deploymentValidation == 'true' && needs.job_module_static_validation.result != 'failure' && needs.job_psrule_test_waf_reliability != 'failure'
+      (fromJson(inputs.workflowInput)).deploymentValidation == 'true' && needs.job_module_static_validation.result != 'failure' && needs.job_psrule_test_waf_reliability.result != 'failure'
     needs:
       - job_module_static_validation
       - job_psrule_test_waf_reliability


.github/workflows/avm.template.module.yml
```diff
@@ -101,7 +101,7 @@ jobs:
     runs-on: ubuntu-latest
     if: |
       !cancelled() &&
-      (fromJson(inputs.workflowInput)).deploymentValidation == 'true' && needs.job_module_static_validation.result != 'failure' && needs.job_psrule_test_waf_reliability != 'failure'
+      (fromJson(inputs.workflowInput)).deploymentValidation == 'true' && needs.job_module_static_validation.result != 'failure' && needs.job_psrule_test_waf_reliability.result != 'failure'
```
AlexanderSehr, line 104: I must have missed the major change to introduce this job. Let's please have a chat why we did this.
For this PR now (though it seems already too late) - should we not maybe fix the modules first before we suddenly block any changes? Mind you we (including me) have module update PRs open and I'd be forced to fix any issues for the owners and unless I don't want to wait for an unspecified amount of time 😬

eriqua: We sure can. Meanwhile let me try to provide some context briefly:

- This PR is only fixing a bug for a change already merged. With the current approach the deployment validation would run, but publishing would not anymore anyway, as it is skipped on the `succeed()` condition. See e.g. [this run](https://github.com/Azure/bicep-registry-modules/actions/runs/8299629678). So this fix is not actually blocking the contribution progress more than what we already have in main.
- The blocker is on WAF reliability rules only and not on the entire PSRule validation, hence the duplicated job. This is temporary meanwhile we get to a stage where all PSRule checks become mandatory. At that point we could easily remove the duplicated reliability job and make the deployment dependent on the entire PSRule validation.
- On the decision to block itself, that's the intention going forward. I'm with you that this will impact the contribution effort and slow down the CARML migration completion. The current standpoint is that enforcing WAF reliability must take precedence over migration.
- The amount of time within which owners are asked to fix reliability checks will not be 'undefined' ideally. We'll open a call for action with a specific deadline.
- Looking at the bright side, the number of modules not WAF reliability compliant are very limited.

Cc @jtracey93 @matebarabas

AlexanderSehr (approved):

.github/workflows/avm.template.module.yml
AlexanderSehr: Thanks for elaborating @eriqua.
I know it's too late already, hence the call out. Approved

PR #1333, opened 2024-03-19
feat: Remove PSRule Suppressions for WAF Pillar Testing

jtracey93:
## Description

This pull request primarily focuses on the modification of rule specifications in the `avm/utilities/pipelines/staticValidation/psrule/.ps-rule/min-suppress.Rule.yaml` and `avm/utilities/pipelines/staticValidation/psrule/ps-rule.yaml` files. The changes involve the removal of specific rules from the `spec:` and `rule:` sections.

Here are the key changes:

* [`avm/utilities/pipelines/staticValidation/psrule/.ps-rule/min-suppress.Rule.yaml`](diffhunk://#diff-1103fa1c7d164036fd3b0f245e6ee0e585597c2eb7f5524e7b35d37b314824d8L20): Removed `Azure.KeyVault.Firewall` from the Key Vault specific rules and `Azure.VM.Standalone` from the Azure Virtual Machine rules in the `spec:` section. [[1]](diffhunk://#diff-1103fa1c7d164036fd3b0f245e6ee0e585597c2eb7f5524e7b35d37b314824d8L20) [[2]](diffhunk://#diff-1103fa1c7d164036fd3b0f245e6ee0e585597c2eb7f5524e7b35d37b314824d8L30)
* [`avm/utilities/pipelines/staticValidation/psrule/ps-rule.yaml`](diffhunk://#diff-070dba5b56ce2fbeee30f96558178ae8bdfee4a84270696216af3d1ed9902131L79): Removed `Azure.KeyVault.PurgeProtect` from the list of ignored rules in the `rule:` section.

These changes seem to indicate a shift in the rules that are being applied or ignored in the static validation process.

## Pipeline Reference

| Pipeline |
| -------- |
|     [avm.res.compute.virtual-machine](https://github.com/jtracey93/bicep-registry-modules/actions/workflows/avm.res.compute.virtual-machine.yml)     |
| [avm.res.batch.batch-account](https://github.com/jtracey93/bicep-registry-modules/actions/workflows/avm.res.batch.batch-account.yml) |

> Failures are expected as the modules are not meeting the test requirements

## Type of Change

  - [ ] Update to documentation

## Checklist

- [] I have run `Set-AVMModule` locally to generate the supporting module files.
- [] My corresponding pipelines / checks run clean and green without any errors or warnings


avm/utilities/pipelines/staticValidation/psrule/.ps-rule/min-suppress.Rule.yaml
```diff
@@ -27,7 +26,6 @@ spec:
     - Azure.ContainerApp.PublicAccess
     # Azure Virtual Machine
     - Azure.VM.AMA
-    - Azure.VM.Standalone
```
eriqua, line 30: On this I'd like @rahalan to comment as I remember he discussed a reason for this to be in place. If so let's maybe add a comment to detail that reason

avm/utilities/pipelines/staticValidation/psrule/ps-rule.yaml
```diff
@@ -76,5 +76,4 @@ rule:
   includeLocal: false
   exclude:
     # Ignore the following rules for all resources
-    - Azure.KeyVault.PurgeProtect
```
eriqua, line 79: We cannot remove this as we need purge protection to be disabled on keyvault in order to be able to continuously deploy and delete the resource but also be able to run idempotency test on the same resource

jtracey93: okay understood. But we are saying that the module defaults should be WAF aligned, so how do we achieve that?

Our tests shouldnt block us from achieving this.

avm/utilities/pipelines/staticValidation/psrule/.ps-rule/min-suppress.Rule.yaml
jtracey93: This would just require in the default test to have an AZ specified.

I will be submitting another test for the VM module to check that the AZ parameter does not have a default value very soon, as this is what we agreed recently and a consumer must provide inputs for the AZ parameter

eriqua: Agree, just wanted to double check with the module owner if the exception is there for any other reason or put just temporarily to speed up the migration (I guess the latter)

avm/utilities/pipelines/staticValidation/psrule/ps-rule.yaml
eriqua: I see the angle and I can assure this has been discussed many times in the past as well.
I'm open to further discuss that, so far we did not find a valid alternative to mitigate this issue.

Purge protection for our CI cannot be enforced until we are able to automatically recover soft-deleted resources. That will not only solve this blocker, but also simplify a lot our CI removal process, currently taking into consideration all soft deleted resources and making sure they are purged for the above reason.

I'd be inclined to have purge protection as an exception for the moment, until we agree on the alternative, since this may take some time to be implemented and we shouldn't block keyvault e2e validation meanwhile. This may also be the case for other resources as well.
@AlexanderSehr @ChrisSidebotham what's your take on that?

avm/utilities/pipelines/staticValidation/psrule/.ps-rule/min-suppress.Rule.yaml
rahalan: @jtracey93 @eriqua the reason was, that aa AZ for just one VM makes no sense and even may create a feeling of better SLA, that is not provided.

avm/utilities/pipelines/staticValidation/psrule/ps-rule.yaml
ChrisSidebotham: I agree with the point your making @eriqua, until we can recover vaults this will cause a headache. Maybe something for us to align with PG on. for example Terraform handles this by recovering the purged vault on "Creation of the Resource" albeit this causes some confusion but would be nice to be handled at the layer above

jtracey93: okay so if i added the exclusion back for the key vault are we good to progress with this PR?

eriqua: For this comment yes, I'll leave the resolution of the other one to @rahalan

ChrisSidebotham: I see no reason why not @eriqua ?

ChrisSidebotham (approved):
lgtm

avm/utilities/pipelines/staticValidation/psrule/ps-rule.yaml
AlexanderSehr: I swear I've seen a soft-deleted resource recovery even for at least one ARM resource (kvlt) - can't remember which one though. @jtracey93 if you know somebody who knows somebody that can make this happen for kvlt, that would be great - yet I assume the PGs knows about this request and just did not implement it (yet).

In any case, the only way to get around this for now would be to autogenerate the name in the test run. Unless you run the test file deployment twice, this would work even with purge protection (like we did for all the encryption key vaults). However, we're already collecting key vaults like candy. @eriqua would you know if we can keep the implement in the test files as is but add a supression in the test files e.g. above the name parameter? This would enable us to remove the rule from the central file. But I'm not sure if PSRule supports this.

eriqua: Let's discuss alternatives to purge enablement offline and I'd agree to merge the remaining changes meanwhile.

PR #1399, opened 2024-03-23
feat: Updated Git tag handling to be more robust

AlexanderSehr:
## Description

- Made sure pipeline fails if git tag is not correcly published
- Added post-publishing test that fails if the git tag is not found
- Tested fail state locally as having a git tag rejected isn't exactly easy to provoke

## Pipeline Reference

| Pipeline |
| -------- |
|  [working run](https://github.com/AlexanderSehr/bicep-registry-modules/actions/runs/8401389568/job/23009717955)  |

## Type of Change

  - [ ] Update to documentation


avm/utilities/pipelines/publish/Confirm-ModuleIsPublished.ps1
```diff
@@ -36,6 +42,16 @@ function Confirm-ModuleIsPublished {
     $retry_seconds = 60
     $index = 0
 
+    #######################################
```
eriqua, line 45: @AlexanderSehr if I get it correctly, in case the tag is not published:
- We make the publishing job fail so that we are aware of the missing tag before the publish index pipeline fails
- We are checking that as a last step after the validate publishing. Hence we are still allowing the module to be published even if the tag is not created
- In that case the workaround so far is to create the tag manually

Is my understanding correct? If so, looks neat 👌

eriqua (approved):
🆗

avm/utilities/pipelines/publish/Confirm-ModuleIsPublished.ps1
AlexanderSehr: Hey @eriqua, yes-ish
- If the tag creation fails, the publish step fails BEFORE it runs the `bicep publish`
- If both worked, the post-validation will check again and fail if it cannot find the tag afterall. However this state should never happen. It's not much more than a fails safe at this point.

So if a tag creation were to fail - you would actually not neat to create it manually but instead either resolve the issue or just re-run the failed publish job in case it was a temporal issue.

eriqua: Would you then block the publishing on the tag creation, although they are not really related? 
I'm a bit hesitant about that but it would also make sense to avoid fixing through manual intervention. 
Especially considering this is very rare to happen. So fine for me. Feel free to resolve

AlexanderSehr: The `manual` part is exactly the reason why I implemented it that way. I was also not sure initially - but re-running the failed job seemed better to me than your aforementioned manual intervention.

PR #1401, opened 2024-03-24
feat: if module list is not in sync with CSV file an issue is created

rahalan:
This code includes a pipeline, that runs daily and checks, if the module list drop down (for module issues) is up-to-date. If not, an issue is created.


.github/workflows/avm.platform.sync-avm-modules-list.yml
```diff
@@ -0,0 +1,36 @@
+name: "avm.platform.sync-avm-modules-list"
```
AlexanderSehr, line 1:
```suggestion
name: .Platform - Sync AVM module list
```

in alignment with changes that will be introduced by #1411

.github/workflows/avm.platform.sync-avm-modules-list.yml
```diff
@@ -0,0 +1,36 @@
...
+  sync-avm-modules-list:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - name: "Checkout"
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: tibdex/github-app-token@v2
```
eriqua, line 18: Are we good with leveraging external actions in this repo?

rahalan: @eriqua yes. We discussed with Alex already. This action is already used by Jack and Chris.

ChrisSidebotham: Resolving as already addressed

ChrisSidebotham (approved):

PR #1402, opened 2024-03-24
feat: Add additional logic if module issue is created

rahalan:
More fine grained logic, regarding commenting and assigning new module issues


rahalan:
Wrong files

PR #1403, opened 2024-03-24
feat: Add additional logic if workflow fails and issue is created

rahalan:
More fine grained logic, regarding commenting and assigning new failing workflow issues


ChrisSidebotham (approved):
lgtm

PR #1411, opened 2024-03-25
feat: Integrated BRM with AVM workflows

AlexanderSehr:
## Description

- Aligned workflow file prefix to `platform.*` 
- Aligned workflow naming to `Platform -` 
- Retired workflows that have been used by the `/modules` folder
- [x] TODO: Cleanup BRM scripts folder

## Type of Change

  - [ ] Update to documentation


.github/workflows/platform.check.psrule.yml
```diff
@@ -1,4 +1,4 @@
-name: "avm.platform.check.psrule"
+name: Platform - Check PSRule
```
eriqua, line 1: If we think this to be up for discussion, what about we remove the platform workflows renaming from the PR and just prioritize the removal of BRM platform workflows, since that is already agreed?

AlexanderSehr: Reaching out interenally to see if we can agree quickly - if not - will undo and also perform the reminaing changes.

eriqua: In general I'm fine with the rename, as long as module workflows keep the alignment with the module name.

If we want to keep platform pipelines on top in the UI, as done in CARML as well, I'd suggest this small change
```suggestion
name: .Platform - Check PSRule
```

eriqua (approved):
🆗

PR #1427, opened 2024-03-26
fix: WAF reliability and static validation in `avm/res/web/site` 

### LINKED ISSUE #1378: [Failed pipeline] avm.res.web.site
Failed run: https://github.com/Azure/bicep-registry-modules/actions/runs/8378764209

### LINKED ISSUE #1510: [Failed pipeline] avm.res.web.site
Failed run: https://github.com/Azure/bicep-registry-modules/actions/runs/8507795545

### LINKED ISSUE #1592: [Failed pipeline] avm.res.web.site
Failed run: https://github.com/Azure/bicep-registry-modules/actions/runs/8565946787

krbar:
## Description

- added PSRule exceptions for `defaults` tests
- removed optional parameters from `defaults` tests
- re-generated readme due to static validation errors

fixes #1378 
fixes #1510 
fixes #1592 

## Pipeline Reference

| Pipeline |
| -------- |
|  [avm.res.web.site](https://github.com/krbar/bicep-registry-modules/actions/workflows/avm.res.web.site.yml) |

## Type of Change

  - [ ] Update to documentation

## Checklist


avm/utilities/pipelines/staticValidation/psrule/.ps-rule/min-suppress.Rule.yaml
```diff
@@ -28,6 +28,10 @@ spec:
     # Azure Virtual Machine
     - Azure.VM.AMA
     - Azure.VM.Standalone
+    # Azure App Service
+    - Azure.AppService.AlwaysOn
```
eriqua, line 32: For each suppressed rule, please add a comment explaining the reason why the rule should be suppressed for defaults test.
I know other rules do not have the comment, but I'd strongly suggest we start doing it.

krbar: @eriqua Would be "because it's failing the default test" a sufficient explanation? ;)
But seriously, can you give me an example of a good explanation?
In this case both `alwaysOn` and `healthCheckPath` are part of the `siteConfig`, which also supports thousand of other settings, so it's difficult to set a default value there...

eriqua: As a rule of thumb, we should convert to Required all parameters needed to make the defaults test WAF compliant. 

The general exception is if those parameters are involving external dependencies, for example private endpoints. In that case, we add the corresponding rule to the suppression group for defaults, but make sure the rule is compliant for WAF-aligned tests, where the test is made compliant leveraging also optional parameters.

This case we have here is another one for which we need to agree on the best way to manage, i.e., when parameters required to make the test WAF compliant are part of a bigger object for which defining a custom type may be an overkill. 

In general, either we agree to suppress them or we brainstorm on how to make them required subproperties. 
I'm wondering if, for example, adding `alwaysOn` and `healthCheckPath` as individual input required parameters and build a logic to add them to `siteConfig` using union may help. 
Cc @AlexanderSehr @ChrisSidebotham @jtracey93

eriqua: @AlexanderSehr I agree on all the above, but this case is still not covered. It's the infamous config input object of web sites, having different properties depending on the type. Some of those subproperties are required for the test to be WAF reliability compliant. We cannot mark them Conditional. Yet, they are not external dependencies. So it seems to me that we are still missing guidance on this case.

Anyway. As per the original comment, if we agree on suppressing a rule for the defaults test, we should start adding a reason for that in the suppression file.

AlexanderSehr: ah ok, gotcha. Thanks for elaborating again (@eriqua). Yes it would be hard to impossible to assume values on a user's behalf if the object is that different. That makes the fact that we cannot have defaults in UDTs even more heartbreaking as this means we also cannot solve this in the future.
Agreeing with all @eriqua wrote above.

krbar: I set a default siteConfig with alwaysOn:
```bicep
param siteConfig object = {
  alwaysOn: true
}
```
which eliminated one of the suppressions. For the others, in my opinion there is no good default as the probe paths are depending on the app implementation. I commented on the suppression accordingly.

krbar: @eriqua @AlexanderSehr Is there anything still missing here. Are you fine to merge it?

tsc-buddy (approved):

avm/utilities/pipelines/staticValidation/psrule/.ps-rule/min-suppress.Rule.yaml
AlexanderSehr: No, plus @tsc-buddy already approved. Quite curious that the owning team was not set as a reviewer :o

AlexanderSehr (approved):

avm/utilities/pipelines/staticValidation/psrule/.ps-rule/min-suppress.Rule.yaml
AlexanderSehr: @eriqua, please resolve if resolved given that it is your comment (:

eriqua: Agree with the changes. Resolving

PR #1466, opened 2024-03-29
feat: Add the creation of 'Microsoft.CognitiveServices/accounts/deployments' in Cognitive Service Module - `avm/res/cognitive-services/account`

### LINKED ISSUE #1131: [AVM Module Issue]: Cognitive Service should add the creation of 'Microsoft.CognitiveServices/accounts/deployments'

### Issue Type?

Feature Request

### Module Name

avm/res/cognitive-services/account

### (Optional) Module Name if not listed above

avm/res/cognitive-services/account

### (Optional) Module Version

_No response_

### Description

Please add the creation of 'Microsoft.CognitiveServices/accounts/deployments'.
For example (azd): https://github.com/Azure/azure-dev/blob/main/templates/common/infra/bicep/core/ai/cognitiveservices.bicep#L38

CC: @jongio

### (Optional) Correlation Id

_No response_

zedy-wj:
## Description

Fixes #1131 

## Pipeline Reference

| Pipeline |
| -------- |
| [avm.res.cognitive-services.account](https://github.com/zedy-wj/bicep-registry-modules/actions/workflows/avm.res.cognitive-services.account.yml)         |

## Type of Change

- [x ] Azure Verified Module updates:
  - [ ] Update to documentation

## Checklist

- [x ] I'm sure there are no other open Pull Requests for the same update/change

@jongio - for notification.


avm/res/cognitive-services/account/main.bicep
```diff
@@ -640,3 +659,17 @@ type lockType = {
...
+
+type deploymentsType = {
+  @description('Optional. Specify the name of cognitive service account deployment.')
+  name: string?
+
+  @description('Required. Properties of Cognitive Services account deployment model.')
+  model: object
+
+  @description('Optional. The resource model definition representing SKU')
+  sku: object?
```
jongio, line 671: Should SKU be a custom type as well?

avm/res/cognitive-services/account/main.bicep
```diff
@@ -640,3 +659,17 @@ type lockType = {
   @description('Optional. Specify the type of lock.')
   kind: ('CanNotDelete' | 'ReadOnly' | 'None')?
 }?
+
+type deploymentsType = {
+  @description('Optional. Specify the name of cognitive service account deployment.')
+  name: string?
+
+  @description('Required. Properties of Cognitive Services account deployment model.')
+  model: object
```
jongio, line 668: Should model be a custom type?

AlexanderSehr (changes requested):
Hey @zedy-wj,

thank you for your PR. 2 asks if I may:
- Could you add the edployment of the new deployments parameter to e.g. the `max` test?
- Could you add a pipeline badge showing that the deployment works and also all static tests pass?

cc: @ilhaan

avm/res/cognitive-services/account/main.bicep
AlexanderSehr: Good idea. Note @zedy-wj UDTs can be be nested, you don't need a seperate type.

avm/res/cognitive-services/account/main.bicep
AlexanderSehr: Good idea. Note @zedy-wj UDTs can be be nested, you don't need a seperate type.

zedy-wj:
@AlexanderSehr - I have updated `max` test. But I encountered an error during Static validation:
```
at $filesAreTheSame | Should -Be $true -Because ('The file hashes before and after applying the `Set-ModuleReadMe` function should be identical and should not have diff </br><pre>{0}</pre>. Please re-run the script for this module''s template.' -f $mdFormattedDiff), /home/runner/work/bicep-registry-modules/bicep-registry-modules/avm/utilities/pipelines/staticValidation/compliance/module.tests.ps1:327
```

I tried running the `Set-ModuleReadMe` function from local but that didn't seem to work.
image

Did I do something wrong? Could you take a look at it, thanks!

AlexanderSehr:
> @AlexanderSehr - I have updated `max` test. But I encountered an error during Static validation:
> 
> [avm.res.cognitive-services.account](https://github.com/zedy-wj/bicep-registry-modules/actions/workflows/cog-services.yml)
> ```
> at $filesAreTheSame | Should -Be $true -Because ('The file hashes before and after applying the `Set-ModuleReadMe` function should be identical and should not have diff </br><pre>{0}</pre>. Please re-run the script for this module''s template.' -f $mdFormattedDiff), /home/runner/work/bicep-registry-modules/bicep-registry-modules/avm/utilities/pipelines/staticValidation/compliance/module.tests.ps1:327
> ```
> 
> I tried running the `Set-ModuleReadMe` function from local but that didn't seem to work.
> image
> 
> Did I do something wrong? Could you take a look at it, thanks!

Well would you look at that - I should update the tests description.

You did one thing wrong in that dot-sourcing a function requires an invocation with leading separate dot: `. <path>`

That being said, I'd recommend to invoke the Set-AVMModule function you can find in the avm/utilities/tools folder. You can invoke it for the module's folder path and should be good to go

zedy-wj:
> Well would you look at that - I should update the test's description (ref #1489).
> 
> You did one thing wrong in that dot-sourcing a function requires an invocation with leading separate dot: `. <path>`
> 
> That being said, I'd recommend to invoke the Set-AVMModule function you can find in the avm/utilities/tools folder. You can invoke it for the module's folder path and should be good to go

@AlexanderSehr - Sounds great. It worked for me! Thanks again for your help!

Here is the status badge, all tests passed:

[avm.res.cognitive-services.account](https://github.com/zedy-wj/bicep-registry-modules/actions/workflows/avm.res.cognitive-services.account.yml)

If there are no other questions, could you help me merge it? Thanks!

avm/res/cognitive-services/account/tests/e2e/max/main.test.bicep
```diff
@@ -185,3 +185,28 @@ module testDeployment '../../../main.bicep' = [
     ]
   }
 ]
+
+module testAIModelDeployment '../../../main.bicep' = {
```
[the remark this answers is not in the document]
AlexanderSehr: Aside from that, nothing to complain about with the rest of the contribution. Great work @zedy-wj 💪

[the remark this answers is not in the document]
zedy-wj: Okay, I added an `ai-model-deployment` folder, below is the latest status badge, please check it:

[avm.res.cognitive-services.account](https://github.com/zedy-wj/bicep-registry-modules/actions/workflows/avm.res.cognitive-services.account.yml)

zedy-wj:
@AlexanderSehr - Updated according to your comments. Below is the latest status badge, please check it:

[avm.res.cognitive-services.account](https://github.com/zedy-wj/bicep-registry-modules/actions/workflows/avm.res.cognitive-services.account.yml)

If there are no other questions, could you help me merge it? Thanks again! 😃

avm/res/cognitive-services/account/tests/e2e/ai-model-deployment/main.test.bicep
```diff
@@ -0,0 +1,63 @@
+targetScope = 'subscription'
+
+metadata name = 'Using deployments parameter set'
+metadata description = 'This instance deploys the module with ai model deployment feature.'
```
AlexanderSehr, line 4:
```suggestion
metadata name = 'Using `deployments` in parameter set'
metadata description = 'This instance deploys the module with the AI model deployment feature.'
```

AlexanderSehr: Please run the `Set-AVMModule` script one last time after this update.

zedy-wj: Updated!

AlexanderSehr: Much appreciated. All good from my side. Thanks

zedy-wj:
@AlexanderSehr - Thank you very much for your continued help!

@ilhaan - Could you take a look at these changes. If there are no other questions, could you help me merge it? Thanks!

avm/res/cognitive-services/account/main.bicep
```diff
@@ -339,6 +342,22 @@ resource cognitiveService 'Microsoft.CognitiveServices/accounts@2023-05-01' = {
...
+resource cognitiveService_deployments 'Microsoft.CognitiveServices/accounts/deployments@2023-05-01' = [
+  for (deployment, index) in (deployments ?? []): {
+    parent: cognitiveService
+    name: deployment.?name ?? '${name}-deployments'
+    properties: {
+      model: deployment.model
+      raiPolicyName: deployment.?raiPolicyName ?? null
+    }
+    sku: deployment.?sku ?? {
+      name: 'Standard'
```
ilhaan, line 355: According to the following doc, this value should be a letter+number code: https://learn.microsoft.com/en-us/azure/templates/microsoft.cognitiveservices/accounts/deployments?pivots=deployment-language-bicep#sku

I believe `Standard` would be the correct values for the `tier` param, which is not required.

avm/res/cognitive-services/account/main.bicep
```diff
@@ -339,6 +342,22 @@ resource cognitiveService 'Microsoft.CognitiveServices/accounts@2023-05-01' = {
...
+      capacity: 20
```
ilhaan, line 356: Why is this 20 by default?

ilhaan (changes requested):

zedy-wj:
@ilhaan - Updated according to your comments. Please re-review it, thanks!

avm/res/cognitive-services/account/main.bicep
```diff
@@ -339,6 +342,21 @@ resource cognitiveService 'Microsoft.CognitiveServices/accounts@2023-05-01' = {
...
 }
 
+@batchSize(1)
+resource cognitiveService_deployments 'Microsoft.CognitiveServices/accounts/deployments@2023-05-01' = [
+  for (deployment, index) in (deployments ?? []): {
+    parent: cognitiveService
+    name: deployment.?name ?? '${name}-deployments'
+    properties: {
+      model: deployment.model
+      raiPolicyName: deployment.?raiPolicyName ?? null
```
AlexanderSehr, line 352:
```suggestion
      raiPolicyName: deployment.?raiPolicyName
```
Redundant. The `??` selects the first non-null entry. If the former is null, the extra null doesn't do much ;)

If updated please also re-run the `Set-AVMModule` script to re-generate the templates.

zedy-wj: Updated!

AlexanderSehr:
One last one from my side @zedy-wj, there is a conflict which is very easy to resolve but needs resolving. To do so please
- Disable actions in your fork (to not trigger any workflows)
- Merge latest upstream
- Enable workflows again
- Merge latest main into this PR's branch (this should show the conflict)
- Accept whatever change (it doesn't matter as conflicting json file is autogenerated)
- Re-run the Set-AVMModule function on the module
- Push

Aside from that, would be grat @ilhaan if you could take a final look and ideally approve the PR :)

zedy-wj:
@AlexanderSehr - Updated according to your comments, thanks!

ilhaan (approved):

zedy-wj:
@AlexanderSehr - @ilhaan has already approved, could you help me merge it? Thanks!

AlexanderSehr (approved):

AlexanderSehr:
Thanks for the notification @zedy-wj , just updated folloing @ilhaan's approval

AlexanderSehr:
Hey @zedy-wj, we may have a bit of a problem / something I'd need your advice about

The PR added a new test for deployments, which is great. The caveat - it doesn't work in our subscription due to the error `"The current subscription does not have feature required by this model 'Format: OpenAI, Name: gpt-35-turbo, Version: 0301'."`.
Now, I tried to look up if it's just a feature that needs registering, but it seems we need to fill out a forms to actually sign the subscription up? Can you confirm? If so, this would impact not only us, but also the subscriptions of the owners and we should try to find an alternative test.

For the moment this means the deployment did fail and the new module version is **not** published.

cc: @ilhaan

zedy-wj:
@AlexanderSehr - Okay, I suspect that the subscription used in the test does not support this feature. Could you help me verify the above guess in the test subscription?

Steps:
1. Open [azure portal](https://ms.portal.azure.com/#home).
2. Create an Azure OpenAI ( choose your resource group and region, fill in name and all other by default).
image
3. Navigate to the OpenAI you just created and click `Model Deployment`, then continue to click `Manage Deployments`
image
4. Click New deployments, please confirm what deployment options are available to you.


When choosing the OpenAI model, I specifically chose `gpt-35-turbo - 0301` due to the limitation of the West Europe region.

AlexanderSehr:
Hey @zedy-wj,
thanks for elaborating :)

@ilhaan, is that ok with you? Personally, I don't know the CongitiveService resource well enough to judge if there is an easier alternative. If there is no easier solution, I'd strongly suggest to add the above instructions to the deployments test as a comment to help others that face the same issue.

Just putting this out there to ensure the service only so hard to maintain as absolutely necessary 😉 

cc: @eriqua, @ChrisSidebotham, @jtracey93 for awareness. Exciting times 😄

ilhaan:
@AlexanderSehr Can we just pick a model we know is available in all regions without extra registration steps?

AlexanderSehr:
> @AlexanderSehr Can we just pick a model we know is available in all regions without extra registration steps?

Right, that's the ask. Does anybody know of one? 😄

ilhaan:
@AlexanderSehr Sorry misunderstood your question earlier. 
 
@jceval any ideas?

jongio:
It's likely due to the sub not being on the Azure OpenAI allow list.  You can find out more here: https://learn.microsoft.com/en-us/legal/cognitive-services/openai/limited-access


You can request here: https://aka.ms/oai/access

jongio:
Unfortunately  there's no way around it.  How about we disable the test by default?

AlexanderSehr:
Ok, so the resource itself is just for OpenAI. In that case your suggestion may be the way to go. @ilhaan how do you feel about that? Truth be told, I don't exactly like the idea of having code in main that is not tested.
But if the only other alternative is to add the AVM subscription alongside every contributor's / owner's subscription via a process to some list I can't say it's much more appealing either.

cc: @eriqua, @ChrisSidebotham, @jtracey93 (please share if you have some thoughts about this too)

Maybe we're in luck and sometime soon this onboarding isn't necessary anymore? 😄

ilhaan:
@AlexanderSehr I am not a fan of either option, but if I had to pick one, I would go with @jongio's suggestion of disabling the test by default.

AlexanderSehr:
Hey @ilhaan,
alright let's do that then. Mind you, there is no 'native' way to do this, so we have get creative and decide for the best option. The following things could do what we want to achieve
- We could remove the test folder (the test is gone and removed from the readme too)
- We could rename the deployment test's `main.test.bicep` to something like `disabled_main.test.bicep` or similar as the CI logic searches for `main.test.bicep` files specifically (would mean the test is not found and not added to the readme)
- We could change the logic inside the test file to only run the deployment test on an impossible condition (would run the test file but not the test and still render it in the readme. The downside: it's hard to see it's disabled / is not validated, even if we add a dislaimer comment.
- ...

cc: @eriqua, @ChrisSidebotham

eriqua:
> Hey @ilhaan, alright let's do that then. Mind you, there is no 'native' way to do this, so we have get creative and decide for the best option. The following things could do what we want to achieve
> 
> * We could remove the test folder (the test is gone and removed from the readme too)
> * We could rename the deployment test's `main.test.bicep` to something like `disabled_main.test.bicep` or similar as the CI logic searches for `main.test.bicep` files specifically (would mean the test is not found and not added to the readme)
> * We could change the logic inside the test file to only run the deployment test on an impossible condition (would run the test file but not the test and still render it in the readme. The downside: it's hard to see it's disabled / is not validated, even if we add a dislaimer comment.
> * ...
> 
> cc: @eriqua, @ChrisSidebotham

Sorry, for me having a test file which is not continuously tested is really the last option. It's too likely to get stale and providing incorrect examples is way worse than not providing them at all.

That said, what about the following:
- Provide instructions on how to set up the environment in the documentation, as done for AADDS
- Short term: consider the test as an accepted known failure when coming from contributors. They can either set up the environment correctly following provided instructions or not
- Long term (effort to be estimated) update the CI to only run the test in upstream, or, for example, depending on a github variable switch. So that we provide contributors with an option to enable it on their side.

jongio:
Can we have a flag like "Microsoft Test" or something like that.  So it is run by our CI, but not by every contributor?

zedy-wj:
@AlexanderSehr - Is there anything I can do currently to fix CI so that new version can be published?

jtracey93:
I think we just need to get this module published as AI is a hot topic right now. So disabling the test to get the new version out would be my suggestion for short term.

Longer term what @eriqua suggested seems pragmatic and a already used approach.

I think right now this option suggest by you @AlexanderSehr:

> We could rename the deployment test's main.test.bicep to something like disabled_main.test.bicep or similar as the CI logic searches for main.test.bicep files specifically (would mean the test is not found and not added to the readme)

Whilst changing the readme generation logic to also include `disabled_main.test.bicep` files is the best compromise. 

**I'm also agreeing with @jongio on changing the prefix from `disabled ` to something else like `msft_ci_only_......` **

AlexanderSehr:
> Hi team. We already use these models in our deployments and need to rely on our own modules until this pr gets merged.
> 
> We always engage customers with a pre-req for a subscription with OpenAI enabled (https://aka.ms/oai/access) - in my opinion that is similar to having to register a resource provider.
> 
> This module is the last one we need to be able to provision our entire solution on AVM. I really wish you decide to enable OpenAI on the subscription you use for CI, so we can get the final bit of our template done. After all, it can't be too long before OpenAI becomes just another resource that can be activated on a subscription.
> 
> Under all circumstances, the documentation should align with what is actually available form the current version of the module. The official documentation currently states, that we can do this: (..)
> 
> But when we try to do that in our deployment, we get the old module from the registry, that does not contain the deployments part of the template module: (..)
> 
> Our IaC superman had to track the error all the way to this repo before he could figure out, why the module didn't work.
> 
> In the readme help for our private registry, we add extra information like ekstra requirements in the readme. Something similar could be done on this component. A simple Info box would do.

100% agree. It's the very unfortunate sideeffect of the PR being merged without the pipeline going through. Something we'll only be able to prevent in the future by enabling contributors to test in our environment (which is on the roadmap, but is no low-hangning fruit)

AlexanderSehr:
@zedy-wj will be working on a PR now

AlexanderSehr:
Hey @zedy-wj, @jongio & @Agazoth,
please note that the publishing went through and you should be able to use the feature now as intended: 
image

PR #1584, opened 2024-04-06
feat: Expand usage of UDT for the AVNM Bicep Module - `avm/res/network/network-manager`

ahmadabdalla:
## Description

Introducing User Defined Types for the Azure Virtual Network Manager (AVNM) Bicep module. Additionally, expanded on the test cases for the maxium to increase test coverage for UDT validations.

## Pipeline Reference

| Pipeline |
| -------- |
|   [avm.res.network.network-manager](https://github.com/ahmadabdalla/bicep-registry-modules/actions/workflows/avm.res.network.network-manager.yml)       |

## Type of Change

  - [ ] Update to documentation

## Checklist


avm/res/network/network-manager/connectivity-configuration/main.bicep
```diff
@@ -25,7 +25,7 @@ param appliesToGroups array = []
 param connectivityTopology string
 
 @sys.description('Conditional. List of hub items. This will create peerings between the specified hub and the virtual networks in the network group specified. Required if connectivityTopology is of type "HubAndSpoke".')
-param hubs array = []
+param hubs hubsType?
```
AlexanderSehr, line 28:
```suggestion
param hubs hubsType
```

You don't need to make both the parameter and type nullable ;)

avm/res/network/network-manager/main.bicep
```diff
@@ -240,3 +238,158 @@ type roleAssignmentType = {
...
+
+    @sys.description('Required. Resource ID of the virtual network.')
+    resourceId: string
+  }[]?
+}[]?
+
+type networkManagerScopeAccessType = ('Connectivity' | 'SecurityAdmin')[]
+
+type networkManagerScopesType = {
+  @sys.description('Optional. Management groups to assign to the network manager to manage.')
```
AlexanderSehr, line 262: I wonder if this would be a good case for `conditional` - in like 'Conditional. (...). Required if no Subscriptions were provided.')` 

On that note - Are these mangement group & subscriptions Ids? If so, I think we should either call this out - or even name the input accordingly to be explicit. I, for my part, "think" you mean the ID, but maybe it's the name? :D

avm/res/network/network-manager/main.bicep
```diff
@@ -240,3 +238,158 @@ type roleAssignmentType = {
...
+
+    @sys.description('Required. Resource type of the hub.')
+    resourceType: 'Microsoft.Network/virtualNetworks'
+  }[]?
+
+  @sys.description('Optional. Delete existing peering connections.')
+  deleteExistingPeering: ('True' | 'False')?
+
+  @sys.description('Optional. Is global configuration.')
+  isGlobal: ('True' | 'False')?
```
AlexanderSehr, line 321: There is actually a new [spec](https://github.com/Azure/Azure-Verified-Modules/pull/803) waiting in a PR that asks primitive types to be declared as such (e.g. use int instead if string for numbers, or boolean instead of string for true/false).
This would mean, this parameter should be changed to `isGlobal: bool` and the place where it is used needs to (optionally) cast the value to a string. I've seen implicit casts in the pasts, so we may not need it.

Same comment for the other string-boolean parameters

avm/res/network/network-manager/security-admin-configuration/main.bicep
```diff
@@ -59,3 +55,70 @@ output resourceId string = securityAdminConfigurations.id
...
+  @sys.description('Required. The name of the admin rule collection.')
+  name: string
+
+  @sys.description('Optional. A description of the admin rule collection.')
+  description: string?
+
+  @sys.description('Required. List of network groups for configuration. An admin rule collection must be associated to at least one network group.')
+  appliesToGroups: {
+    @sys.description('Required. The resource ID of the network group.')
+    networkGroupId: string
```
AlexanderSehr, line 75: Should ideally be `networkGroupResourceId` though I wonder what a network group is to begin with. A `network security group`? If so, I recommend to call it such.

Please note, if changed to e.g. `networkSecurityGroupResourceId`, also the location where it is used must be updated wth a map function to map it to `the expected `networkGroupId`

AlexanderSehr: Also mind you, as you implemented the UDT in mutliple files, it would also need to be fixed in multiple files.

Looking forward to the time when types can simply be imported via a types.bicep ([ref](https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/bicep-import))

avm/res/network/network-manager/network-group/main.bicep
```diff
@@ -12,10 +12,10 @@ param name string
 
 @maxLength(500)
 @sys.description('Optional. A description of the network group.')
-param description string = ''
+param description string?
 
 @sys.description('Optional. Static Members to create for the network group. Contains virtual networks to add to the network group.')
-param staticMembers array = []
+param staticMembers staticMembersType?
```
AlexanderSehr, line 18: Only either the parameter or the type must be nullable

avm/res/network/network-manager/main.bicep
ahmadabdalla: That's a good proposal. I would agree and align with the `Conditional` approach. These are resource IDs:

```json
            "managementGroups": [
                "/providers/Microsoft.Management/managementGroups/<mgId>"
            ],
            "subscriptions": [
                "/subscriptions/<subId>
            ],
```

I will call it out on the parameter to be more explicit.

avm/res/network/network-manager/main.bicep
ahmadabdalla: Interesting. I would avoid hiding such things and align to the API as much as possible. That way its a standard consumption method from an API and IaC perspective and less abstraction code to maintain. But if that's the path that AVM is going to take then I will wait for that PR to be merged. Would be great to know if that's something that will be as a recommendation for the module owner or something that is a must have.

AlexanderSehr: I'd argue it's probably a 'Should' that targets improved usability of the resource, which in many instances is just plain bad (case an point CMK or boolean strings 😄 )

avm/res/network/network-manager/main.bicep
AlexanderSehr: Can't see the update yet

ahmadabdalla: Updated

avm/res/network/network-manager/network-group/main.bicep
ahmadabdalla: Updated

avm/res/network/network-manager/main.bicep
ahmadabdalla: Hey @AlexanderSehr I have done some testing on the casting approach.

image

Here is an [example PR](https://github.com/ahmadabdalla/bicep-registry-modules/pull/14/files#diff-2b0d9ee4256af0d2862695cca181bbca439dd6beb31013e1bb33c8b0b8b9f546R44): comparing the current UDT implementation as per this PR with an approach for casting.

I def took a toll on testing and validation a lot more than it was before. And seems more code to maintain. What are your thoughts on this?

I have some questions around the `default` behaviour for a UDT param of type `bool`. Because I had to use casting, I now am enforced in my test files to explicitly pass the parameter either `true` `false` compared to the past.  Is there a specific pattern I should follow here as per AVM standards?

image

See above picture as an example. If I do not provide these values, my deployments now fail :/

avm/res/network/network-manager/security-admin-configuration/rule-collection/rule/main.bicep
```diff
@@ -29,10 +29,10 @@ param description string = ''
 param access string
 
 @sys.description('Optional. List of destination port ranges. This specifies on which ports traffic will be allowed or denied by this rule. Provide an (*) to allow traffic on any port. Port ranges are between 1-65535.')
-param destinationPortRanges array = []
+param destinationPortRanges destinationPortRangesType?
```
AlexanderSehr, line 32: Why the UDT if it's anyways just a string array?

```suggestion
param destinationPortRanges string[]?
```

avm/res/network/network-manager/security-admin-configuration/rule-collection/rule/main.bicep
```diff
@@ -58,10 +58,10 @@ param priority int
 param protocol string
 
 @sys.description('Optional. List of destination port ranges. This specifies on which ports traffic will be allowed or denied by this rule. Provide an (*) to allow traffic on any port. Port ranges are between 1-65535.')
-param sourcePortRanges array = []
+param sourcePortRanges sourcePortRangesType?
```
AlexanderSehr, line 61: Same comment as before
```suggestion
param sourcePortRanges string[]?
```

avm/res/network/network-manager/security-admin-configuration/rule-collection/rule/main.bicep
```diff
@@ -58,10 +58,10 @@ param priority int
...
 
 @sys.description('Optional. The source filter can be an IP Address or a service tag. Each filter contains the properties AddressPrefixType (IPPrefix or ServiceTag) and AddressPrefix (using CIDR notation (e.g. 192.168.99.0/24 or 2001:1234::/64) or a service tag (e.g. AppService.WestEurope)). Combining CIDR and Service tags in one rule filter is not permitted.')
-param sources array = []
+param sources sourcesType?
```
AlexanderSehr, line 64:
```suggestion
param sources sourcesType
```

Type is already nullable

avm/res/network/network-manager/security-admin-configuration/rule-collection/main.bicep
```diff
@@ -35,28 +35,28 @@ resource ruleCollection 'Microsoft.Network/networkManagers/securityAdminConfigur
...
       networkManagerName: networkManager.name
       securityAdminConfigurationName: securityAdminConfigurationName
       ruleCollectionName: ruleCollection.name
       name: rule.name
       access: rule.access
-      description: contains(rule, 'description') ? rule.description : ''
-      destinationPortRanges: contains(rule, 'destinationPortRanges') ? rule.destinationPortRanges : []
-      destinations: contains(rule, 'destinations') ? rule.destinations : []
+      description: rule.?description ?? ''
+      destinationPortRanges: rule.?destinationPortRanges ?? []
```
AlexanderSehr, line 53: Just a suggestion (so feel free to just resolve), but you could actually leave the else case out when referencing the child module. 

Example
- If in the child module `param destinationPortRanges` has a default in the child, and this line here is just `destinationPortRanges: rule.?destinationPortRanges`, then it will automatically pick the default in the child
- If in the child module `param destinationPortRanges` has no default but is nullable like in your case, then no value is passed through (which I guess also is fine)

avm/res/network/network-manager/main.bicep
AlexanderSehr: Hey @ahmadabdalla,
maybe we can find a slot some time to actually have a chat about the subject, as it's just easier to talk about than write 😄 
Anyhow, the modules I've seen implement this on scale did usually not introduce much added code. Usually it was not much more than introducing e.g. a `string(parameter)` or `'${parameter}'` where a parameter was used an a string was expected, while in others it made the module actually easier as conditions didn't need to check for explicit string values in conditions like `if(parameter == 'true')`. But naturally, the argument you made (i.e., deviating from the API) was exactlythe discussion to have. In like 'is it worth it'. At least for now we opted for 'yes' on the backdrop that we want to improve/easy a deployments usuability when we can.

Regarding the default behavior - if a parameter is optional, you can still just make it nullable. Also booleans. Does this answer the question?

ahmadabdalla: Got everything working now. Ready for review

avm/res/network/network-manager/security-admin-configuration/main.bicep
ahmadabdalla: I would like to propose keeping it as, unless it's a hard requirement.

AlexanderSehr: Hey @ahmadabdalla, it would be as per the [spec](https://azure.github.io/Azure-Verified-Modules/specs/shared/#id-snfr22---category-inputs---parametersvariables-for-resource-ids) and align with the other modules. But if you want, I can ask the team how 'strict'  we want to be about the rule, and if we can ignore it in cases.

avm/res/network/network-manager/connectivity-configuration/main.bicep
```diff
@@ -66,3 +66,29 @@ output resourceId string = connectivityConfiguration.id
...
+  @sys.description('Optional. Flag if use hub gateway.')
+  useHubGateway: bool?
+}[]
+
+type hubsType = {
+  @sys.description('Required. Resource Id of the hub.')
+  resourceId: string
+
+  @sys.description('Required. Resource type of the hub.')
+  resourceType: 'Microsoft.Network/virtualNetworks'
```
AlexanderSehr, line 93: Do understand it right that this is the only allowed type? If so, would it not make sense to just make it optional and set it to set value by default?

avm/res/network/network-manager/main.bicep
```diff
@@ -240,3 +238,158 @@ type roleAssignmentType = {
...
+  staticMembers: {
+    @sys.description('Required. The name of the static member.')
+    name: string
+
+    @sys.description('Required. Resource ID of the virtual network.')
+    resourceId: string
+  }[]?
+}[]?
+
+type networkManagerScopeAccessType = ('Connectivity' | 'SecurityAdmin')[]
```
AlexanderSehr, line 259: Just curious, is this the same as

```bicep
@allowed([
  'Connectivity'
  'SecurityAdmin'
])
param networkManagerScopeAccesses string[]
```
?

avm/res/network/network-manager/connectivity-configuration/main.bicep
ahmadabdalla: Currently yes but given how its implemented. Appears like potentially supporting something else down the track like virtual wan. See [documentation](https://learn.microsoft.com/en-us/azure/virtual-network-manager/faq#can-an-azure-virtual-wan-hub-be-part-of-a-network-group).

avm/res/network/network-manager/main.bicep
ahmadabdalla: Yes

avm/res/network/network-manager/security-admin-configuration/main.bicep
jtracey93: Hey @ahmadabdalla, is there a reason to want to keep it as `networkGroupId` instead of aligning to the spec?

ahmadabdalla: @AlexanderSehr @jtracey93 this has now been addressed and tested

AlexanderSehr (approved):

PR #1687, opened 2024-04-16
fix: Adding conditions on app setttings config - `avm/res/web/site`

JFolberth:
## Description

Updated 
- API Version for sites 2022-09-01 -> 2023-01-01
- PSRule Updates
- WAF Updates
- API Version for hybridConnectionNamespaces/relays 2022-9-01 -> 2023-01-01
- private-endpoint: 0.4.0 -> 0.4.1
- ReadMEs Appropriately Updated

## Pipeline Reference

| Pipeline |
| - |
| [avm.res.web.site](https://github.com/JFolberth/bicep-registry-modules/actions/workflows/avm.res.web.site.yml) |
          
## Type of Change

  - [ ] Update to documentation

## Checklist


tsc-buddy:
Hi @JFolberth Thanks for submitting a PR. Can you please complete the description with some context around the change, why, what it addresses/improves. 

Also, would it be possible for you to run our validation pipeline for these module changes and add a status badge to the PR description? You can find instructions for doing this here: https://azure.github.io/Azure-Verified-Modules/contributing/bicep/bicep-contribution-flow/#1-setup-your-azure-test-environment.

JFolberth:
@tsc-buddy  will do. I have also been updating the APIs and PSRule tests as noticed in the pipeline it was giving recommendations.

JFolberth:
@tsc-buddy  I think this is ready for review....Everything is passing but fail on avm publish

avm/res/web/site/tests/e2e/waf-aligned/main.test.bicep
```diff
@@ -17,6 +17,12 @@ param resourceLocation string = deployment().location
 @description('Optional. A short identifier for the kind of deployment. Should be kept short to not run into resource-name length-constraints.')
 param serviceShort string = 'wswaf'
 
+@description('Optional. Tags of the resource.')
+param tags object = {
```
AlexanderSehr, line 21: While it's a good thought, please move the value down into the test as a value. Reason being that the `Set-AVMModule` script carves the entire test block out of this file, formatts it and puts it into the readme. Anything that is not inside the test block is hence ignored. However, as we want to show as much as possible in the `Usage Examples` it's pretty much the opposite of what we'd want. 💪

avm/res/web/site/tests/e2e/webApp.defaults/main.test.bicep
```diff
@@ -57,9 +57,15 @@ module testDeployment '../../../main.bicep' = [
       siteConfig: {
         healthCheckPath: '/healthz'
         alwaysOn: true
+        ftpsState: 'FtpsOnly'
```
AlexanderSehr, line 60: The values you added - are they all now mandatory with the latest update, or did you set them because PSRule was complaining about them?

I strongly suspect the latter. There is a spec in AVM somewhere for it, but to avoid looking it up, let me elaborate here
- The `waf-aligned` tests should contain every parameter we'd need to deploy the module in a waf-compliant way (and ideally no more)
- The `defaults` test in contrast should deploy the module with only the required parameters so that the defaults are used. These defaults should hence be WAF-compliant.
Now here's the catch: That would mean the latter may fail some PSRules. To address those there are 2 options:
- If it's a simple value that can have a default (e.g., `localAuth: false`) we should set it in the template file
- If not - then we should add it to the suppresion rules for the default tests you can find [here](https://github.com/Azure/bicep-registry-modules/blob/main/avm/utilities/pipelines/staticValidation/psrule/.ps-rule/min-suppress.Rule.yaml). We don't exactly advertise it, as it should be a 'last resort'

Now here's the double-catch. When we applied this strategy we soon found a handful of cases where one may wonder if some of the rules should be applied in general. I think you found one of those which says 'resource xzy should have a managed identity' or similar. The thought of the rule is good - if 2 services should communicate they should use a managed identity. But not every service you deploy communicates outbound, so what's the point of enforcing a managed identity? These would either also go into the suppresion rules as you can see for example here
https://github.com/Azure/bicep-registry-modules/blob/b5d76b9c9931ba130db315643edcfc064e855bd2/avm/utilities/pipelines/staticValidation/psrule/.ps-rule/min-suppress.Rule.yaml#L17
and or should be called out in the PSRule repository as a rule that should be reconsidered. For example [here](https://github.com/Azure/Azure-Verified-Modules/issues/349#issuecomment-2061504618). 
We're collaborating closely with @BernieWhite to provide feedback and help with the WAF-compliance tests, and naturally, also your feedback is well received.

AlexanderSehr: Same feedback for the `webAppLinux.defaults` test

JFolberth: @AlexanderSehr for App Service I feel this should be required...there aren't many scenarios where a web service isn't contacting some data source.

I agree the defaults should be updated...would consider this case the some of these are just good practice...minTLS, HTTP2, SFTP. The question/concern/feedback before making a change to this is who/whom/how should determine this?

avm/res/web/site/tests/e2e/waf-aligned/main.test.bicep
JFolberth: Update made and README updated per c1ff5453e7602fcd17b5dfd13814a9acd3f17c8f

avm/res/web/site/tests/e2e/webApp.defaults/main.test.bicep
AlexanderSehr: That would be the module owner(s). In this case Sir @tsc-buddy to the rescue.

Processwise, module owners (unless orphaned) are automatically added to PRs to their modules to review. The only cases where the CI maintainers may bypasss this are batch updates (if we find for example an issue with an implementation existing in many modules that we fix at once).

AlexanderSehr: Also beware a sizable amount of conflicts for this PR (likely due to the recently merged fixed for the WAF-reliablity tests).

My advice
- Disable actions in your fork's settings
- Merge upstream & resolve conflicts
- Enable actions again

As your PR comes directly from main, you may need to do the conflict resolution via your local termal + VSCode though. I doubt GitHub will be keen on merging upstream in the current setup.

JFolberth: Conflicts are resolved now

avm/res/web/site/tests/e2e/waf-aligned/main.test.bicep
```diff
@@ -17,6 +17,9 @@ param resourceLocation string = deployment().location
 @description('Optional. A short identifier for the kind of deployment. Should be kept short to not run into resource-name length-constraints.')
 param serviceShort string = 'wswaf'
 
+@description('Optional. Tags of the resource.')
+param tags object?
+
```
AlexanderSehr, line 22:
```suggestion
```

Not used

JFolberth:
@AlexanderSehr is there anything need from me to get this change committed?

JFolberth:
@AlexanderSehr  I am going to close this and nuke my main, repull and properly set this up via branching as learned a lot from the first time doing this :) Weren't that many changes to the specific module.

AlexanderSehr:
> @AlexanderSehr I am going to close this and nuke my main, repull and properly set this up via branching as learned a lot from the first time doing this :) Weren't that many changes to the specific module.

Happy to have a session some time to show you ways to avoid the nuke 😉

PR #1696, opened 2024-04-18
feat: Replications and WAF alignment for App Store - `avm/res/app-configuration/configuration-store`

### LINKED ISSUE #1502: [WAF - Reliability Review] - `avm/res/app-configuration/configuration-store`

### Issue Type?

Feature Request

### Module Name

avm/res/app-configuration/configuration-store

### (Optional) Module Name if not listed above

_No response_

### (Optional) Module Version

_No response_

### Description

Hello team,

The AVM core team are performing a review of all AVM modules to ensure they are aligned to the WAF reliability pillar by default, and in the future they will be checking against other WAF pillars also. This is a requirement we have before we can go `v1.0.0` for any AVM module.

Links to AVM specs:

- https://azure.github.io/Azure-Verified-Modules/specs/shared/#id-sfr2---category-composition---waf-aligned
- https://azure.github.io/Azure-Verified-Modules/specs/shared/#id-sfr5---category-composition---availability-zones

Links to WAF Reliability, APRL & Product Docs recommendations used for analysis:

- https://learn.microsoft.com/azure/well-architected/reliability/checklist
- https://learn.microsoft.com/en-us/azure/azure-app-configuration/faq?toc=%2Fazure%2Freliability%2Ftoc.json&bc=%2Fazure%2Freliability%2Fbreadcrumb%2Ftoc.json#how-does-app-configuration-ensure-high-data-availability

Whilst reviewing this module the following issues have been found that are non-compliant with the AVM specs that need to be resolved in a new version release that can be seen in the PSRule for Azure run here: https://github.com/Azure/bicep-registry-modules/actions/workflows/avm.res.app-configuration.configuration-store.yml#summary-23300173139

> The geo-replica PSRule check may need a [suppression rule](https://azure.github.io/PSRule.Rules.Azure/concepts/suppression/) creating as it may only be suitable for a pattern module and not a resource module. The AVM `.ps-rule` directory can be found here:  https://github.com/Azure/bicep-registry-modules/tree/main/avm/utilities/pipelines/staticValidation/psrule/.ps-rule

Please make these changes in a PR to the parameter defaults so a new minor version can be published.

Thanks

The AVM Core Team

### (Optional) Correlation Id

_No response_

JFolberth:
## Description

- Added replicas
- Update settings per pester results
- ReadMes Update
- Configured replicas for WAF alignment
- Removed orphan Markdown

Fixes #1502
Closes #1502
-->

## Pipeline Reference

| Pipeline |
| -------- |
| [avm.res.app-configuration.configuration-store](https://github.com/JFolberth/bicep-registry-modules/actions/workflows/avm.res.app-configuration.configuration-store.yml)|

## Type of Change

  - [x] Update to documentation

## Checklist


JFolberth:
@matebarabas @eriqua 

Looking for feedback on this on two items.

I was not sure on replicas what to do with locations...I found avm\res\aad\domain-service\tests\e2e\waf-aligned\main.test.bicep where locations are hard coded and figured maybe this is how it should be handled?

I also per https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.AppConfig.AuditLogs/ Audit logs are required and I wasn't sure what to put for diagnostic log locations for the default testing. I left them all (Event Hub, Storage Account, Workspace). I felt this was heavy and not sure on guidance.

avm/res/app-configuration/configuration-store/version.json
```diff
@@ -1,7 +1,7 @@
 {
     "$schema": "https://aka.ms/bicep-registry-module-version-file-schema#",
-    "version": "0.1",
+    "version": "0.1.1",
```
eriqua, line 3: Hey @JFolberth, no need to update the patch version, since that in Bicep is done automatically by the CI.

However, as per semver, given the addition of a child resource, and with that a new feature, I'd suggest bumping the minor version
```suggestion
    "version": "0.2",
```

avm/res/app-configuration/configuration-store/tests/e2e/defaults/main.test.bicep
```diff
@@ -43,6 +56,17 @@ module testDeployment '../../../main.bicep' = [
     params: {
       name: '${namePrefix}${serviceShort}001'
       location: resourceLocation
+      enablePurgeProtection: true
```
eriqua, line 59: This can be a challenge in our testing, as we usually create an delete resources

eriqua:
> @matebarabas @eriqua
> 
> Looking for feedback on this on two items.
> 
> I was not sure on replicas what to do with locations...I found avm\res\aad\domain-service\tests\e2e\waf-aligned\main.test.bicep where locations are hard coded and figured maybe this is how it should be handled?
> 
> I also per https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.AppConfig.AuditLogs/ Audit logs are required and I wasn't sure what to put for diagnostic log locations for the default testing. I left them all (Event Hub, Storage Account, Workspace). I felt this was heavy and not sure on guidance.

For the last point, those rules are currently not blocking as they are not part of the WAF reliability pillar. I'm not saying we shouldn't care about them, just that are not a priority 1 and can be taken for example in a next PR. 

That said, as a rule of thumb, defaults tests should only cover required parameters, so audit logs rules can be exempted for defaults tests only. The exemption file specific for defaults tests is https://github.com/Azure/bicep-registry-modules/blob/main/avm/utilities/pipelines/staticValidation/psrule/.ps-rule/min-suppress.Rule.yaml

avm/res/app-configuration/configuration-store/tests/e2e/defaults/main.test.bicep
JFolberth: It is working in the test; however, the real failure is the order of operations as it feels the config store gets deleted first then the replica...which throws an error that the parent has already been deleted. I guess would ask how is this being accommodated for other resources who have purge protection such as Key Vault and Log Analytics?



I am not high on it being the default as it is not available in free tier and not sure on repeat logic as there already is logic to not allow if  `sku=='Free'` .I would suggest keeping it as false so to easily accommodate for more flexibility. 

`enablePurgeProtection: sku == 'Free' ? false : enablePurgeProtection`

eriqua: I totally see the angle @JFolberth. And believe me a wormhole likely opens every time the discussion flexibility/usability vs security is raised 😄 
When it comes to WAF and especially reliability by default, that generally wins over flexibility. So the idea is to make the resource easier to deploy with WAF compliance than without.

However in this case I think we can achieve both. The implemented logic you mentioned should also work with the default value set to `true`, and that would be used by default unless the sku is set to free. What about adding more context in the enablePurgeProtection @description? Something like: `defaults to true unless 'sku' is set to 'Free', since purge protection is not available in Free tier.`

avm/res/app-configuration/configuration-store/replicas/main.bicep
```diff
@@ -0,0 +1,31 @@
+metadata name = 'App Configuration Replica Configurations'
```
eriqua, line 1:
```suggestion
metadata name = 'App Configuration Replicas'
```

avm/res/app-configuration/configuration-store/tests/e2e/defaults/main.test.bicep
```diff
@@ -43,6 +45,9 @@ module testDeployment '../../../main.bicep' = [
...
+      disableLocalAuth: true
+      replicaLocations: ['eastus', 'westus']
```
eriqua, line 50: Defaults tests are supposed to only cover input parameters that are required. Ref https://azure.github.io/Azure-Verified-Modules/specs/bicep/#defaults-tests-must 

For this reason, I'd suggest to remove these optional parameters from this test, and add an exclusion to the defaults PSRule suppression group for the rule `Azure.AppConfig.GeoReplica`: https://github.com/Azure/bicep-registry-modules/blob/main/avm/utilities/pipelines/staticValidation/psrule/.ps-rule/min-suppress.Rule.yaml

Regarding `disableLocalAuth`, the PSRule Azure.AppConfig.DisableLocalAuth is not in the reliability pillar, thus non blocking for now. However, if we want to fix that already, I'd suggest to set its default value to `true` in the main.bicep template instead.

eriqua: For replicas, yes, absolutely. Ideally we should also add them to the max test, which usually covers all possible features and child resources to maximize parameter validation.

avm/res/app-configuration/configuration-store/tests/e2e/defaults/main.test.bicep
```diff
@@ -24,6 +24,8 @@ param namePrefix string = '#_namePrefix_#'
 // Dependencies //
 // ============ //
 
+// Diagnostics
+// ===========
```
eriqua, line 28:
```suggestion
```

eriqua:
> > @matebarabas @eriqua
> > Looking for feedback on this on two items.
> > I was not sure on replicas what to do with locations...I found avm\res\aad\domain-service\tests\e2e\waf-aligned\main.test.bicep where locations are hard coded and figured maybe this is how it should be handled?
> > I also per https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.AppConfig.AuditLogs/ Audit logs are required and I wasn't sure what to put for diagnostic log locations for the default testing. I left them all (Event Hub, Storage Account, Workspace). I felt this was heavy and not sure on guidance.
> 
> For the last point, those rules are currently not blocking as they are not part of the WAF reliability pillar. I'm not saying we shouldn't care about them, just that are not a priority 1 and can be taken for example in a next PR.
> 
> That said, as a rule of thumb, defaults tests should only cover required parameters, so audit logs rules can be exempted for defaults tests only. The exemption file specific for defaults tests is https://github.com/Azure/bicep-registry-modules/blob/main/avm/utilities/pipelines/staticValidation/psrule/.ps-rule/min-suppress.Rule.yaml

@JFolberth now finally replying on the first point :)

The CI uses a region rotation for the deployment. So every test deployment may target a different location randomly picked up from [this list](https://github.com/Azure/bicep-registry-modules/blob/main/avm/utilities/pipelines/e2eValidation/regionSelector/Get-AzAvailableResourceLocation.ps1#L36): `eastus`, `uksouth`, `northeurope` and `eastasia`. 
That is unless a location is hardcoded in the test file, like the example you mentioned for AADDS. In this case, I guess we can rely on the default logic for the primary resource, while we should hardcode the location for the replica in the test, as you did already. 

I'm not an expert of App Store, but I'd expect it is not allowed to use the same location for the primary resource and the replica. So if, for example, the rotation logic picks 'eastus' for the primary resource, using 'eastus' for the replica would probably fail the deployment. Is that correct? If so,  for the hardcoded replica location we should pick a set not including the 4 locations we are rotating from.

JFolberth:
> > > @matebarabas @eriqua
> > > Looking for feedback on this on two items.
> > > I was not sure on replicas what to do with locations...I found avm\res\aad\domain-service\tests\e2e\waf-aligned\main.test.bicep where locations are hard coded and figured maybe this is how it should be handled?
> > > I also per https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.AppConfig.AuditLogs/ Audit logs are required and I wasn't sure what to put for diagnostic log locations for the default testing. I left them all (Event Hub, Storage Account, Workspace). I felt this was heavy and not sure on guidance.
> > 
> > 
> > For the last point, those rules are currently not blocking as they are not part of the WAF reliability pillar. I'm not saying we shouldn't care about them, just that are not a priority 1 and can be taken for example in a next PR.
> > That said, as a rule of thumb, defaults tests should only cover required parameters, so audit logs rules can be exempted for defaults tests only. The exemption file specific for defaults tests is https://github.com/Azure/bicep-registry-modules/blob/main/avm/utilities/pipelines/staticValidation/psrule/.ps-rule/min-suppress.Rule.yaml
> 
> @JFolberth now finally replying on the first point :)
> 
> The CI uses a region rotation for the deployment. So every test deployment may target a different location randomly picked up from [this list](https://github.com/Azure/bicep-registry-modules/blob/main/avm/utilities/pipelines/e2eValidation/regionSelector/Get-AzAvailableResourceLocation.ps1#L36): `eastus`, `uksouth`, `northeurope` and `eastasia`. That is unless a location is hardcoded in the test file, like the example you mentioned for AADDS. In this case, I guess we can rely on the default logic for the primary resource, while we should hardcode the location for the replica in the test, as you did already.
> 
> I'm not an expert of App Store, but I'd expect it is not allowed to use the same location for the primary resource and the replica. So if, for example, the rotation logic picks 'eastus' for the primary resource, using 'eastus' for the replica would probably fail the deployment. Is that correct? If so, for the hardcoded replica location we should pick a set not including the 4 locations we are rotating from.

I am not overly familiar with it app stores as well. I have updated. My thought was with two to 1.) ensure we hit a region that the resource is not deployed to and 2.) ensure it can handle multiple locations. I have just updated it not to use `eastus` so this should be accomplished.

avm/res/app-configuration/configuration-store/tests/e2e/max/main.test.bicep
```diff
@@ -67,6 +67,7 @@ module testDeployment '../../../main.bicep' = [
       name: '${namePrefix}${serviceShort}001'
       location: resourceLocation
       createMode: 'Default'
+      replicaLocations: ['eastus', 'westus']
```
eriqua, line 70:
```suggestion
      replicaLocations: ['centralus', 'westus']
```

eriqua:
> > > > @matebarabas @eriqua
> > > > Looking for feedback on this on two items.
> > > > I was not sure on replicas what to do with locations...I found avm\res\aad\domain-service\tests\e2e\waf-aligned\main.test.bicep where locations are hard coded and figured maybe this is how it should be handled?
> > > > I also per https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.AppConfig.AuditLogs/ Audit logs are required and I wasn't sure what to put for diagnostic log locations for the default testing. I left them all (Event Hub, Storage Account, Workspace). I felt this was heavy and not sure on guidance.
> > > 
> > > 
> > > For the last point, those rules are currently not blocking as they are not part of the WAF reliability pillar. I'm not saying we shouldn't care about them, just that are not a priority 1 and can be taken for example in a next PR.
> > > That said, as a rule of thumb, defaults tests should only cover required parameters, so audit logs rules can be exempted for defaults tests only. The exemption file specific for defaults tests is https://github.com/Azure/bicep-registry-modules/blob/main/avm/utilities/pipelines/staticValidation/psrule/.ps-rule/min-suppress.Rule.yaml
> > 
> > 
> > @JFolberth now finally replying on the first point :)
> > The CI uses a region rotation for the deployment. So every test deployment may target a different location randomly picked up from [this list](https://github.com/Azure/bicep-registry-modules/blob/main/avm/utilities/pipelines/e2eValidation/regionSelector/Get-AzAvailableResourceLocation.ps1#L36): `eastus`, `uksouth`, `northeurope` and `eastasia`. That is unless a location is hardcoded in the test file, like the example you mentioned for AADDS. In this case, I guess we can rely on the default logic for the primary resource, while we should hardcode the location for the replica in the test, as you did already.
> > I'm not an expert of App Store, but I'd expect it is not allowed to use the same location for the primary resource and the replica. So if, for example, the rotation logic picks 'eastus' for the primary resource, using 'eastus' for the replica would probably fail the deployment. Is that correct? If so, for the hardcoded replica location we should pick a set not including the 4 locations we are rotating from.
> 
> I am not overly familiar with app stores as well. I have updated. My thought was with two to 1.) ensure we hit a region that the resource is not deployed to and 2.) ensure it can handle multiple locations. I have just updated it not to use `westus` and `centralus` so this should be accomplished.

Sounds great. Just added a suggestion for the other test file to do the same. Overall it looks good. As a final step, could you please 
- Rerun the set-AVMModule utility to update the Readmes
- Retrigger the pipeline manually to verify if the run works after enabling purge protection by default. Based on the result I may need to look into the CI deployment removal logic.

Thanks!

JFolberth:
> > > > > @matebarabas @eriqua
> > > > > Looking for feedback on this on two items.
> > > > > I was not sure on replicas what to do with locations...I found avm\res\aad\domain-service\tests\e2e\waf-aligned\main.test.bicep where locations are hard coded and figured maybe this is how it should be handled?
> > > > > I also per https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.AppConfig.AuditLogs/ Audit logs are required and I wasn't sure what to put for diagnostic log locations for the default testing. I left them all (Event Hub, Storage Account, Workspace). I felt this was heavy and not sure on guidance.
> > > > 
> > > > 
> > > > For the last point, those rules are currently not blocking as they are not part of the WAF reliability pillar. I'm not saying we shouldn't care about them, just that are not a priority 1 and can be taken for example in a next PR.
> > > > That said, as a rule of thumb, defaults tests should only cover required parameters, so audit logs rules can be exempted for defaults tests only. The exemption file specific for defaults tests is https://github.com/Azure/bicep-registry-modules/blob/main/avm/utilities/pipelines/staticValidation/psrule/.ps-rule/min-suppress.Rule.yaml
> > > 
> > > 
> > > @JFolberth now finally replying on the first point :)
> > > The CI uses a region rotation for the deployment. So every test deployment may target a different location randomly picked up from [this list](https://github.com/Azure/bicep-registry-modules/blob/main/avm/utilities/pipelines/e2eValidation/regionSelector/Get-AzAvailableResourceLocation.ps1#L36): `eastus`, `uksouth`, `northeurope` and `eastasia`. That is unless a location is hardcoded in the test file, like the example you mentioned for AADDS. In this case, I guess we can rely on the default logic for the primary resource, while we should hardcode the location for the replica in the test, as you did already.
> > > I'm not an expert of App Store, but I'd expect it is not allowed to use the same location for the primary resource and the replica. So if, for example, the rotation logic picks 'eastus' for the primary resource, using 'eastus' for the replica would probably fail the deployment. Is that correct? If so, for the hardcoded replica location we should pick a set not including the 4 locations we are rotating from.
> > 
> > 
> > I am not overly familiar with app stores as well. I have updated. My thought was with two to 1.) ensure we hit a region that the resource is not deployed to and 2.) ensure it can handle multiple locations. I have just updated it not to use `westus` and `centralus` so this should be accomplished.
> 
> Sounds great. Just added a suggestion for the other test file to do the same. Overall it looks good. As a final step, could you please
> 
> * Rerun the set-AVMModule utility to update the Readmes
> * Retrigger the pipeline manually to verify if the run works after enabling purge protection by default. Based on the result I may need to look into the CI deployment removal logic.
> 
> Thanks!

All updated!

avm/res/app-configuration/configuration-store/tests/e2e/defaults/main.test.bicep
JFolberth: Done! I also had to add one to the min rule on Audit Logs.

avm/res/app-configuration/configuration-store/tests/e2e/defaults/main.test.bicep
eriqua: Unresolved this, since purge protection enablement in test files is currently blocking the validation pipeline. My suggestion is still the one of the first comment of this thread:

- Disable the feature in all tests and add a comment next to the assignment mentioning that is not WAF compliant but only an exception for deployment validation
- Add an exception for this rule in the psrule setting file, as done for keyvault for the same reason https://github.com/Azure/bicep-registry-modules/blob/main/avm/utilities/pipelines/staticValidation/psrule/ps-rule.yaml#L78

avm/res/app-configuration/configuration-store/main.bicep
```diff
@@ -26,10 +26,10 @@ param sku string = 'Standard'
 param createMode string = 'Default'
 
 @description('Optional. Disables all authentication methods other than AAD authentication.')
-param disableLocalAuth bool = false
+param disableLocalAuth bool = true
 
-@description('Optional. Property specifying whether protection against purge is enabled for this configuration store.')
-param enablePurgeProtection bool = false
+@description('Optional. Property specifying whether protection against purge is enabled for this configuration store.Defaults to true unless sku is set to Free, since purge protection is not available in Free tier.')
```
eriqua, line 31: Sorry, just noticed this minor one
```suggestion
@description('Optional. Property specifying whether protection against purge is enabled for this configuration store. Defaults to true unless sku is set to Free, since purge protection is not available in Free tier.')
```

JFolberth: Thanks @eriqua I have committed this change.

eriqua: Hey @JFolberth thanks. We are very close! Please run the Set-AVMModule utility to reflect the change in the main.json as well. Also please rerun the pipeline so that it can hopefully have a green badge

eriqua (approved):
🆗

PR #1719, opened 2024-04-19
fix: Recovered missing role assignment workflow

AlexanderSehr:
## Description

Recovered missing role assignment workflow


eriqua (approved):
:OK:

PR #1751, opened 2024-04-25
fix: minor bugs in worflow automation regarding failed pipelines

rahalan:
Fixes:
- not the full team of the bicep team was used, making the @ mentioning not working
- wrong comment was created, regarding the owner, if owner assignment failed due to GH bug


avm/utilities/pipelines/platform/Set-AvmGitHubIssueOwnerConfig.ps1
```diff
@@ -42,10 +42,6 @@ function Set-AvmGitHubIssueOwnerConfig {
   if ($issue.title.StartsWith('[AVM Module Issue]')) {
     $moduleName = ($issue.body.Split("`n") -match 'avm/(?:res|ptn)')[0].Trim().Replace(' ', '')
 
-    if ([string]::IsNullOrEmpty($moduleName)) {
```
eriqua, line 45: Why removing this?

rahalan: @eriqua  Máté asked for it, as it is already covered by a different logic (you cannot select a non-existing module).

eriqua: How can we make sure this is run only on the specific issue template where the condition is met by default? I guess it would be enough to create an issue with title starting with `[AVM Module Issue]` to enter this condition, isn't that the case?

rahalan: @eriqua  Good point. I reverted it back.

eriqua (approved):
🆗

PR #1772, opened 2024-04-29
fix: Allow empty parameters in subnet `avm/res/network/virtual-network`

### LINKED ISSUE #1681: [AVM Module Issue]: How to conditionally define network security group in subnets

### Issue Type?

I'm not sure

### Module Name

avm/res/network/virtual-network

### (Optional) Module Version

0.1.1

### Description

I'm unable to find a way to conditionally define network security group for a subnet. For example I would like to define NSG id if the subnet's name is AzureBastionSubnet but not for others. The most obvious approach would be:

    subnets: [for subnet in subnets: {
      name: subnet.name
      addressPrefix: subnet.addressPrefix
      networkSecurityGroupResourceId: if(subnet.name == 'AzureBastionSubnet') ? nsg.id : null
    }]

But it seems that the value if set can't be null:
```Cannot parse the request. (Code:
     | InvalidRequestFormat)    - Value for the id property is invalid.
     | Expecting a string. Actual value is Null. Path
     | properties.subnets[0].properties.networkSecurityGroup.
     | (Code:InvalidJsonPropertyType)

### (Optional) Correlation Id

_No response_

### PR DESCRIPTION
**Pull request by cloudchristoph**
## Description

You cannot provide the following parameters in your subnet parameter, if they are empty:

- networkSecurityGroup
- natGateway
- routeTable

As soon as they are set, they have to have a correct resource id. 

This PR implements additional checks for given, but empty, parameters.
I also added a new subnet definition in the `max` test to check for the correct behaviour.

Fixes #1681
Closes #1681

## Pipeline Reference

| Pipeline |
| -------- |
| [avm.res.network.virtual-network](https://github.com/cloudchristoph/bicep-registry-modules/actions/workflows/avm.res.network.virtual-network.yml) |

## Type of Change

- [ ] Update to CI Environment or utilities (Non-module effecting changes)
  - [x] Update to documentation

## Checklist

### DIFF
--- a/avm/res/network/virtual-network/main.bicep
+++ b/avm/res/network/virtual-network/main.bicep
@@ -130,12 +130,12 @@ resource virtualNetwork 'Microsoft.Network/virtualNetworks@2023-04-01' = {
             : []
           delegations: contains(subnet, 'delegations') ? subnet.delegations : []
           ipAllocations: contains(subnet, 'ipAllocations') ? subnet.ipAllocations : []
-          natGateway: contains(subnet, 'natGatewayResourceId')
+          natGateway: contains(subnet, 'natGatewayResourceId') && !empty(subnet.natGatewayResourceId)
             ? {
                 id: subnet.natGatewayResourceId
               }
             : null
-          networkSecurityGroup: contains(subnet, 'networkSecurityGroupResourceId')
+          networkSecurityGroup: contains(subnet, 'networkSecurityGroupResourceId') && !empty(subnet.networkSecurityGroupResourceId)
             ? {
                 id: subnet.networkSecurityGroupResourceId
               }
@@ -146,7 +146,7 @@ resource virtualNetwork 'Microsoft.Network/virtualNetworks@2023-04-01' = {
           privateLinkServiceNetworkPolicies: contains(subnet, 'privateLinkServiceNetworkPolicies')
             ? subnet.privateLinkServiceNetworkPolicies
             : null
-          routeTable: contains(subnet, 'routeTableResourceId')
+          routeTable: contains(subnet, 'routeTableResourceId') && !empty(subnet.routeTableResourceId)
             ? {
                 id: subnet.routeTableResourceId
               }
--- a/avm/res/network/virtual-network/README.md
+++ b/avm/res/network/virtual-network/README.md
@@ -190,6 +190,13 @@ module virtualNetwork 'br/public:avm/res/network/virtual-network:<version>' = {
       }
       {
         addressPrefix: '<addressPrefix>'
+        name: 'az-subnet-x-004'
+        natGatewayResourceId: ''
+        networkSecurityGroupResourceId: ''
+        routeTableResourceId: ''
+      }
+      {
+        addressPrefix: '<addressPrefix>'
         name: 'AzureBastionSubnet'
         networkSecurityGroupResourceId: '<networkSecurityGroupResourceId>'
       }
@@ -318,6 +325,13 @@ module virtualNetwork 'br/public:avm/res/network/virtual-network:<version>' = {
           "networkSecurityGroupResourceId": "<networkSecurityGroupResourceId>",
           "privateEndpointNetworkPolicies": "Disabled",
           "privateLinkServiceNetworkPolicies": "Enabled"
+        },
+        {
+          "addressPrefix": "<addressPrefix>",
+          "name": "az-subnet-x-004",
+          "natGatewayResourceId": "",
+          "networkSecurityGroupResourceId": "",
+          "routeTableResourceId": ""
         },
         {
           "addressPrefix": "<addressPrefix>",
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by cloudchristoph** [comment]
> Hi @cloudchristoph , would you mind to add the !empty also for: privateEndpointNetworkPolicies and privateLinkServiceNetworkPolicies? To set the subnet within the vnet the same way we have inside the subnet module (
> 
> https://github.com/Azure/bicep-registry-modules/blob/0509c539a2fc1c9f48cfb9f7f85c117c19be896b/avm/res/network/virtual-network/subnet/main.bicep#L104
> 
> ) . Thank you very much

Sure. But for this parameter and the others you're already allowed to give an empty array or `null` as value and it still works. This wasn't the case for the three changed parameters.

Correct me, if I'm wrong.

I would like to ask one more question: 
What will be allowed as an as “not set parameter”? Now it is a mixture of `''`, `null`, and `[]`.

File: avm/res/network/virtual-network/main.bicep [hunk 1]
```diff
@@ -130,12 +130,12 @@ resource virtualNetwork 'Microsoft.Network/virtualNetworks@2023-04-01' = {
...
           delegations: contains(subnet, 'delegations') ? subnet.delegations : []
           ipAllocations: contains(subnet, 'ipAllocations') ? subnet.ipAllocations : []
-          natGateway: contains(subnet, 'natGatewayResourceId')
+          natGateway: contains(subnet, 'natGatewayResourceId') && !empty(subnet.natGatewayResourceId)
             ? {
                 id: subnet.natGatewayResourceId
               }
             : null
-          networkSecurityGroup: contains(subnet, 'networkSecurityGroupResourceId')
+          networkSecurityGroup: contains(subnet, 'networkSecurityGroupResourceId') && !empty(subnet.networkSecurityGroupResourceId)
```
AlexanderSehr, line 138: I'm not sure if this is a great idea. I recall we used to have a related discussion about empty checks a long time ago in CARML, and the main catch was that as soon as you add this check, you're moving yourself into a corner.

To be precise, you can, for example, always set the networkSecurityGroup, but once you did, you can never unset it again (back to no network security group). Please correct me if I'm wrong.

cloudchristoph: It does remove the NSG.

My first module call:
```bicep
module vnet '../avm/res/network/virtual-network/main.bicep' = {
  name: 'vnet-add'
  scope: rg
  params: {
    name: 'vnet-tests'
    location: rg.location
    addressPrefixes: ['10.0.0.0/24']
    subnets: [
      {
        name: 'default'
        addressPrefix: '10.0.0.0/26'
        networkSecurityGroupResourceId: nsg.outputs.resourceId
      }
    ]
  }
}
```
Result:
image


My second module call:
```bicep
module vnet '../avm/res/network/virtual-network/main.bicep' = {
  name: 'vnet-add'
  scope: rg
  params: {
    name: 'vnet-tests'
    location: rg.location
    addressPrefixes: ['10.0.0.0/24']
    subnets: [
      {
        name: 'default'
        addressPrefix: '10.0.0.0/26'
        networkSecurityGroupResourceId: ''
      }
    ]
  }
}
```
Result:
image


But after your comments yesterday, I'm ultimately undecided whether this is the right way to go. 🤷‍♂️
Your solution using the union() function also works, even if I think it's not that "discoverable" or easy to use by end users.

Let me know what you think.

cloudchristoph: Forgot the third test - Removing the parameter.

After re-applying the NSG with the first template, I've used this module call and it also has removed the NSG:
```bicep
module vnet '../avm/res/network/virtual-network/main.bicep' = {
  name: 'vnet-add'
  scope: rg
  params: {
    name: 'vnet-tests'
    location: rg.location
    addressPrefixes: ['10.0.0.0/24']
    subnets: [
      {
        name: 'default'
        addressPrefix: '10.0.0.0/26'
      }
    ]
  }
}
```

AlexanderSehr: Hey @cloudchristoph,
thanks for the due dilligence. That's good to see. In this case I'd pivot towards your suggestion as it may add some code to the template and is ultimately not needed when using the `union()`, yet from where I'm standing it's simply adding another case handling (i.e., somebody wants to provide an empty value) without taking anything away. 

What do you think?

cloudchristoph: Yes, it doesn't take away any functionality, is backwards compatible and simplifies writing.

For now, there is nothing I would add. In the future maybe we could write tests that test both adding and removing route tables / NSGs etc.

AlexanderSehr: There certainly is an option. A test template could deploy one instance with and without the parameter and as we support post-deployment pester tests, we could validate that a subnet in question has a property set or not.

AlexanderSehr:
Hey @elbatane, the PR looks good to me. Please also take a look :)

elbatane (approved):

PR #1808, opened 2024-05-01
feat: New Module `avm/ptn/finops-toolkit/finops-hub`

sebassem:
## Description
A pattern module for the [finops-hub](https://github.com/microsoft/finops-toolkit)

## Pipeline Reference

| Pipeline |
| -------- |
|  [avm.ptn.finops-toolkit.finops-hub](https://github.com/sebassem/bicep-registry-modules/actions/workflows/avm.ptn.finops-toolkit.finops-hub.yml)        |

## Type of Change

  - [ ] Update to documentation

## Checklist


avm/ptn/finops-toolkit/finops-hub/main.bicep
```diff
@@ -0,0 +1,222 @@
...
+    tags: resourceTags
+    tagsByResource: tagsByResource
+  }
+}
+
+//------------------------------------------------------------------------------
+// Key Vault for storing secrets
+//------------------------------------------------------------------------------
+
+module keyVault 'modules/keyVault.bicep' = {
```
ChrisSidebotham, line 138: My question here, Ideally we have patterns built on top of resource modules. would we not be better to leverage the AVM Modules instead of custom module calls for the top level resources. I understand we have a need to create resource before the nested loop creation can take place but I fear especially with WAF & Resiliency alignment we are losing the benefit of the AVM Atomic Modules by not using them

ChrisSidebotham: @sebassem - Apologies this was supposed to be surrounding the data factory object not the KV

sebassem: If I understood correctly, this comment is about using the dataFactory resource and not the AVM module for it , correct? If yes, the problem is that there is a property that is not yet in the schema "PipelineBillingEnabled". If there is a way to add it in the AVM module, I can make the switch

ChrisSidebotham: @clintgrove - Can you take a look at this so we can have this ptn module leverage your DF Module

sebassem: Commenting to bump this ask so we can finalize the module :)

clintgrove: I agree with Seif, the use of any() is necessary here to add PipelineBillingEnabled and there is slim chance that this will be needed by more than even 5% of our users of the AVM module so its best to use the Factory resource call here instead of the AVM module call to Factory

ChrisSidebotham: Happy to resolve this - thanks clint

ChrisSidebotham (approved):
LGTM

PR #1872, opened 2024-05-07
fix: Moved Databricks removal into explicit removal (before RG) for managed resource removal

### LINKED ISSUE #1581: [AVM Question/Feedback]: Databricks Connector not correctly removed

### Description

The `databricks/workspace` module creates an additional `-managed` resource group that is not removed by the CI. This can cause issues upon re-run if the location selector choses a different location to deploy to:

Example:
image

This means we need to investigate why this is the case, if there is a property we can set (similar to how you can specify on the VM that it should remove its NIC & PIP) or whether we can add something to the removal logic (e.g. adding it to the order in which resources are removed - to before the resource group is removed). 
```[tasklist]
### Tasks
```

AlexanderSehr:
## Description

Moved Databricks removal into explicit removal (before RG) for managed resource removal

Fixes #1581

## Pipeline Reference

| Pipeline |
| -------- |
|   [avm.res.databricks.workspace](https://github.com/AlexanderSehr/bicep-registry-modules/actions/workflows/avm.res.databricks.workspace.yml)       |

## Type of Change

  - [ ] Update to documentation


avm/utilities/pipelines/e2eValidation/resourceRemoval/Initialize-DeploymentRemoval.ps1
```diff
@@ -92,6 +92,7 @@ function Initialize-DeploymentRemoval {
             'Microsoft.Compute/virtualMachines',
             'Microsoft.VirtualMachineImages/imageTemplates', # Must be removed before their MSI
             'Microsoft.ManagedIdentity/userAssignedIdentities',
+            'Microsoft.Databricks/workspaces',
```
eriqua, line 95: Tried and failed hard

AlexanderSehr: Like the pipeline failed? Or the RGs remained? My 3 tests so far worked, but each time the Databricks deployment itself had issues

eriqua: The RG remained. Let's maybe see if the behavior changed and this workaround works now. If still not, as discussed, we can work on a custom post removal logic.

AlexanderSehr: Re-running one last time in the hopes that it's not getting stuck on the dependency issue again - like for the last 4 times.

AlexanderSehr: Can't get around the issue during the deployment. Will close the PR for now

PR #1918, opened 2024-05-13
feat: update logic to create a meaningful issue, if the module list template is out of sync with the CSV files

rahalan:
If the template file is not in sync with the CSV files, an issue was created. The old issue was not very helpful, thus the new issue will show exactly what it is about and shows in detail the diff. Here you can see how such an issue will look like: https://github.com/Azure/GHPolicyTest/issues/189


matebarabas (commented):
Based on the [dummy issue](https://github.com/Azure/GHPolicyTest/issues/189) this solution created as part of its testing, the outcomes look good - thanks for incorporating my feedback. I'd leave the more thorough review of the code on @eriqua and @AlexanderSehr.

ChrisSidebotham (approved):

ChrisSidebotham:
> Based on the [dummy issue](https://github.com/Azure/GHPolicyTest/issues/189) this solution created as part of its testing, the outcomes look good - thanks for incorporating my feedback. I'd leave the more thorough review of the code on @eriqua and @AlexanderSehr.

@matebarabas - I reviewed it :)

PR #1931, opened 2024-05-14
feat: Added UDTs for Private and Public DNS zone records and fixed typos in PR template - `avm/res/network/dns-zone`, `avm/res/network/private-dns-zone`

johnlokerse:
## Description

This pull request adds user-defined types for the (private and public) DNS zone records. Updated the version.json to 0.3.
Also, fixed some typos in the PR template.

Tagging module owner: @ChrisSidebotham 

## Pipeline Reference

| Pipeline |
| -------- |
|[avm.res.network.dns-zone](https://github.com/johnlokerse/bicep-registry-modules/actions/workflows/avm.res.network.dns-zone.yml)|
|[avm.res.network.private-dns-zone](https://github.com/johnlokerse/bicep-registry-modules/actions/workflows/avm.res.network.private-dns-zone.yml)|

## Type of Change

- [ ] Update to CI Environment or utilities (Non-module effecting changes)
  - [x] Update to documentation

## Checklist


avm/res/network/private-dns-zone/main.bicep
```diff
@@ -294,3 +292,135 @@ type lockType = {
   @description('Optional. Specify the type of lock.')
   kind: ('CanNotDelete' | 'ReadOnly' | 'None')?
 }?
+
+type zoneBaseType = {
```
AlexanderSehr, line 296: Hey @johnlokerse,
I think I get why you introduced this, that is, to not repeat the same 4 properties over and over. Whily I appreciate the reduced code, I'm not sure I like the artificial property this introduces 😄 

I wonder if there is a way to union User Defined Types without added an extra property for it. @ChrisSidebotham / @eriqua would you know?

johnlokerse: Yeah, I get what you mean... the `base` property can be weird/confusing.

I don't think that there's a way to unionize it. I have talked about this implementation with @ChrisSidebotham and either this or adding `name`, `metadata` and `ttl` property is the way to go. The last option is what other modules have right now though.

Off-topic, what would be waaay cooler and useful (cc @alex-frankel 😉) is the inheritance of types like this for example:

```bicep
type zoneBaseType = {
  @description('Required. The name of the record.')
  name: string

  @description('Optional. The metadata of the record.')
  metadata: object?

  @description('Optional. The TTL of the record.')
  ttl: int?
}

type aType inherits zoneBaseType= {
  aRecords: {
    ipv4Address: string
  }[]
}[]?
```

AlexanderSehr: > Yeah, I get what you mean... the `base` property can be weird/confusing.
> 
> I don't think that there's a way to unionize it. I have talked about this implementation with @ChrisSidebotham and either this or adding `name`, `metadata` and `ttl` property is the way to go. The last option is what other modules have right now though.
> 
> Off-topic, what would be waaay cooler and useful (cc @alex-frankel 😉) is the inheritance of types which would be a use-case for this. Something like this for example:
> 
> ```bicep
> type zoneBaseType = {
>   @description('Required. The name of the record.')
>   name: string
> 
>   @description('Optional. The metadata of the record.')
>   metadata: object?
> 
>   @description('Optional. The TTL of the record.')
>   ttl: int?
> }
> 
> type aType inherits zoneBaseType= {
>   aRecords: {
>     ipv4Address: string
>   }[]
> }[]?
> ```

If that doesn't justify an issue in the Bicep repository, I don't know what is ;)

Anyhow, ultimately I'd argue it's the decision of the module owner, i.e., @ChrisSidebotham. 
Just saying if it were up to me, I'd recommend to keep the object flat (without the `base`) because it may be confusing to begin with, sure, but even more so because it may be even more confusing once we can import types for the bicep-types which certainly don't have this `base` property 😉

johnlokerse: Ping @ChrisSidebotham for visibility 😄

ChrisSidebotham: damn I certainly missed this! Did you open an issue on the bicep GH for the inherits? Funnily enough even copolition wants to do:
```bicep 
type arecord = base
& {
aRecords: string[]
ttl: int
}
```

I happy with the usage of `base` in the example with a hope to move to an `inherits` action later. Just thinking about naming though would something like `commonProperties` be better suited than `base`?

AlexanderSehr: That's the thing. If this feature happens to be implemented and an `inhertis` becomes available - users of the module would all need to remove the `base` property from they deployments as it's a breaking change (moving from the nested, to the flat structure). I just don't see the point in adding this non-existend property to reduce the amount of code by 15 lines or so 😄 

That said, regarding what you wrote (the snippet) @ChrisSidebotham, does that work, or was that just for illustration purposes? Because if it works, that would be great 😄

ChrisSidebotham: Just for illustration... from copilot... so copilot wants it to work! @alex-frankel / @shenglol is there a plan to have this fucntion of `type A inherits B` etc?

johnlokerse: @ChrisSidebotham @AlexanderSehr Raised the issue / requested feature here: https://github.com/Azure/bicep/issues/14135

johnlokerse: @AlexanderSehr @ChrisSidebotham We can go for the "standard" implementation for now (adding the base properties to each record type) and when the inherit feature request is reviewed and accepted we could revisit this?

ChrisSidebotham: @johnlokerse - We have also raised this via our PG Connections internally. Let's see what we get response wise then figure out an action plan!

eriqua: > @AlexanderSehr @ChrisSidebotham We can go for the "standard" implementation for now (adding the base properties to each record type) and when the inherit feature request is reviewed and accepted we could revisit this?

FWIW mid term I'd also vouch for a standard implementation with a flat object, also for consistency with other modules. 

The inheritance feature request looks great (thanks @johnlokerse for raising that) and we cannot expect that to be implemented short term. 

From the above discussion I understand the proposed base property may introduce a breaking change once the inherit feature or importing types become fully available.

If that is the case, open question: while I always appreciate reducing duplicated code, would the workaround be worth the effort?

AlexanderSehr: Fair point. Any thoughts @ChrisSidebotham? We can also have a quick chat for that matter 😄 We've been commenting for a while now

ChrisSidebotham: Agree with the points above - Waiting to hear back from Johnny (who I think has already been working on this)

johnlokerse: I agree too 😃 Have you guys heard anything internally about the (potential) inheritance feature?

AlexanderSehr: @ChrisSidebotham brought it up with the ARM PG to raise awareness :) Not sure if it went beyond that yet, but I wouldn't recommend to block the PR over it.

AlexanderSehr: Ah nice, thanks. The thinking is good though - I really hope the PG comes around with the feature.

johnlokerse: @ChrisSidebotham @AlexanderSehr Latest test run without base type is green: [avm.res.network.dns-zone](https://github.com/johnlokerse/bicep-registry-modules/actions/workflows/avm.res.network.dns-zone.yml) [avm.res.network.private-dns-zone](https://github.com/johnlokerse/bicep-registry-modules/actions/workflows/avm.res.network.private-dns-zone.yml)

AlexanderSehr: Resolved it is

ChrisSidebotham (approved):
lgtm

PR #1948, opened 2024-05-16
Use null value in publicNetworkAccess instead of empty string 

jceval:
## Description

Request addresses the conditional statement logic for network access status based on the values of publicNetworkAccess and networkAcls. The expression ensures that if publicNetworkAccess is not empty, that value is used; otherwise, if networkAcls is not empty, the status is considered enabled; otherwise, it’s disabled. This aligns with the intended behavior of prioritizing publicNetworkAccess over networkAcls for determining network access status.

This change makes it so that an empty string passed to the module, will fail the deployment. If the value is not passed, then a null value will be interpreted as the default 'Disabled' behavior. This can only be overridden by explicitly passing 'Enabled'.

## Pipeline Reference

| Pipeline |
| -------- |
|          |

## Type of Change

    - [ X ] The bug was found by the module author, and no one has opened an issue to report it yet.
  - [ ] Update to documentation

## Checklist

- [ X ] I'm sure there are no other open Pull Requests for the same update/change

### DIFF
--- a/avm/res/cognitive-services/account/main.bicep
+++ b/avm/res/cognitive-services/account/main.bicep
@@ -64,11 +64,10 @@ param diagnosticSettings diagnosticSettingType
 
 @description('Optional. Whether or not public network access is allowed for this resource. For security reasons it should be disabled. If not specified, it will be disabled by default if private endpoints are set and networkAcls are not set.')
 @allowed([
-  ''
   'Enabled'
   'Disabled'
 ])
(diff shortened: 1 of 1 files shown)


AlexanderSehr:
Hey @jceval, I take it it still doesn't work? 😄

jceval:
I just had a messy branch 😂
incoming other PR ....

PR #2017, opened 2024-05-23
fix: Only use isHnsEnabled property when explicitly specified in `avm/res/storage/storage-account`

### LINKED ISSUE #1522: [Question/Feedback]: Ignore (unchanged) read-only properties in Bicep during deployment

### Description

I'm using an AVM Bicep module to deploy a storage account that was created by a different language to deploy infrastructure.  I'm looking to migrate the infrastructure code to Bicep in combination with Azure deployment stacks to make our setup simpler and faster.

However, it seems like _because_ the resource already exists, it throws the following error:

```
Code: DeploymentFailed
Message: At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/arm-deployment-operations for usage details.
Target: /subscriptions/xxxx/resourceGroups/rg-xxx/providers/Microsoft.Resources/deployments/xxx
Exception Details:      (AccountPropertyCannotBeUpdated) The property 'isHnsEnabled' was specified in the input, but it cannot be updated as it is read-only. For more information, see - https://aka.ms/storageaccountupdate
```

I've tried setting the property `enableHierarchicalNamespace` for the `storage-account` AVM to `null`, `false`, `true`, and I've left it out entirely as well. But it always results into the same error.

I expect this same issue will arise when deploying other resources as well. It would help me if I can either ignore (unchanged) fields, or import resources somehow such that Azure at least thinks it has deployed that resource before.

There's [an open issue](https://github.com/Azure/azure-rest-api-specs/issues/18441) which is somewhat related, but this is about the actual Bicep resource, whereas my issue is with the Bicep AVM.

bartcode:
## Description

Instead of always providing the `isHnsEnabled` property, only provide it when the `enableHierarchicalNamespace` parameter to the `storage-account` module is set to `true` or `false`. Otherwise, leave it out.

Important notes:
I'm making the `enableHierarchicalNamespace` parameter a `bool?`. Please verify carefully whether I'm missing a scenario there which would lead to this being a breaking change. In the existing tests, the scenario of not setting the parameter (to `true`) and setting it is already covered.

This closes #1522 .

## Pipeline Reference

| Pipeline |
| -------- |
|  [avm.res.storage.storage-account](https://github.com/bartcode/bicep-registry-modules/actions/workflows/avm.res.storage.storage-account.yml)  |

## Type of Change

  - [ ] Update to documentation

## Checklist

### DIFF
--- a/avm/res/storage/storage-account/main.bicep
+++ b/avm/res/storage/storage-account/main.bicep
@@ -122,7 +122,7 @@ param allowBlobPublicAccess bool = false
 param minimumTlsVersion string = 'TLS1_2'
 
 @description('Conditional. If true, enables Hierarchical Namespace for the storage account. Required if enableSftp or enableNfsV3 is set to true.')
-param enableHierarchicalNamespace bool = false
+param enableHierarchicalNamespace bool?
 
 @description('Optional. If true, enables Secure File Transfer Protocol for the storage account. Requires enableHierarchicalNamespace to be true.')
 param enableSftp bool = false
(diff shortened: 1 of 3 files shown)


avm/res/storage/storage-account/main.bicep
```diff
@@ -327,95 +327,97 @@ resource storageAccount 'Microsoft.Storage/storageAccounts@2022-09-01' = {
...
+          }
+      allowBlobPublicAccess: allowBlobPublicAccess
+      publicNetworkAccess: !empty(publicNetworkAccess)
+        ? any(publicNetworkAccess)
+        : (!empty(privateEndpoints) && empty(networkAcls) ? 'Disabled' : null)
+      azureFilesIdentityBasedAuthentication: !empty(azureFilesIdentityBasedAuthentication)
+        ? azureFilesIdentityBasedAuthentication
+        : null
+    },
+    enableHierarchicalNamespace != null ? { isHnsEnabled: enableHierarchicalNamespace } : {}
```
AlexanderSehr, line 419: Hey @bartcode,
is it absolutely necessary to apply the union? The reason I'm asking is because adding a union has the unfortunate side-effect of removing all support for linter (i.e., suggesting parameters etc.).

Did just making the parameter nullable not work? In other words, did the API throw an error?
I.e., when `isHnsEnabled: enableHierarchicalNamespace` would resolve to `isHnsEnabled: null`

bartcode: @AlexanderSehr , unfortunately, that's the whole reason I'm making this PR. The error I'm getting, I'm mentioning in #1522 .

AlexanderSehr: Just a quick update @bartcode, I'm still looking into this. Sorry that it hasn't been faster, but I want you to know that the PR is not forgotten :)

bartcode: Thanks for the update! There's no hurry. Just wanted to make a contribution. I have disabled my Azure subscription now, though. Because somehow this little thing here led to a 300 EUR Azure bill, which I'm not too happy with.

AlexanderSehr: > Thanks for the update! There's no hurry. Just wanted to make a contribution. I have disabled my Azure subscription now, though. Because somehow this little thing here led to a 300 EUR Azure bill, which I'm not too happy with.

I'm sorry to hear. Did the workflows run on their schedule? If so, please note that there is a workflow [.Platform - Toggle AVM workflows](https://github.com/Azure/bicep-registry-modules/actions/workflows/platform.toggle-avm-workflows.yml) that you should run at least once to disable all workflows to prevent this from happening in forks. It's called out in the [contribution guide](https://azure.github.io/Azure-Verified-Modules/contributing/bicep/bicep-contribution-flow/#33-set-readwrite-workflow-permissions) and we tried to drag attention to it. If this happend to be the case for you, would you have an idea how to call this out in a better way? Something that would have immediately caught your attention? We certainly don't want contributors to burn their budget. Please let me/us know.

To the topic at hand I did run a small test that 'seemed to have worked' - but one step at a time. When you wrote 
> _I've tried setting the property enableHierarchicalNamespace for the storage-account AVM to null, false, true, and I've left it out entirely as well. But it always results into the same error._
did you mean you try to deploy the module as is with the 3 different values? If so, I think what happened is the following
- When providing `true`, to the module, it set `true`
- When provideing `false` to the module, it set the value to `null`
- When providing `null` to the module, it also passed `null` as it would fall back to the default

That being said, here is what I tested:
1. I did remove the `isHnsEnabled` from the module and deployed a new storage account (which should mean that it's not set, like in your TF example)
2. I did add the property again, but made the parameter that sets it `nullable` like this:

```
@description('Conditional. If true, enables Hierarchical Namespace for the storage account. Required if enableSftp or enableNfsV3 is set to true.')
param enableHierarchicalNamespace bool?

resource storageAccount 'Microsoft.Storage/storageAccounts@2022-09-01' = {
  name: name
  (...)
  properties: {
   (...)
   isHnsEnabled: enableHierarchicalNamespace
  }
}
```
3. I did deploy the template again without setting the value and it did work without a problem.

I understand, point 3 would have failed before, right? First deploying the module without `isHnsEnabled` configured, and then all of a sudden with and it started failing.

That said - I must have gotten something wrong, right? I mean it does work (seemingly) - but comparing it with the current implementation in main and the 3 values you provided for testing using that latest version, the value that is passed to the resource provider seems to be the same. E.g., if you provided `false` or `null` it also passed `null` to the resource provider in the original implementation. So why does my test not fail? 😄

bartcode: As for the charges, I _thought_ I had specifically _enabled_ the storage account AVM tests, so didn't see a specific need to disable workflows. You're entirely right it's stated in the documentation. It's now listed as a "tip", but I'd argue it's not so much a "tip" as it is an important warning. I've sent a support request to the billing team, by the way. And they're saying they might refund a part of it, which could be around 100 EUR. That's better than nothing, but it's still not the kind of money I was hoping to lose here.

> did you mean you try to deploy the module as is with the 3 different values? If so, I think what happened is the following

Yes, that's exactly what I did. And as a workaround for now (the business needs to continue), I ended up using the `storageAccount` resource directly, and leaving out the `isHnsEnabled` property.

Your example should, indeed, result into the same error. However, there's a tiny difference here, because we initially provisioned the storage account through Terraform, which doesn't use a "Deployment" the way ARM/Bicep does. Could it be that the storage account is created slightly differently there? I'd say they _should_ work the same in the end, but that's really the only difference I see here.

(I'm actually off for two weeks now, but I can try and verify this when I'm back.)

AlexanderSehr: > As for the charges, I thought I had specifically enabled the storage account AVM tests, so didn't see a specific need to disable workflows. But, as you assumed correctly, the scheduled workflows have been running. You're entirely right it's stated in the documentation. It's now listed as a "tip", but I'd argue it's not so much a "tip" as it is an important warning. I've sent a support request to the billing team, by the way. And they're saying they might refund a part of it, which could be around 100 EUR. That's better than nothing, but it's still not the kind of money I was hoping to lose here.

I'll bring this up with the team to see what other options we may have to prevent this from happening in the future. Current actions seem not sufficient.

> Yes, that's exactly what I did. And as a workaround for now (the business needs to continue), I ended up using the storageAccount resource directly, and leaving out the isHnsEnabled property.
>
> Your example should, indeed, result into the same error. However, there's a tiny difference here, because we initially provisioned the storage account through Terraform, which doesn't use a "Deployment" the way ARM/Bicep does. Could it be that the storage account is created slightly differently there? I'd say they should work the same in the end, but that's really the only difference I see here.

I don't think the deployment object would make a difference as it is just metadata. However, that doesn't mean that Terraform may not run some diverging magic regardless when deploying. It would be interesting to see how the storage account object looks like when one clicks on 'Export Template' in the Portal after deploying the Storage Account after deploying it via Terraform - and to compare the same with one that is deployed via the module / ARM / Bicep. 

I only have the latter example in front of me and I can at least say that the `isHnsEnabled` value is not set, even after running both deployments as per my previous tests.

I'll try to run the end to end test with a leading Terraform deployment (as per your scenario) once I find a moment. In the meatime, enjoy your time off 💪

bartcode: @AlexanderSehr , I genuinely don't understand, but somehow I can't reproduce it anymore either and it seems to magically work now (sigh). I tried to 'redeploy' a storage account that was first deployed using Terraform and it worked on the first try this time.

Thanks for your efforts, though. And I'll hopefully bring an actual contribution next time.

I'll close the PR.

AlexanderSehr: Oh wow, thanks @bartcode. Believe it or not - but it's still very much on my TODO list, just that the end of MS's fiscal is unfortuantely a bit crazy to which end I didn't get to it yet.
I guess I'm happy to hear that it 'seems' to work now, but I anyways want to apologize that I wasn't able to get around to it sooner.

bartcode: @AlexanderSehr , you might want to read [this](https://github.com/Azure/bicep-registry-modules/issues/1522#issuecomment-2180327310). I wasn't the only one apparently, and something seems to have been fixed in the meantime. The issue is still there for older storage accounts, which obviously makes it a bit hard to reproduce it.

PR #2263, opened 2024-06-07
feat: Added GH CLI support for 'Trigger Workflows for Branch' script

AlexanderSehr:
## Description

- Added GH CLI support for 'Trigger Workflows for Branch' script
- Done in preparation for a workflow that could centrally trigger all module pipelines on a schedule (and use the GH CLI)
  - Using this script will be very simple for this use case, effectively only requiring the schedule and invoking the script with it's default parameters. 
- Also added a feature that by default excludes disabled workflows from the trigger (as those would fail the trigger event)
- Validated locally..This is no platform script (yet)
- Small fixes like the regex filter

## Type of Change

  - [ ] Update to documentation


avm/utilities/tools/Invoke-WorkflowsForBranch.ps1
eriqua, line 1: If not urgent I'd be keen to wait for the platform workflow to understand the full picture

AlexanderSehr: Hey @eriqua,
the 2 topics are split on purpose as they're not strictly connected (this PR simply enables the CH CLI for a tool that we're already using).
The later workflow PR will sabotage my fork as I need to merge the workflow into main to test it. Want to delay this as long as possible and then merge it via a new PR as fast as possible. As the workflow will be rather simple, I suspect that will be easy - but this one here needs to go in first. 

Regardless, if you have a question about what the idea behind the later workflow would be, please reach out.

You can however imagine something like this which I already drafted: https://github.com/AlexanderSehr/bicep-registry-modules/blob/users/alsehr/scheduledWorkflowPart2/.github/workflows/platform.trigger-module-workflows.yml

avm/utilities/tools/Invoke-WorkflowsForBranch.ps1
```diff
@@ -118,26 +147,46 @@ function Get-GitHubModuleWorkflowList {
         [string] $RepositoryName,
 
         [Parameter(Mandatory = $false)]
-        [string] $Filter = 'avm.res.*'
+        [switch] $IncludeDisabled,
+
+        [Parameter(Mandatory = $false)]
+        [string] $Filter = 'avm\.(?:res|ptn)'
```
ChrisSidebotham, line 153: Happy with this PR, Should we expand this to cover `utl` though?

AlexanderSehr: Good point. I guess we can start added this. I guess we'll need to do run a general update sometime soon anyways, but doesn't hurt to start. Will update

AlexanderSehr: done @ChrisSidebotham

ChrisSidebotham (dismissed):

AlexanderSehr:
Anybody?

jtracey93:
@ChrisSidebotham approved this originally but seems your push of a commit wiped this?

avm/utilities/tools/Invoke-WorkflowsForBranch.ps1
```diff
@@ -88,16 +114,19 @@ Get a list of all GitHub module workflows
...
 
 .PARAMETER RepositoryName
 Mandatory. The name of the repository to fetch the workflows from.
 
+.PARAMETER IncludeDisabled
+Optional. Set if you want to also include disabled workflows in the result.
+
 .PARAMETER Filter
-Optional. A filter to apply when fetching the workflows. By default we fetch all module workflows (avm.res.*).
+Optional. A regex filter to apply when fetching the workflows. By default we fetch all module workflows (avm.res.*).
```
eriqua, line 129: Either the below or we should update the removed expression to also include ptn and utl
```suggestion
Optional. A regex filter to apply when fetching the workflows. By default we fetch all module workflows.
```

avm/utilities/tools/Invoke-WorkflowsForBranch.ps1
eriqua: Sounds fair @AlexanderSehr , this is useful on its own, even in case we are not going to add scheduled pipelines back. Resolving

eriqua (approved):
:ok:

PR #2345, opened 2024-06-13
fix: Added explicit databricks post-removal

AlexanderSehr:
## Description

- Even though the documentation claims otherwise, managed RGs are not consistently removed by Databricks
- Fixed the rg naming to align across tests & module (and WAF for that matter)
- This PR adds custom logic to address this in case the native removal did not work. 
  > Note: It's functionality depends on the explicit prefix & suffic set by the module for the managed-rgs. If they were to change, the test would need to change

image

## Pipeline Reference

| Pipeline |
| -------- |
|    [avm.res.databricks.workspace](https://github.com/Azure/bicep-registry-modules/actions/workflows/avm.res.databricks.workspace.yml)      |

## Type of Change

  - [ ] Update to documentation


avm/utilities/pipelines/e2eValidation/resourceRemoval/helper/Invoke-ResourcePostRemoval.ps1
```diff
@@ -152,6 +152,29 @@ function Invoke-ResourcePostRemoval {
...
+            # NOTE: This requires that the provided value uses the suffix '-managed'
+            $managedResourceGroupName = "rg-$resourceGroupName-managed"
+            if (Get-AzResourceGroup -Name $managedResourceGroupName -ErrorAction 'SilentlyContinue') {
+                Write-Verbose ('[*] Removing managed resource group [{0}] of workspace [{1}]' -f $managedResourceGroupName, $resourceName) -Verbose
+                if ($PSCmdlet.ShouldProcess(('Managed resource group [{0}]' -f $managedResourceGroupName), 'Remove')) {
+                    $null = Remove-AzResourceGroup -Name $managedResourceGroupName -Force
+                }
+            }
+            # If the `managedResourceGroupResourceId` parameter was NOT set during deployment, we should look for a resource group with the default name and remove it as it is not automatically removed
+            # NOTE: This requires that the default value uses the suffix '-rg'
```
eriqua, line 168: I see in the bicep template that was updated to be a prefix `rg-` instead. 

Anyway, why not using `-managed` for the default as well at this point? Or should we use what the portal creates by default?

AlexanderSehr: Good catch.

Regarding the `managed`, that's up to @clintgrove to decide.
The `-rg` stuff I just did for alignement and because it's WAF-aligned.

AlexanderSehr: Alternativly @eriqua, I guess I could just have the logic try and find all resource groups who's name contains either the workspace name (default test case) or the workspace's resource group name (all others). What do you think?

clintgrove: I do believe that the managed resource group does contain the databricks workspace name in it, so searching for RGs that contain the name would be an approach

clintgrove: Its not a big deal, but the managed resource group, will it start with the word rg- and then have the databricks workspace name after it? I think naming it "managed" might be better then you can search for that to make it even safer to find and delete. So you will be looking for managed-<databricks workspace name>

AlexanderSehr: Hey @clintgrove,
thanks for the response. To ensure we're not talking past each other (and this may actually come down to what was migrated from CARML):

## Today the implementation forthe managed-rg is:
main.bicep default (used by default test)
```bicep
managedResourceGroupId: !empty(managedResourceGroupResourceId)
      ? managedResourceGroupResourceId
      : '${subscription().id}/resourceGroups/${name}-rg'
```

main.test test parameter implementation (used by waf & max)
```bicep
managedResourceGroupResourceId: '${subscription().id}/resourceGroups/rg-${resourceGroupName}-managed'
```

## My original proposal
```bicep
managedResourceGroupId: !empty(managedResourceGroupResourceId)
      ? managedResourceGroupResourceId
      : '${subscription().id}/resourceGroups/rg-${name}' # <== This is what I changed
```

main.test test parameter implementation (used by waf & max)
```bicep
managedResourceGroupResourceId: '${subscription().id}/resourceGroups/rg-${resourceGroupName}-managed'
```
---

As you can see in the `main.test.bicep` it's using the resourceGroupName, not the resourceName for the managed rg. Naturally, we can change this to whatever we want. It's just a test value.

How about this suggestion then
## Today the implementation forthe managed-rg is:
main.bicep default (used by default test)
```bicep
managedResourceGroupId: !empty(managedResourceGroupResourceId)
      ? managedResourceGroupResourceId
      : '${subscription().id}/resourceGroups/rg-${name}-managed'
```

main.test test parameter implementation (used by waf & max)
```bicep
managedResourceGroupResourceId: '${subscription().id}/resourceGroups/rg-${name}-managed'
```
With the one caveat that for `name` in the `main.test.bicep` I'd need to have a variable with the name. Currently the value is directly assigned to the test's `name` parameter. But that's a small change.

If we were to use that implemetnation, the removal logic could look for any resource group that's like `"rg-$resourceName-managed"` in the removal logic. And it would work consistently unless somebody provides some other crazy name in the test case in the future :)

eriqua: Stupid question: how much effort would it be to use the deployment output? The managed rg is an output parameter. At that point we would not need to enforce any name

clintgrove:
I am good with your suggestion Alex. Go ahead

avm/utilities/pipelines/e2eValidation/resourceRemoval/helper/Invoke-ResourcePostRemoval.ps1
AlexanderSehr: That's an interesting question @eriqua 

The outputs are written already into the pipeline (as needed by the Post-Deployment-Tests logic we introduced a while back). The way it works is that any output that is returned by a test-file is written into a `DeploymentOutputs` hashtable that is passed into each test. Whether needed / provided or not.

So we could do the same with the removal logic and pass a `DeploymentOutputs` or similar named hashtable through the following scripts
- `Initialize-DeploymentRemoval`
- `Remove-Deployment`
- `Remove-ResourceList`
- `Remove-ResourceListInner`
- `Invoke-ResourceRemoval` & `Invoke-ResourcePostRemoval`

To have it available in the removal scripts too - most notably the last two at the end of the call stack.

Complexity-wise this isn't much to be concerned about - but - it's a change to the CI stack, which would also render this PR a bit bigger in impact. If we agree, I can certainly implement it though.

To make full use of the logic though, the following changes would need to happen to the databricks tests
- The managedRG value must be added as an output to the databricks `main.bicep` module
- The output must be passed as a test file output
done

eriqua: Let's discuss it offline, but I'd suggest to take this as a next step, if beneficial, and have this mitigated meanwhile since already implemented. Plus @clintgrove already gave his "good to go" 🙂 
Also considering I'm not sure the approach would work in case of deployment failure.

The only step before approval from my side is the previous comment regarding rg custom and default names, if we want to leverage them. I think the default rg should be the same created by default when deploying the resource from the portal. Based on that we should implement the custom logic and not adapt the default name to make it easier for the ci to remove it (or at least that should be the last option)

AlexanderSehr: @AlexanderSehr - TODO: Deploy a databricks workspace and check what the managed RG is named like

eriqua (approved):
🆗

PR #2462, opened 2024-06-16
fix: comply with WAF test for maintenance configurations

rahalan:
Added the possibility to deploy configuration assignments to a VM and updated the tests (default and WAF aligned) to utilize that

"Successful" run: [avm.res.compute.virtual-machine](https://github.com/rahalan/bicep-registry-modules/actions/workflows/avm.res.compute.virtual-machine.yml)

reason for being red: I am not allowed to deploy nVidia VMs in my tenant due to policy restrictions


avm/res/compute/virtual-machine/main.bicep
```diff
@@ -131,6 +131,9 @@ param backupPolicyName string = 'DefaultPolicy'
 @description('Optional. The configuration for auto-shutdown.')
 param autoShutdownConfig object = {}
 
+@description('Optional. The resource Id of a maintenance configuration for this VM.')
+param maintenanceConfigurationId string = ''
```
AlexanderSehr, line 135: As per some spec that definitely exist, if this is a resource ID it should be in the name.

AlexanderSehr (changes requested):

avm/res/compute/virtual-machine/tests/e2e/linux.defaults/dependencies.bicep
```diff
@@ -35,6 +38,40 @@ resource virtualNetwork 'Microsoft.Network/virtualNetworks@2023-04-01' = {
...
+        ]
+      }
+      linuxParameters: {
+        classificationsToInclude: [
+          'Critical'
+          'Security'
+        ]
+      }
+    }
+    // configurationType: 'Regular'
```
AlexanderSehr, line 71: Leftover?

AlexanderSehr (changes requested):

avm/res/compute/virtual-machine/tests/e2e/linux.defaults/main.test.bicep
```diff
@@ -37,6 +37,7 @@ module nestedDependencies 'dependencies.bicep' = {
   params: {
     location: resourceLocation
     virtualNetworkName: 'dep-${namePrefix}-vnet-${serviceShort}'
+    maintenanceConfigurationName: 'dep-${namePrefix}-mc-${serviceShort}'
```
AlexanderSehr, line 40: Should only be set in the WAF tests as this is an optional value. For the default tests, if a PSRule test fails for a parameter that requires a user input (as opposed to e.g., a boolean) then we should exclude it via the PS Rule min configuration file (with a comment to explain why)

AlexanderSehr (changes requested):

avm/res/compute/virtual-machine/tests/e2e/linux.defaults/main.test.bicep
AlexanderSehr: The same is true for all other default parameters

avm/res/compute/virtual-machine/tests/e2e/windows.defaults/main.test.bicep
```diff
@@ -41,6 +41,7 @@ module nestedDependencies 'dependencies.bicep' = {
   params: {
     location: resourceLocation
     virtualNetworkName: 'dep-${namePrefix}-vnet-${serviceShort}'
+    maintenanceConfigurationName: 'dep-${namePrefix}-mc-${serviceShort}'
```
AlexanderSehr, line 44: Same comment as for Linux default test

AlexanderSehr (changes requested):

avm/res/compute/virtual-machine/main.bicep
rahalan: fixed

avm/res/compute/virtual-machine/tests/e2e/linux.defaults/dependencies.bicep
rahalan: removed

avm/res/compute/virtual-machine/tests/e2e/linux.defaults/main.test.bicep
rahalan: but the name is not optional, but required

avm/res/compute/virtual-machine/tests/e2e/windows.defaults/main.test.bicep
rahalan: name is not optional

avm/res/compute/virtual-machine/tests/e2e/linux.defaults/main.test.bicep
AlexanderSehr: Hey @rahalan,
I think we might be talking past each other. The `maintenanceConfigurationName` here is mandatory yes - but only because you deploy a maintenanceConfiguration in the dependencies. However, I question if that is necessary as you only deploy it becausse you set 3 new parameters in the defaults test
```
      patchMode: 'AutomaticByPlatform'
      bypassPlatformSafetyChecksOnUserSchedule: true
      maintenanceConfigurationResourceId: nestedDependencies.outputs.maintenanceConfigurationResourceId
```
which I'd argue should probably not be set in a test that wants to test if the module works if you only set required parameters 😉

rahalan: @AlexanderSehr without those parameters you cannot set a maintenanceConfiguration. And I deploy the config, because I need to set in according to WAF.

rahalan: done

avm/res/compute/virtual-machine/tests/e2e/windows.defaults/main.test.bicep
rahalan: done

avm/utilities/pipelines/staticValidation/psrule/.ps-rule/min-suppress.Rule.yaml
```diff
@@ -31,6 +31,7 @@ spec:
     - Azure.ContainerApp.PublicAccess
     # Azure Virtual Machine
     - Azure.VM.AMA
+    - Azure.VM.MaintenanceConfig
```
AlexanderSehr, line 34:
```suggestion
    - Azure.VM.MaintenanceConfig # Excluded as it requires user input
```

AlexanderSehr (approved):

PR #2467, opened 2024-06-18
fix: `avm/res/network/network-security-group` retired security rules child module and introduced UDT as a replacement

### LINKED ISSUE #2185: [AVM Question/Feedback]: Duplicate security rules in avm.res.network.network-security-group

### Description

Hello, I have started to use AVM recently and experienced server timeouts when attempting to deploy NSG rules using the NSG module. According to Microsoft support, this was caused by the ARM and Network Resource provider resources being out of sync, I'm still having these issues even after resources have been synced, so the root cause is still unknown and investigation still in progress.

On closer inspection of the network security group, I could see that security rules are being created twice. Once on [L:74 ](https://github.com/Azure/bicep-registry-modules/blob/c3f9b321118d0809ab5fb66f5552e31cd54d3c01/avm/res/network/network-security-group/main.bicep#L74C5-L74C18) of the network-security-group module and a second time on [L:122](https://github.com/Azure/bicep-registry-modules/blob/c3f9b321118d0809ab5fb66f5552e31cd54d3c01/avm/res/network/network-security-group/main.bicep#L122) using a security-rule module.

I'm keen to understand why it was implemented this way as it seems redundant and I'm wondering if that's causing the resources to be out-of-sync. Cheers

ahmadabdalla:
## Description

Closes #2185 

This pull request removes the `securityRules` child module. See issue for more details. In addition to improve the customer experience with the module:

- Introduced User Defined Types (UDT) on the `securityRules` property. Allowing customers to have IntelliSense on the properties.
- Simplifying the referencing to the rules containing application security groups, instead of passing an `id` property per App Security Group, and replaced it with Resource IDs, while performing the mapping within the module.
- Added additional test cases in the `max` case to ensure we have good coverage for the security rules parameter.
- Updated module version.

## Pipeline Reference

| Pipeline |
| -------- |
|   [avm.res.network.network-security-group](https://github.com/ahmadabdalla/bicep-registry-modules/actions/workflows/avm.res.network.network-security-group.yml)       |

## Type of Change

  - [ ] Update to documentation

## Checklist


AlexanderSehr (approved):
Thanks a lot for the improvements you so casually added while removing the child module 😄 Approved

PR #2502, opened 2024-06-20
fix: `res/compute/gallery` fixes

### LINKED ISSUE #2359: [Bug Report]: Image creation purchase plan information error
### Image creation purchase plan information error

When using Microsoft.compute/galleries/images, the part of the code that inputs Plan information is incorrect.  As this causes an error in the portal when using a Microsoft image.  If you try and deploy a Vm from the compute gallery you get a validation error that there is no plan name.

When deploying it has the below in the arm template that you can download before creating the resource.

```
purchasePlan: {}
```
But as it's a Microsoft image there is no plan information.  If you delete that part and then rerun the script it validates and will then let you proceed.

This is the code that is causing the issue.
```
    purchasePlan: {
      product: !empty(productName) ? productName : null
      name: !empty(planName) ? planName : null
      publisher: !empty(planPublisherName) ? planPublisherName : null
    }
```
I think it needs to be something like this or similar so that it does not pass any plan information

```
purchasePlan: !empty(purchasePlan) ? purchasePlan : null
```

### To reproduce

This is run from a bicep solution that uses azure image builder to build a custom image that then puts that image into the compute gallery.

### Code snippet

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #2403: [Bug Report]: Missing properties for Compute/Gallery for Bicep when creating images
### Describe the bug

When trying to create a Compute/Gallery with images in array, the properties "isAcceleratedNetworkSupported" and "isHibernateSupported" aren't passed.

They exist as parameters for Compute/Gallery/Image but when comes the moment to create images from Compute/Gallery the properties are just not passed.

It is a fairly simple bug to solve, as a lot of the stuff surrounding the solution is already in place.

### To reproduce

Deploy a Compute/Gallery via Bicep with an image array specifying isAccelereatedNetworkSupported to true.

You will that the property is not applied!

I believe the explanation of this bug is quite simple. The Compute/Gallery main.json doesn't pass the property.

Screenshot 2024-04-26 at 9 25 49 AM

galleries_images is missing the params for "isAcceleratedNetworkSupported" and "isHibernateSupported".

### Code snippet

```Bicep
param galImages = [
  {
    hyperVGeneration: 'V2'
    maxRecommendedMemory: 16
    maxRecommendedvCPUs: 8
    minRecommendedMemory: 4
    minRecommendedvCPUs: 2
    isAcceleratedNetworkSupported: true
    name: 'xrserver-devkit'
    offer: 'xrserver-devkit'
    osState: 'Generalized'
    securityType: 'TrustedLaunch'
    osType: 'Linux'
    publisher: 'example'
    sku: 'xrserver_devkit'
  }
  {
    hyperVGeneration: 'V2'
    maxRecommendedMemory: 16
    maxRecommendedvCPUs: 8
    minRecommendedMemory: 4
    minRecommendedvCPUs: 2
    isAcceleratedNetworkSupported: true
    name: 'xrserver-scaleset'
    offer: 'xrserver-scaleset'
    securityType: 'TrustedLaunch'
    osState: 'Generalized'
    osType: 'Linux'
    publisher: 'example'
    sku: 'xrserver_scaleset'
  }
]

module gallery '../../modules/compute/gallery/main.bicep' = {
  name: '${uniqueString(deployment().name, location)}-dev-cgcom'
  params: {
    name: galName
    location: location
    enableDefaultTelemetry: true
    images: galImages
    roleAssignments: [
      {
        principalIds: [
          galUserAssignedIdentity.outputs.principalId
        ]
        principalType: 'ServicePrincipal'
        roleDefinitionIdOrName: 'Reader'
      }
    ]
    lock: ''
  }
}
```

### Relevant log output

_No response_

ReneHezser:
## Description

- implement custom type instead of separate properties for images
- bugfixes

Fixes #2403 
Fixes #2359 

## Pipeline Reference

| Pipeline |
| -------- |
|   [avm.res.compute.gallery](https://github.com/ReneHezser/bicep-registry-modules/actions/workflows/avm.res.compute.gallery.yml)       |

## Type of Change

  - [x] Update to documentation

## Checklist


matebarabas:
> Important
> 
> If this is a module-related PR, being submitted by the sole owner of the module, the AVM core team must review and approve it (as module owners can't approve their own PRs).
> 
> **To indicate this PR needs the core team''s attention, apply the "Needs: Core Team 🧞" label!**
> 
> The core team will only review and approve PRs that have this label applied!

@ReneHezser, please note this new requirement! Thanks!

avm/res/compute/gallery/image/main.bicep
```diff
@@ -12,108 +12,13 @@ param location string = resourceGroup().location
...
-@sys.description('Optional. List of the excluded disk types (e.g., Standard_LRS).')
-param excludedDiskTypes array = []
+@description('Required. Properties used to create the image.')
+param galleryImage imageType
 
 @sys.description('Optional. Array of role assignments to create.')
 param roleAssignments roleAssignmentType
 
-@sys.description('Optional. Tags for all resources.')
+@sys.description('Optional. Tags for all resources. This takes precedence over the tags of the galleryImage.tags property.')
```
AlexanderSehr, line 21: Please note that given that each module is treated like an indipendent unit, the added sentense does not make much sense. It would only if implemented in a UDT of the parent

avm/res/compute/gallery/image/main.bicep
```diff
@@ -134,67 +39,61 @@ var builtInRoleNames = {
...
   name: galleryName
 }
 
-resource image 'Microsoft.Compute/galleries/images@2022-03-03' = {
+resource image 'Microsoft.Compute/galleries/images@2023-07-03' = {
   name: name
   parent: gallery
   location: location
-  tags: tags
+  tags: tags // use from the parameter, as the galleryImage.tags have been stored there in the gallery.bicep before calling the template
```
AlexanderSehr, line 50: Not sure I understand what the comment is telling me. But I do have the same feedback as for the update of it's parameter description 😄

avm/res/compute/gallery/image/main.bicep
```diff
@@ -12,108 +12,13 @@ param location string = resourceGroup().location
...
-@sys.description('Optional. The publisher ID.')
-param planPublisherName string?
-
-@sys.description('Optional. The end of life date of the gallery Image Definition. This property can be used for decommissioning purposes. This property is updatable. Allowed format: 2020-01-10T23:00:00.000Z.')
-param endOfLife string = ''
-
-@sys.description('Optional. List of the excluded disk types (e.g., Standard_LRS).')
-param excludedDiskTypes array = []
+@description('Required. Properties used to create the image.')
+param galleryImage imageType
```
AlexanderSehr, line 16: Please note that this is not a pattern we usually apply and I would recommend not using it. Moving all properties into a UDT means that we take any option for smart defaults etc. away / hide them in the template. Top-level properties of any module should always be their own parameters and only of those parameters in turn are of type object / array they should implement a UDT.

avm/res/compute/gallery/image/main.bicep
```diff
@@ -256,3 +155,120 @@ type roleAssignmentType = {
...
+      publisher: string
+
+      @sys.description('Required. The name of the gallery image definition publisher.')
+      offer: string
+
+      @sys.description('Required. The name of the gallery image definition SKU.')
+      sku: string
+    }
+
+    @sys.description('Required. This property allows the user to specify whether the virtual machines created under this image are `Generalized` or `Specialized`.')
```
AlexanderSehr, line 221: Defaults in this and other parameters should not be part of the description as they're anyways part of the metadata (via the allowedSet)

avm/res/compute/gallery/image/main.bicep
```diff
@@ -256,3 +155,120 @@ type roleAssignmentType = {
...
+      @sys.description('Required. The publisher ID.')
+      publisher: string
+    }?
+
+    @sys.description('Optional. The properties describe the recommended machine configuration for this Image Definition. These properties are updatable.')
+    recommended: {
+      @sys.description('Optional. Describes the resource range (1-128 CPU cores). Defaults to min=1, max=4.')
+      vCPUs: resourceRangeType?
+
+      @sys.description('Optional. Describes the resource range (1-4000 GB RAM). Defaults to min=4, max=16.')
```
AlexanderSehr, line 247: Where are these defaults coming from?

avm/res/compute/gallery/image/main.bicep
```diff
@@ -256,3 +155,120 @@ type roleAssignmentType = {
...
+      diskTypes: string[]
+    }?
+
+    @sys.description('Optional. The end of life date of the gallery image definition. This property can be used for decommissioning purposes. This property is updatable.')
+    endOfLifeDate: string?
+
+    @sys.description('Optional. The Eula agreement for the gallery image definition.')
+    eula: string?
+
+    // features are skipped for now, as we prefer typed parameters
```
AlexanderSehr, line 204: leftover?

avm/res/compute/gallery/image/main.bicep
```diff
@@ -256,3 +155,120 @@ type roleAssignmentType = {
...
+    @sys.description('Optional. The hypervisor generation of the Virtual Machine. If this value is not specified, then it is determined by the securityType parameter. If the securityType parameter is specified, then the value of hyperVGeneration will be V2, else V1.')
+    hyperVGeneration: 'V1' | 'V2'?
+
+    @sys.description('Required. This is the gallery image definition identifier.')
+    identifier: {
+      @sys.description('Required. The name of the gallery image definition offer.')
+      publisher: string
+
+      @sys.description('Required. The name of the gallery image definition publisher.')
+      offer: string
```
AlexanderSehr, line 215: The metadata is flipped

avm/res/compute/gallery/image/main.bicep
```diff
@@ -256,3 +155,120 @@ type roleAssignmentType = {
   @sys.description('Optional. The Resource Id of the delegated managed identity resource.')
   delegatedManagedIdentityResourceId: string?
 }[]?
+
+@export()
```
AlexanderSehr, line 159: While I applaud the idea in order to save code, I strongly advice against wrapping the entire properties block. 
Instead, I'd advice to create one UDT per non-top level property that is either an array or object (like the `identifier` property), mark them all via `export()` and use them in a UDT in the parent.

avm/res/compute/gallery/main.bicep
```diff
@@ -9,27 +9,40 @@ param name string
 @sys.description('Optional. Location for all resources.')
 param location string = resourceGroup().location
 
+@sys.description('Optional. Enable/Disable usage telemetry for module.')
+param enableTelemetry bool = true
+
+// ============ //
+// Parameters   //
+// ============ //
+
```
AlexanderSehr, line 18: What's with the parameters above? 😄

ReneHezser: Again copied from "some template" and thought this had a purpose.

avm/res/compute/gallery/image/main.bicep
ReneHezser: I've reverted parts of the type, as discussed earlier.

avm/res/compute/gallery/image/main.bicep
ReneHezser: They are assigned when calling the image template from the main loop.

avm/res/compute/gallery/image/main.bicep
ReneHezser: I've changed the parameters after our conversation.

avm/res/compute/gallery/main.bicep
ReneHezser: They are... also... never mind.

avm/res/compute/gallery/main.bicep
```diff
@@ -140,36 +156,36 @@ module galleries_images 'image/main.bicep' = [
...
-      planName: image.?planName
-      planPublisherName: image.?planPublisherName
-      endOfLife: image.?endOfLife
-      excludedDiskTypes: image.?excludedDiskTypes
+      purchasePlan: image.?purchasePlan
+      endOfLifeDate: image.?endOfLife
+      disallowed: { diskTypes: image.?excludedDiskTypes ?? [] }
       roleAssignments: image.?roleAssignments
-      tags: image.?tags ?? tags
+      tags: tags
```
AlexanderSehr, line 180:
```suggestion
      tags: image.?tags ?? tags
```

avm/res/compute/gallery/main.bicep
```diff
@@ -220,4 +236,88 @@ type roleAssignmentType = {
...
+  description: string?
+
+  @sys.description('Required. This property allows you to specify the type of the OS that is included in the disk when creating a VM from a managed image.')
+  osType: ('Linux' | 'Windows')
+
+  @sys.description('Required. This property allows the user to specify the state of the OS of the image.')
+  osState: ('Generalized' | 'Specialized')
+
+  @sys.description('Required. This is the gallery image definition identifier.')
+  identifier: {
```
AlexanderSehr, line 257: If you 'export()` the types in the child, you should be able to import them here just as you wanted to originally

avm/res/compute/gallery/image/main.bicep
AlexanderSehr: That's for a child-module sadly not a good option as you must imagine that this module may at some point be deployed on it's own where the defaults won't apply. But - now that you split the UDT I can add some alternatives :)

avm/res/compute/gallery/image/main.bicep
```diff
@@ -12,108 +14,69 @@ param location string = resourceGroup().location
...
 
-@sys.description('Optional. The description of this gallery Image Definition resource. This property is updatable.')
-param description string?
+@sys.description('Optional. Describes the gallery image definition purchase plan. This is used by marketplace images.')
+param purchasePlan purchasePlanType?
 
-@sys.description('Optional. The Eula agreement for the gallery Image Definition. Has to be a valid URL.')
-param eula string?
+@sys.description('Optional. Describes the resource range (1-128 CPU cores). Defaults to min=1, max=4.')
+param vCPUs resourceRangeType?
```
AlexanderSehr, line 33: Now that you split the UDT into actual parameters we can make full use if their feature set. 2 changes, the first here:
```suggestion
param vCPUs resourceRangeType = { 
  min: 1 
  max: 4 
}
```
and the other in the parent `main.bicep`

avm/res/compute/gallery/main.bicep
```diff
@@ -140,36 +156,36 @@ module galleries_images 'image/main.bicep' = [
...
-      offer: image.offer
-      sku: image.sku
-      minRecommendedvCPUs: image.?minRecommendedvCPUs
-      maxRecommendedvCPUs: image.?maxRecommendedvCPUs
-      minRecommendedMemory: image.?minRecommendedMemory
-      maxRecommendedMemory: image.?maxRecommendedMemory
+      osState: image.osState
+      identifier: image.identifier
+      vCPUs: image.?vCPUs ?? { min: 1, max: 4 }
+      memory: image.?memory ?? { min: 4, max: 16 }
```
AlexanderSehr, line 167:
```suggestion
      vCPUs: image.?vCPUs
      memory: image.?memory
```

With the default values fully implemented in the child-modules, we don't need them here anymore as they're automatically pulled if `null` is passed :)

avm/res/compute/gallery/main.bicep
```diff
@@ -220,4 +236,63 @@ type roleAssignmentType = {
...
+  @sys.description('Required. This property allows you to specify the type of the OS that is included in the disk when creating a VM from a managed image.')
+  osType: ('Linux' | 'Windows')
+
+  @sys.description('Required. This property allows the user to specify the state of the OS of the image.')
+  osState: ('Generalized' | 'Specialized')
+
+  @sys.description('Required. This is the gallery image definition identifier.')
+  identifier: identifierType
+
+  @sys.description('Optional. Describes the resource range (1-128 CPU cores). Defaults to min=1, max=4.')
```
AlexanderSehr, line 260: Question: You have the range so neatly written here. I wonder - should we add the min & max directly as annotations to the property in the type instead? Then we get direct language support.
Same question for the memory parameter

avm/res/compute/gallery/image/main.bicep
```diff
@@ -12,108 +14,69 @@ param location string = resourceGroup().location
...
 @sys.description('Optional. The release note uri. Has to be a valid URL.')
 param releaseNoteUri string?
 
-@sys.description('Optional. The product ID.')
-param productName string?
+@sys.description('Optional. The security type of the image. Requires a hyperVGeneration V2.')
+param securityType ('Standard' | 'TrustedLaunch' | 'ConfidentialVM' | 'ConfidentialVMSupported')?
+
+@sys.description('Optional. Specify if the image supports accelerated networking. Defaults to true.')
+param isAcceleratedNetworkSupported bool?
```
AlexanderSehr, line 45:
```suggestion
@sys.description('Optional. Specify if the image supports accelerated networking.')
param isAcceleratedNetworkSupported bool = true
```


I'd suggest to check this for other parameters too. With it now being a parameter, we can have a default and hence can set the default directly 💪

avm/res/compute/gallery/main.bicep
AlexanderSehr: Ah, now I see 2 different properties use the same UDT. I'd almost suggest to split the to enable the rules as subjectively it's always better to fail before trying to even deploy, but you call 😄 Feel free to resolve

avm/res/compute/gallery/image/main.bicep
AlexanderSehr: > in the main I am using the UDT, where I can't set the default. The default will be set on the param in the image.bicep

As intended. It's pulled in 😉 I just split the statement in two

AlexanderSehr (approved):

PR #2565, opened 2024-06-28
fix: APIM SKUs not deploying due to incorrect param logic

### LINKED ISSUE #2561: [AVM Module Issue]: Setting up 'AvailabilityZones' is not supported in Sku :'Developer'

### Issue Type?

Feature Request

### Module Name

avm/res/api-management/service

### (Optional) Module Version

0.2.0

### Description

Hi,

I want to deploy an APIM instance in the SKU "Developer".  The Developer SKU defaults to 1 instance count.

When the following bicep code is used:

```
module apim 'br/public:avm/res/api-management/service:0.2.0' = {
  name: 'apim9999'
  params: {
    name: 'testapim9999'
    publisherEmail: publisherEmail
    publisherName: publisherName
    virtualNetworkType: 'Internal'
    subnetResourceId: subnetRef.id
    location: location
    sku: 'Developer'
    ....
  }
}
```

i get the error:

> [{"code":"NotSupported","message":"Setting up 'AvailabilityZones' is not supported in Sku :'Developer'.

If I look at the module, the SKU parameter will always be 2 if I choose Developer. But technically that's not possible ?!

https://github.com/Azure/bicep-registry-modules/blob/c58ce674ad0876a44fc44983f1bb03b300b76eb7/avm/res/api-management/service/main.bicep#L85
https://github.com/Azure/bicep-registry-modules/blob/c58ce674ad0876a44fc44983f1bb03b300b76eb7/avm/res/api-management/service/main.bicep#L224

How to proceed to deploy APIM with the Developer SKU ?

### (Optional) Correlation Id

_No response_

### LINKED ISSUE #2617: [AVM Module Issue]: ApiManagement Loggers should depend upon NamedValues

### Issue Type?

Feature Request

### Module Name

avm/res/api-management/service

### (Optional) Module Version

0.2.0

### Description

Loggers have a credential parameter normally used for the AppInsights instrumentation key, however credentials can also reference namedValues. 

However the AVM attempts to generate the loggers and namedValues in parallel module deployments which causes deployment failures if a logger depends on a namedValue that does not yet exist, for example:

```AppInsights_InstrumentationKey (Code: ValidationError, Target: One or more Properties ['{0}'] specified are missing.)```

Propose updating the `main.bicep` so that the loggers module depends upon the namedValues module so that logger credentials can use namedValues. 

**Background**

Every AVM bicep deployment containing AppInsights or other loggers will automatically generate namedValues - _creating duplicated namedValues on every deployment_. 

I believe this is a 'known issue' in underlying APIM resource provider:
https://github.com/Azure/bicep/issues/7368
https://github.com/Azure/azure-powershell/issues/12228
https://www.kenmuse.com/blog/connecting-apim-and-appinsights-using-bicep/

The only available workaround is to manually define named values for instrumentation keys rather than using auto-generated values.

**Proposed Change**

File: [/avm/res/api-management/service/main.bicep#L428](https://github.com/Azure/bicep-registry-modules/blob/6bfef3b5819d4a8c750f41e0b0d76c2943f6afd1/avm/res/api-management/service/main.bicep#L428)

Replace existing:
```
module service_loggers 'loggers/main.bicep' = [
  for (logger, index) in loggers: {
    name: '${uniqueString(deployment().name, location)}-Apim-Logger-${index}'
    params: {
      name: logger.name
      apiManagementServiceName: service.name
      credentials: contains(logger, 'credentials') ? logger.credentials : {}
      isBuffered: contains(logger, 'isBuffered') ? logger.isBuffered : true
      loggerDescription: contains(logger, 'loggerDescription') ? logger.loggerDescription : ''
      loggerType: contains(logger, 'loggerType') ? logger.loggerType : 'azureMonitor'
      targetResourceId: contains(logger, 'targetResourceId') ? logger.targetResourceId : ''
    }
  }
]
```

with additional `dependsOn` :
```
module service_loggers 'loggers/main.bicep' = [
  for (logger, index) in loggers: {
    name: '${uniqueString(deployment().name, location)}-Apim-Logger-${index}'
    params: {
      name: logger.name
      apiManagementServiceName: service.name
      credentials: contains(logger, 'credentials') ? logger.credentials : {}
      isBuffered: contains(logger, 'isBuffered') ? logger.isBuffered : true
      loggerDescription: contains(logger, 'loggerDescription') ? logger.loggerDescription : ''
      loggerType: contains(logger, 'loggerType') ? logger.loggerType : 'azureMonitor'
      targetResourceId: contains(logger, 'targetResourceId') ? logger.targetResourceId : ''
    }
    dependsOn: [
      service_namedValues
    ]
  }
]
```

### (Optional) Correlation Id

_No response_

### PR DESCRIPTION
**Pull request by tony-box**
## Description

Closes #2561 
Closes #2617 

- Added tests for different SKUs to ensure no more issues in the future
- Disable some APIM PSRule tests that are $$ and not necessary for non-prod deployments
- fix logic to ensure various SKUs can be deployed and use availability zones and scale units
- add `dependsOn` serivice_namedValues within service_loggers to allow for a namedValue credential to be passed to loggers within same pipeline run

## Pipeline Reference

| Pipeline |
| -------- |
|    [avm.res.api-management.service](https://github.com/tony-box/bicep-registry-modules/actions/workflows/avm.res.api-management.service.yml)      |

## Type of Change

  - [x] Update to documentation

## Checklist

### CODE REVIEW

File: avm/utilities/pipelines/staticValidation/psrule/ps-rule.yaml [hunk 1]
```diff
@@ -82,3 +82,6 @@ rule:
     - Azure.Storage.UseReplication
     - Azure.AppConfig.PurgeProtect
     - Azure.APIM.MultiRegion # Team agreed this is too expensive for most use cases and is safe to ignore.
+    - Azure.APIM.AvailabilityZone # Only available for Premium SKU
```
eriqua, line 85: @jtracey93 what's your take on these suppressions and this one for zones in particular? My last update is we should prioritize WAF alignment (and resiliency in particular) over cost. Is that still the case?

If not, and if we want to go for the suppression, I think we should suppress them only for the defaults test, and keep them active for the waf-aligned one.

avm/res/api-management/service/main.bicep
```diff
@@ -73,16 +73,10 @@ param roleAssignments roleAssignmentType
   'StandardV2'
   'BasicV2'
 ])
-param sku string = 'Premium'
+param sku string = 'Premium' // Note: the default is set to Premium so that a default deployment will align with Microsoft's WAF-aligned best practices. In most cases, non-prod deployments should use a lower-tier SKU.
```
eriqua, line 76: If useful info I'd suggest to move to the description metadata

avm/res/api-management/service/tests/e2e/defaults.consumptionSku/main.test.bicep
```diff
@@ -0,0 +1,59 @@
+targetScope = 'subscription'
+
+metadata name = 'Using only defaults with a Basic SKU'
```
eriqua, line 3: Please update metadata name and description

avm/res/api-management/service/tests/e2e/defaults.developerSku/main.test.bicep
```diff
@@ -0,0 +1,59 @@
+targetScope = 'subscription'
+
+metadata name = 'Using only defaults with a Basic SKU'
```
eriqua, line 3: Please update metadata name and description

avm/utilities/pipelines/staticValidation/psrule/ps-rule.yaml
tony-box: my vote is to suppress for default tests and leave for WAF. I think most people will not want to deploy a Premium APIM sku by default, and that is unfortunately the only SKU that supports multiple AZs.

avm/res/api-management/service/tests/e2e/defaults.consumptionSku/main.test.bicep
eriqua, line 1: For this and other added tests, why calling them defaults? I think we can safely rename them to the used sku only. This is for 2 reasons:
- defaults are only using required parameters, all these tests are using additional parameters, so they are more scenario based tests and cannot be identified as defaults
- if tests folders do not contain `defaults` or `waf-aligned`, PSRule checks will not run on them. Therefore we can avoid adding PSRule suppressions for them.

avm/res/api-management/service/tests/e2e/defaults.v2Sku/main.test.bicep
```diff
@@ -46,6 +46,14 @@ module testDeployment '../../../main.bicep' = [
       sku: 'BasicV2'
       publisherEmail: 'apimgmt-noreply@mail.windowsazure.com'
       publisherName: '${namePrefix}-az-amorg-x-001'
+      managedIdentities: {
+        systemAssigned: true
+      }
+      tags: {
```
eriqua, line 52: We can avoid using tags for cleaner code if that does not add value to the specific sku test. This comment applied to all other tests as well

avm/utilities/pipelines/staticValidation/psrule/ps-rule.yaml
tony-box: The `default` and `waf-aligned` test deployments both use Premium SKUs and have multi-az enabled. Since I renamed the test folders for the other SKU deployments so they don't include "default" in the name, maybe we can add these back?

jtracey93: Hey, yup this one is a requirement even for the defaults I'm afraid. This is what we state one of the values of AVM modules are, so we must align and make it zone redundant by default and using the premium SKU.

eriqua: Hey @tony-box, as per above comment please consider the following:

- The AVM spec covering commitment for WAF-alignment: https://azure.github.io/Azure-Verified-Modules/specs/shared/#id-sfr2---category-composition---waf-aligned
- There we mention WAF **should** be set by default whenever possible. That should be prioritized over cost optimization, but it **must** be possible to override the values if needed.
- This seems very applicable to the premium sku for APIM. It is Premium by default to allow WAF properties available only for Premium. However, different skus are possible by overriding the default value. This is what the test cases for different skus are showing.

Therefore, please remove the 4 APIM exceptions from the general PSRule settings and:
- if possible, values making PSRule compliant should be set as default values
- if not possible, e.g. if they involve recommendations requiring an external dependency/resource to be deployed and configured, then we still need to try having the waf-aligned test compliant. Default tests in these exceptional cases can have suppressions. The right file where to add them is not the general PSRule settings, but this one: https://github.com/Azure/bicep-registry-modules/blob/main/avm/utilities/pipelines/staticValidation/psrule/.ps-rule/min-suppress.Rule.yaml

avm/res/api-management/service/README.md
```diff
@@ -45,12 +45,82 @@ The following section provides usage examples for the module, which were used to
...
 
-- [Using only defaults](#example-1-using-only-defaults)
-- [Using large parameter set](#example-2-using-large-parameter-set)
-- [Test deploying apim v2 sku](#example-3-test-deploying-apim-v2-sku)
-- [WAF-aligned](#example-4-waf-aligned)
+- [Using only defaults with a Consumption SKU](#example-1-using-only-defaults-with-a-consumption-sku)
+- [Using only defaults](#example-2-using-only-defaults)
+- [Using only defaults with a Developer SKU](#example-3-using-only-defaults-with-a-developer-sku)
+- [Using large parameter set](#example-4-using-large-parameter-set)
+- [Test deploying apim v2 sku](#example-5-test-deploying-apim-v2-sku)
```
eriqua, line 52: I'd suggest to update the corresponding title metadata to align with other sku tests, e.g., "Using a Basicv2 SKU"

avm/res/api-management/service/tests/e2e/developerSku/main.test.bicep
```diff
@@ -0,0 +1,54 @@
+targetScope = 'subscription'
+
+metadata name = 'Using only defaults with a Developer SKU'
```
eriqua, line 3: Do we want to show the usage of a developer sku or in particular a developer sku defaults?
- In the former case, please update title and description to not mention defaults.
- In the latter case, please remove any parameter which is not of type Required.

avm/res/api-management/service/tests/e2e/v2Sku/main.test.bicep
eriqua, line 1: Do we want to show the usage of a BasicV2 sku or in particular a BasicV2 sku defaults?

In the former case, please update title and description to not mention defaults.
In the latter case, please remove any parameter which is not of type Required.

avm/res/api-management/service/tests/e2e/consumptionSku/main.test.bicep
```diff
@@ -0,0 +1,54 @@
+targetScope = 'subscription'
+
+metadata name = 'Using only defaults with a Consumption SKU'
```
eriqua, line 3: Do we want to show the usage of a Consumption sku or in particular a Consumption sku defaults?

In the former case, please update title and description to not mention defaults.
In the latter case, please remove any parameter which is not of type Required.

avm/utilities/pipelines/staticValidation/psrule/ps-rule.yaml
tony-box: @eriqua -- already removed the exceptions in my modified PR and have been trying to get it to pass validation tests. Am  having trouble with the WAF test and some (seemingly) unrelated permission/identity issue.

avm/res/api-management/service/tests/e2e/v2Sku/main.test.bicep
```diff
@@ -1,7 +1,7 @@
 targetScope = 'subscription'
 
-metadata name = 'Test deploying apim v2 sku'
-metadata description = 'This instance deploys the module using a v2 SKU with the minimum set of required parameters.'
+metadata name = 'Deploying an APIM v2 sku.'
```
eriqua, line 3: Minor, bare with me, just for consistency with others
```suggestion
metadata name = 'Deploying an APIM v2 sku'
```

avm/res/api-management/service/main.bicep
```diff
@@ -75,14 +75,8 @@ param roleAssignments roleAssignmentType
...
 
-@description('Optional. The instance size of this API Management service. Not supported with V2 SKUs. If using Consumption, sku should = 0.')
-@allowed([
-  0
-  1
-  2
-  3
-])
-param skuCount int = 2
+@description('Conditional. The scale units for this API Management service. Required if using Basic, Standard, or Premium skus. For range of capacities for each sku, reference https://azure.microsoft.com/en-us/pricing/details/api-management/.')
```
eriqua, line 78: This is not defined clearly and under discussion even within the team, so not willing to block the PR on it, but just wanted to share my view.

If we add a default value to an input parameter, shouldn't the parameter type be Optional? At the end of the day, if the user does not provide any value, the default one is set. I see a Conditional type as a parameter for which a value is required to be passed only if a specific condition is satisfied.
In this case we are saying that the value is only set for sku Basic Standard and Premium. Developer and Consumer will be enforced to specific values.
If it was me I would add the above info to the description, keep the default value but make the parameter Optional.

Cc @AlexanderSehr as we had a brief discussion on this topic today

avm/res/api-management/service/tests/e2e/defaults/main.test.bicep
```diff
@@ -36,13 +36,23 @@ resource resourceGroup 'Microsoft.Resources/resourceGroups@2021-04-01' = {
...
+    name: '${uniqueString(deployment().name, resourceLocation)}-test-${serviceShort}-${iteration}'
+    params: {
+      name: '${namePrefix}${serviceShort}001'
+      location: resourceLocation
+      publisherEmail: 'apimgmt-noreply@mail.windowsazure.com'
+      publisherName: '${namePrefix}-az-amorg-x-001'
+      managedIdentities: {
+        systemAssigned: true
+      }
+      tags: {
```
eriqua, line 51: A default test should only include required parameters. Optional parameters should be removed.
If failing WAF, we should evaluate one of the following:
- remove the parameter from the test and add the corresponding failing rule to the min suppression group (if e.g. involving an external dependency) 
- keep the parameter in the test and make the corresponding parameter type as required
- remove the parameter from the test, keep its parameter type Optional, but add a default value satisfying the WAF condition

eriqua:
Sorry @tony-box for the long review time as I was OOF a few days. These above are my last comments before approval. Great job on this one!

avm/res/api-management/service/main.bicep
AlexanderSehr: Jap, we need to discuss thisd a bit further. One could also argue a conditional parameter is just required if something else is, regardless of it's value. You just know that in one case you **have** to set it, while in others you can or don't. Stay tuned :)

dhayes-ccl:
Any updates on when this will be merged into main?

eriqua:
Hey @tony-box, any update on this one? Please just let me know if the latest comments make sense :slightly_smiling_face:

avm/res/api-management/service/main.bicep
tony-box: Going to resolve since this is still a discussion. I'll leave it as-is for now.

tony-box:
@dhayes-ccl @eriqua sorry for the delay. I made some logic changes to address https://github.com/Azure/bicep-registry-modules/pull/2565#discussion_r1669428150 with default settings for managed identities. The automated test deployments are currently running--once they are all green, I will request re-approval.

avm/res/api-management/service/tests/e2e/defaults/main.test.bicep
tony-box: updated code to use more logic for setting managed identity. It's now completely optional and will default to a system managed identity unless explicitly overridden.

avm/res/api-management/service/main.bicep
```diff
@@ -158,14 +152,24 @@ var formattedUserAssignedIdentities = reduce(
   (cur, next) => union(cur, next)
 ) // Converts the flat array to an object like { '${id1}': {}, '${id2}': {} }
 
-var identity = !empty(managedIdentities)
+var identity = !empty(managedIdentities) && managedIdentities.?disabled == true
```
AlexanderSehr, line 155: Hey @tony-box, jumping in for @eriqua while she is off.
What is going on here? 😄 

The identity is by default nullable. Why would somebody set the property, but then set it to disabled? 
If there is a good reason, we should proably add it to the [AVM spec](https://azure.github.io/Azure-Verified-Modules/specs/shared/interfaces/#managed-identities).

avm/res/api-management/service/tests/e2e/defaults/main.test.bicep
```diff
@@ -36,13 +36,20 @@ resource resourceGroup 'Microsoft.Resources/resourceGroups@2021-04-01' = {
...
+    params: {
+      name: '${namePrefix}${serviceShort}001'
+      location: resourceLocation
+      publisherEmail: 'apimgmt-noreply@mail.windowsazure.com'
+      publisherName: '${namePrefix}-az-amorg-x-001'
+      tags: {
+        'hidden-title': 'This is visible in the resource name'
+        Environment: 'Non-Prod'
+        Role: 'DeploymentValidation'
+      }
```
AlexanderSehr, line 52: Please remove the tags from the test as they're not required :)

avm/res/api-management/service/tests/e2e/waf-aligned/main.test.bicep
```diff
@@ -182,7 +182,7 @@ module testDeployment '../../../main.bicep' = [
         }
       ]
       managedIdentities: {
-        systemAssigned: true
+        systemAssigned: false
         userAssignedResourceIds: [
           nestedDependencies.outputs.managedIdentityResourceId
         ]
```
AlexanderSehr, line 188: Required for WAF-alignment? Like in general?

avm/res/api-management/service/main.bicep
tony-box: I wanted the default behavior of the moudle to create a system assigned identity if `managedIdentities` is null or left out of the template--this is to abide PSRule requirements to have an identity assigned to APIM. 

I also wanted to retain the ability for someone to deploy the module and explicitly disable managed identites if they see fit, but I wanted that to be a very intentional decision.

avm/res/api-management/service/tests/e2e/waf-aligned/main.test.bicep
tony-box: Not requred for WAF. WAF requires that at least 1 MI is assigned to the APIM. I just wanted to have a different test case for my new MI logic than what is in the max deployment test case.

So:
max deployment = deploy both system assigned and user assigned 
waf deployment = deploy just user assigned

avm/res/api-management/service/tests/e2e/defaults/main.test.bicep
tony-box: Whoops, I missed this one, thanks. Removed.

avm/res/api-management/service/tests/e2e/waf-aligned/main.test.bicep
AlexanderSehr: Hey @tony-box,
thanks for elaborating.

avm/res/api-management/service/main.bicep
AlexanderSehr: In that case PSRule tricked you. Will reach out offline

tony-box: Changes for managedIdentity logic were reverted so the behavior is to not deploy an MI if this param is left blank. PR includes a suppression to the ManagedIdentity PSRule for APIM to get around it blocking a minimum deployment.

AlexanderSehr (approved):

dhayes-ccl:
Thanks for this fix! All looks good to me!

PR #2636, opened 2024-07-04
feat: avm.res.net-app.net-app-account

### LINKED ISSUE #2327: [Failed pipeline] avm.res.net-app.net-app-account
Failed run: https://github.com/Azure/bicep-registry-modules/actions/runs/9471673764

fbinotto:
## Description

Fixed resource duplication for one of the test cases.
Added zones to align with WAF.
Added networkFeatures to align with WAF.
Enforced region in test files to bypass quota restrictions.

## Pipeline Reference

| Pipeline |
| -------- |
|          |

[avm.res.net-app.net-app-account](https://github.com/fbinotto/bicep-registry-modules/actions/workflows/avm.res.net-app.net-app-account.yml)

## Type of Change

- [X ] Azure Verified Module updates:
  - [X ] Bugfix containing backwards-compatible bug fixes, and I have NOT bumped the MAJOR or MINOR version in `version.json`:
    - [X ] Someone has opened a bug report issue, and I have included "Closes #{bug_report_issue_number}" in the PR description.
  - [X ] Update to documentation

## Checklist

- [X ] I'm sure there are no other open Pull Requests for the same update/change
- [X ] I have run `Set-AVMModule` locally to generate the supporting module files.
- [X ] My corresponding pipelines / checks run clean and green without any errors or warnings


avm/res/net-app/net-app-account/tests/e2e/max/dependencies.bicep
```diff
@@ -1,5 +1,6 @@
-@description('Optional. The location to deploy to.')
-param location string = resourceGroup().location
+// enforcing location due to quote restrictions
+#disable-next-line no-hardcoded-location
+var enforcedLocation = 'australiaeast'
```
AlexanderSehr, line 3: No need to hardcode this here, the enforced location is passed in from outside and overwritte. In other words, the change in the `dependencies.bicep` file can be undone 😉

avm/res/net-app/net-app-account/tests/e2e/nfs3/dependencies.bicep
```diff
@@ -1,5 +1,6 @@
-@description('Optional. The location to deploy to.')
-param location string = resourceGroup().location
+// enforcing location due to quote restrictions
+#disable-next-line no-hardcoded-location
+var enforcedLocation = 'australiaeast'
```
AlexanderSehr, line 3: No need to hardcode this here, the enforced location is passed in from outside and overwritte. In other words, the change in the `dependencies.bicep` file can be undone 😉

avm/res/net-app/net-app-account/capacity-pool/main.bicep
```diff
@@ -97,6 +97,8 @@ module capacityPool_volumes 'volume/main.bicep' = [
       subnetResourceId: volume.subnetResourceId
       exportPolicyRules: contains(volume, 'exportPolicyRules') ? volume.exportPolicyRules : []
       roleAssignments: contains(volume, 'roleAssignments') ? volume.roleAssignments : []
+      networkFeatures: contains(volume, 'networkFeatures') ? volume.networkFeatures : 'Standard'
+      zones: contains(volume, 'zones') ? volume.zones : ['1']
```
AlexanderSehr, line 101:
```suggestion
      networkFeatures: volume.?networkFeatures
      zones: volume.?zones
```

This syntax should actually automatically pull the default from the referenced `volumne` module, effectively saving you one place to maintain :)

avm/res/net-app/net-app-account/tests/e2e/waf-aligned/main.test.bicep
```diff
@@ -39,12 +40,12 @@ resource resourceGroup 'Microsoft.Resources/resourceGroups@2021-04-01' = {
...
 module testDeployment '../../../main.bicep' = [
-  for iteration in ['init', 'idem']: {
+  for iteration in ['001', '002']: {
     scope: resourceGroup
-    name: '${uniqueString(deployment().name, resourceLocation)}-test-${serviceShort}-${iteration}'
+    name: '${uniqueString(deployment().name, enforcedLocation)}-test-${serviceShort}-${iteration}'
     params: {
-      name: '${namePrefix}${serviceShort}001'
-      location: resourceLocation
+      name: '${namePrefix}${serviceShort}${iteration}'
```
AlexanderSehr, line 47: It's unlikely you'd want to change this as it would mean each loop deploys a new instance. That's why the `iteration` is only used in the `deployment name` and not the resource name 😉

AlexanderSehr (changes requested):
Hey @fbinotto,

thanks for the contribution. I added a couple suggestions and comments for parts that you actually don't need to update to achieve want you want 💪 

Please take a look (:

avm/res/net-app/net-app-account/tests/e2e/waf-aligned/main.test.bicep
fbinotto: I was getting this error before and that's the reason for the changes.

"Message": "The request contains duplicate resource Ids, all resource ids must be unique and are case insensitive (e.g. resourcename1 and ResourceName1 are the same). The duplicate resources are: Resource Id '/subscriptions/***/resourcegroups/dep-***-netapp.netappaccounts-nanaawaf-rg/providers/microsoft.netapp/netappaccounts/***nanaawaf001' is listed 2 times. Please check the request and try again.",

Can you advise?

avm/res/net-app/net-app-account/tests/e2e/nfs3/dependencies.bicep
fbinotto: Agreed, thanks!

avm/res/net-app/net-app-account/capacity-pool/main.bicep
fbinotto: Agreed, thanks!

avm/res/net-app/net-app-account/tests/e2e/waf-aligned/main.test.bicep
```diff
@@ -22,6 +23,11 @@ param serviceShort string = 'nanaawaf'
 @description('Optional. A token to inject into the name of each resource. This value can be automatically injected by the CI.')
 param namePrefix string = '#_namePrefix_#'
 
+@description('Optional. The tags to apply to all resources.')
+param tags object = {
+  service: 'netapp'
+}
```
AlexanderSehr, line 29: The tags should be part of the test itself, not in a variable or parameter. For context: This is not wrong - but - our utilities extract & format the `testDeployment` block of each test folder into the readme. If a value is not in that block, it's also not going to be in the readme. Instead it would just show `tags: tags`.

avm/res/net-app/net-app-account/tests/e2e/waf-aligned/main.test.bicep
AlexanderSehr: it 'should' look like this:
```bicep
module testDeployment '../../../main.bicep' = [
  for iteration in ['init', 'idem']: {
    scope: resourceGroup
    name: '${uniqueString(deployment().name, enforcedLocation)}-test-${serviceShort}-${iteration}'
    params: {
      location: enforcedLocation
      tags: {
        service: 'netapp'
      }
    }
  }
]
```

The idea is that the `name` parameter is always the same so that the module is deployed idempotently. It definitely did work like this: https://github.com/Azure/bicep-registry-modules/actions/runs/9610748798/job/26508175247 (you can see in the deployment log that it ran 2 deployments, once with the name `init`, one with the name `idem`)

AlexanderSehr: Given that it failed with the error you're mentioning in the run right after: https://github.com/Azure/bicep-registry-modules/actions/runs/9618668724 I wonder if something on the backend-side changed. I'll try to dive a bit deeper :)

AlexanderSehr (changes requested):

avm/res/net-app/net-app-account/tests/e2e/waf-aligned/main.test.bicep
AlexanderSehr: Curiously enough, it failed with the error you're mentioning in the run right after: https://github.com/Azure/bicep-registry-modules/actions/runs/9618668724, AND the run before: https://github.com/Azure/bicep-registry-modules/actions/runs/9586083119/job/26433550470 . And there was no code change in the middle. So that doens't make much sense

avm/res/net-app/net-app-account/tests/e2e/waf-aligned/main.test.bicep
AlexanderSehr: Also interesting: I tested it now without the loop - and it still has the same error. @fbinotto I know these things are super fun, but could you open an issue in the Bicep repository for this?

I'm not even sure what this 
```json
 {
  "Code": "PreFlightRequestDuplicateResourceIds",
  "Message": "The request contains duplicate resource Ids, all resource ids must be unique and are case insensitive (e.g. resourcename1 and ResourceName1 are the same). The duplicate resources are: Resource Id '/subscriptions/<subId>/resourcegroups/dep-aslx-netapp.netappaccounts-nanaawaf-rg/providers/microsoft.netapp/netappaccounts/aslxnanaawaf001' is listed 2 times. Please check the request and try again.",
  "Target": null,
  "Details": null
}
```

is supposed to tell me.

fbinotto: @AlexanderSehr I can open the issue, no problem. However, is this really an issue? In my view, this is supposed to fail, not sure how it was working before. Because it is basically trying to deploy 2 NetApp accounts with the same name. As far as I can see because they are part of the same task and same loop, there will be an attempt to deploy both without waiting the deletion of one of them like it happens when deploying the different test cases.

avm/res/net-app/net-app-account/tests/e2e/waf-aligned/main.test.bicep
fbinotto: Thanks. Fixed.

avm/res/net-app/net-app-account/tests/e2e/waf-aligned/main.test.bicep
AlexanderSehr: Hey @fbinotto,
that's exactly the purpose of the test 😉 By default, all IaC templates should be idempotent. That means you should be able to deploy them over and over and over and the engine should only update the infrastructure if there was an update to the template. In other words (and in our case), the NetApp account should be created during the first deployment and be updated in the second (or rather not be updated as the template should be idempotent). In the specs, we call this an [idempotency test](https://azure.github.io/Azure-Verified-Modules/specs/shared/#id-snfr7---category-testing---idempotency-tests).

That being said, in the test file you may notice the annotation `@batchSize(1)` right above the test deployment. This snippet tells Bicep to deploy the code in the subsequent loop with only 1 item at a time. In other words, it deploys them in sequence as opposed to in parallel. (if you were to specify e.g., 2, it would deploy 2 items at the same time etc. etc.).

I hope that made sense 😄

fbinotto: @AlexanderSehr yes, it makes sense. How do we proceed from here? Do I create the issue and leave this PR opened until that is fixed?

AlexanderSehr: The only think that comes to mind is to remove the idempotency loop, as much as I don't want to suggest it. It's not a 'solution' but just a patch hiding the actually issue.

However, I did take a look through the past [pipeline runs](https://github.com/Azure/bicep-registry-modules/actions/workflows/avm.res.net-app.net-app-account.yml) and while it used to work very very consistently, the first [run](https://github.com/Azure/bicep-registry-modules/actions/runs/9423997333) where it stopped working does not show any code changes to the module whatsoever.

Before we do that though, please do me the favor and try one other thing, that is, revert your changes and change the name from a `001` to an `002` like
```bicep
@batchSize(1)
module testDeployment '../../../main.bicep' = [
  for iteration in ['init', 'idem']: {
    scope: resourceGroup
    name: '${uniqueString(deployment().name, resourceLocation)}-test-${serviceShort}-${iteration}'
    params: {
      name: '${namePrefix}${serviceShort}002'
     (...)
     ]
   }
]
```
and re-run the deployment tests.

The reason I'm asking is because it could be the case that something in the backend is just messed up. If that doesn't work, let's remove the loop even though it goes against the specs.

cc: @eriqua for awareness

fbinotto: Same issue @AlexanderSehr

AlexanderSehr: Damn, ok let's remove it then. @jtracey93 that's unfortunately one more module we cannot test for idempotency - be it for an unkown reason. 😞 I hope that's alright

fbinotto: @AlexanderSehr  Done. Can you please approve. Thanks

AlexanderSehr: Hey @fbinotto, much appreciated, thanks. Can you confirm that you re-ran that test and it works now? The pipeline badge is still red, hence I'm asking 😉

fbinotto: I can't rerun.. my sponsorship subscriptions got cancelled... I have requested new subscriptions but it will take some time... and then I will also have to request quota for NetApp...

AlexanderSehr: Oh my. Ok then I'll run the test for you as soon as I can locally. It's a lot less painful than merging the PR in, running the potential for the pipeline to fail and never publish 😄 Will let you know asap

AlexanderSehr: Passed: 
image

AlexanderSehr (approved):

PR #2659, opened 2024-07-05
feat: added vpa to `avm/res/container-service/managed-cluster`

PixelRobots:
Signed-off-by: PixelRobots <22979170+PixelRobots@users.noreply.github.com>

## Description

Add VPA as it has now gone GA

## Pipeline Reference

| Pipeline |
| -------- |
|    [avm.res.container-service.managed-cluster](https://github.com/PixelRobots/bicep-registry-modules/actions/workflows/avm.res.container-service.managed-cluster.yml)      |

## Type of Change

  - [ ] Update to documentation

## Checklist


avm/res/container-service/managed-cluster/main.bicep
```diff
@@ -369,6 +369,16 @@ param identityProfile object?
 @description('Optional. Enables Kubernetes Event-driven Autoscaling (KEDA).')
 param kedaAddon bool = false
 
+@description('Optional. Whether to enable VPA add-on in cluster. Default value is false.')
+param vpaAddon bool = false
```
ilhaan, line 373: @PixelRobots Is this bool redundant? Can we just use the config of `vpaAddonAutoscaling` if a user wants to use this feature or not and set the `enabled` param in `verticalPodAutoscaler`? I am trying to avoid a situation where someone sets `vpaAddonAutoscaling` but does not set `vpaAddon` to true and wonders why the change didn't go through.

avm/res/container-service/managed-cluster/main.bicep
```diff
@@ -628,6 +638,10 @@ resource managedCluster 'Microsoft.ContainerService/managedClusters@2024-03-02-p
       keda: {
         enabled: kedaAddon
       }
+      verticalPodAutoscaler: {
+        addonAutoscaling: vpaAddonAutoscaling
+        enabled: vpaAddon
```
ilhaan, line 643: Based on my earlier comment we can set this to something like this: 
```suggestion
        enabled: (vpaAddonAutoscaling == "Enabled") ? 'enabled' : 'disabled'
```
That way this feature can be controlled by one param instead of two.

PixelRobots: Looking at the API docs addonAutoscaling also enables vpa add-on but also configures vpa for the AKS addona. You might not always want that so I did it separately.

The API docs say "Whether VPA add-on is enabled and configured to scale AKS-managed add-ons."

I can do some further testing to confirm if you like.

Thoughts?

ilhaan: @PixelRobots Not sure I follow. Are you saying that we should add a check to see if addonAutoscaling is enabled before enabling vpaAddon?

PixelRobots: I believe they are independent of each other. 

Both will enable vpa. One will also configure the AKS managed addons, think CSI drivers, etc to make use of the vpa.

ilhaan: So what happens if I set my params this way: 
```bicep
vpaAddon = false
vpaAddonAutoscaling = 'Enabled'
```
Should a check be added to ensure vpaAddon is `true` before configuring `addonAutoscaling`?

avm/res/container-service/managed-cluster/main.bicep
```diff
@@ -628,6 +638,10 @@ resource managedCluster 'Microsoft.ContainerService/managedClusters@2024-03-02-p
...
+      }
```
ilhaan, line 644: Perhaps this should be something like this: 
```suggestion
      verticalPodAutoscaler: {
        addonAutoscaling: (vpaAddon != false) ? vpaAddonAutoscaling : 'Disabled'
        enabled: vpaAddon
      }
```

ilhaan (changes requested):

avm/res/container-service/managed-cluster/main.bicep
PixelRobots: Seems addonAutoscaling is in preview, so i have removed it.

avm/res/container-service/managed-cluster/main.bicep
PixelRobots: Seems addonAutoscaling is in preview, so i have removed it.

ilhaan:
@PixelRobots Thanks for the update. Can you please re-run the pipeline to verify the new updates you pushed are good to go?

PixelRobots:
> @PixelRobots Thanks for the update. Can you please re-run the pipeline to verify the new updates you pushed are good to go?

PixelRobots:
> @PixelRobots Thanks for the update. Can you please re-run the pipeline to verify the new updates you pushed are good to go?

re ran and all working.

PixelRobots:
any update on pushing this change?

AlexanderSehr:
Hey @ilhaan, could you take another look?

ilhaan (approved):

ilhaan:
@PixelRobots @AlexanderSehr approved and ready to merge.

AlexanderSehr:
@ilhaan & @JPEasier, please note, the module update was **not** published because the WAF-reliability tests did not pass. Please address them as soon as you can and make sure this code makes it into the registry.

PixelRobots:
> @ilhaan & @JPEasier, please note, the module update was **not** published because the WAF-reliability tests did not pass. Please address them as soon as you can and make sure this code makes it into the registry.
> 
> cc: @PixelRobots

I have fixed the module with this pr, apart from the default tests, which I believe should not be tested against for this ps rule. https://github.com/Azure/bicep-registry-modules/pull/3228

PR #2740, opened 2024-07-18
fix: Added baseTime parameter overwrite to CI

AlexanderSehr:
## Description

Many of our test cases implement the parameter `param baseTime string = utcNow('u')` to generate unique names.
While a good idea in theory, it has the flaw that also each retry will lead to a new value which can cause re-runs to fail if for example a key vault referenced for CMK suddenly changes.
The proposed solution is to set the value once per pipeline run in the workflow and hence overwrite the default value of each template IF the template contains the `baseTime` parameter.

## Pipeline Reference

| Pipeline |
| -------- |
|      [avm.res.compute.disk-encryption-set](https://github.com/AlexanderSehr/bicep-registry-modules/actions/workflows/avm.res.compute.disk-encryption-set.yml)   (has `baseTime` parameter)  |
| [avm.res.analysis-services.server](https://github.com/AlexanderSehr/bicep-registry-modules/actions/workflows/avm.res.analysis-services.server.yml) (does not have the `baseTime` parameter) |

## Type of Change

  - [ ] Update to documentation


.github/actions/templates/avm-validateModuleDeployment/action.yml
```diff
@@ -213,6 +213,11 @@ runs:
               resourceLocation = '${{ steps.get-resource-location.outputs.resourceLocation }}'
             }
           }
+          if($moduleTemplatePossibleParameters -contains 'baseTime') {
+            $functionInput.AdditionalParameters += @{
+              baseTime = (Get-Date).ToString('u') # Setting base time explcitly so that any potential retry will reuse the same timestamp
```
eriqua, line 218:
```suggestion
              baseTime = (Get-Date).ToString('u') # Setting base time explicitly so that any potential retry will reuse the same timestamp
```

.github/actions/templates/avm-validateModuleDeployment/action.yml
```diff
@@ -269,6 +274,11 @@ runs:
...
```
eriqua, line 279:
```suggestion
              baseTime = (Get-Date).ToString('u') # Setting base time explicitly so that any potential retry will reuse the same timestamp
```

eriqua (approved):
:OK

PR #2900, opened 2024-07-31
feat: `avm/ptn/ai-platform/baseline` Add VNet, Bastion and VM to module

cecheta:
## Description

This PR adds the second half of https://github.com/Azure/Azure-Verified-Modules/issues/1150, adding the VNet, virtual machine, and Bastion host. The configuration for the VM is based on https://github.com/Azure/azure-quickstart-templates/blob/master/quickstarts/microsoft.machinelearningservices/machine-learning-end-to-end-secure/modules/dsvmjumpbox.bicep in the Azure docs.

Note there are a number of failing tests in the pipeline that I'm not sure why they're failing, as I thought I'd fixed them:

* `Azure.Deployment.AdminUsername`
* `Azure.NSG.LateralTraversal`
* `Azure.VM.AMA`

`Azure.VNET.UseNSGs` is also failing for `AzureBastionSubnet`, however I don't think this is necessary for Bastion (and even if it was added, it would then fail `Azure.NSG.LateralTraversal`).

## Pipeline Reference

| Pipeline |
| -------- |
| [avm.ptn.ai-platform.baseline](https://github.com/cecheta/bicep-registry-modules/actions/workflows/avm.ptn.ai-platform.baseline.yml) |

## Type of Change

  - [ ] Update to documentation

## Checklist

### DIFF
--- a/avm/ptn/ai-platform/baseline/README.md
+++ b/avm/ptn/ai-platform/baseline/README.md
@@ -202,9 +202,9 @@ module baseline 'br/public:avm/ptn/ai-platform/baseline:<version>' = {
       enabled: true
       encryptionAtHost: false
       imageReference: {
-        offer: 'dsvm-win-2019'
+        offer: 'dsvm-win-2022'
         publisher: 'microsoft-dsvm'
-        sku: 'server-2019'
+        sku: 'winserver-2022'
         version: 'latest'
       }
       maintenanceConfigurationResourceId: '<maintenanceConfigurationResourceId>'
@@ -341,9 +341,9 @@ module baseline 'br/public:avm/ptn/ai-platform/baseline:<version>' = {
         "enabled": true,
         "encryptionAtHost": false,
         "imageReference": {
-          "offer": "dsvm-win-2019",
+          "offer": "dsvm-win-2022",
           "publisher": "microsoft-dsvm",
-          "sku": "server-2019",
+          "sku": "winserver-2022",
           "version": "latest"
         },
         "maintenanceConfigurationResourceId": "<maintenanceConfigurationResourceId>",
--- a/avm/ptn/ai-platform/baseline/README.md
+++ b/avm/ptn/ai-platform/baseline/README.md
@@ -105,7 +105,7 @@ module baseline 'br/public:avm/ptn/ai-platform/baseline:<version>' = {
     // Required parameters
     name: '<name>'
     // Non-required parameters
-    virtualMachineSettings: {
+    virtualMachineConfiguration: {
       adminPassword: '<adminPassword>'
       adminUsername: '<adminUsername>'
     }
@@ -130,7 +130,7 @@ module baseline 'br/public:avm/ptn/ai-platform/baseline:<version>' = {
       "value": "<name>"
     },
     // Non-required parameters
-    "virtualMachineSettings": {
+    "virtualMachineConfiguration": {
       "value": {
         "adminPassword": "<adminPassword>",
         "adminUsername": "<adminUsername>"
@@ -159,10 +159,10 @@ module baseline 'br/public:avm/ptn/ai-platform/baseline:<version>' = {
     // Required parameters
     name: 'aipbmax'
     // Non-required parameters
-    applicationInsightsSettings: {
+    applicationInsightsConfiguration: {
       name: 'appi-aipbmax'
     }
-    bastionSettings: {
+    bastionConfiguration: {
       disableCopyPaste: true
       enabled: true
       enableFileCopy: true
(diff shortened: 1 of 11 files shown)


avm/ptn/ai-platform/baseline/main.bicep
```diff
@@ -38,6 +38,41 @@ param applicationInsightsSettings applicationInsightsSettingType
...
+param bastionSettings bastionSettingType
+
+@description('Optional. Settings for the virtual machine.')
+param virtualMachineSettings virtualMachineSettingType
+
+// ============== //
+// Variables      //
+// ============== //
+
+var createVirtualNetwork = virtualNetworkSettings.?enabled != false
```
AlexanderSehr, line 54: Curious question: Instead of an extra `enabled` flag, why not apply the logic: If the user provided e.g., the `virtualNetworkSettings` then you assume they want to deploy it?

Also (it may be a bit late though, but I wanted to mention it regardless): I wonder if `*Configuration` might be a better term than `*Settings` for the parameters. Just a thought :)

avm/ptn/ai-platform/baseline/main.bicep
```diff
@@ -61,6 +96,177 @@ resource avmTelemetry 'Microsoft.Resources/deployments@2023-07-01' = if (enableT
...
+    scaleUnits: bastionSettings.?scaleUnits
+    tags: tags
+  }
+}
+
+module virtualMachine 'br/public:avm/res/compute/virtual-machine:0.5.3' = if (createVirtualMachine) {
+  name: '${uniqueString(deployment().name, location)}-virtual-machine'
+  params: {
+    name: virtualMachineSettings.?name ?? 'vm-${name}'
+    computerName: take(virtualMachineSettings.?name ?? 'vm-${name}', 15)
```
AlexanderSehr, line 216: I wonder if it might not be 'safer' to restrict the VM name to only 15 characters in the first place. Sure the 'name' can technically be longer than the 'computerName' - but at the same time just 'taking the first 15 chars' might lead to unexpected results for the user. Thoughts?

avm/ptn/ai-platform/baseline/main.bicep
cecheta: Hi @AlexanderSehr , the reason for the enabled flag is because I want the default behaviour for the module to include the VNet. So if no VNet settings are supplied, I still want the VNet deployed

Setting `virtualNetworkSettings.enabled` to `true`, would be redundant, but the flag is still necessary for those who do not need the VNet. Same for the other `enabled` flags.

As for configuration instead of settings, will happily change :)

avm/ptn/ai-platform/baseline/main.bicep
cecheta: Good idea, I've added the restriction, however I've moved the `take` to the default computer name. otherwise I would have to restrict the entire `name` param just for this one VM setting.

avm/ptn/ai-platform/baseline/main.bicep
cecheta: Renamed settings to configuration

avm/ptn/ai-platform/baseline/main.bicep
AlexanderSehr: Sounds like a reasonable compromise. Thanks @cecheta. Resolving comment...

avm/ptn/ai-platform/baseline/README.md
```diff
@@ -183,41 +283,105 @@ module baseline 'br/public:avm/ptn/ai-platform/baseline:<version>' = {
...
+    "virtualMachineConfiguration": {
+      "value": {
+        "adminPassword": "<adminPassword>",
+        "adminUsername": "<adminUsername>",
+        "enableAadLoginExtension": true,
+        "enableAzureMonitorAgent": true,
+        "enabled": true,
+        "encryptionAtHost": false,
+        "imageReference": {
+          "offer": "dsvm-win-2019",
```
ross-p-smith, line 344: Is there a 2022 version?

avm/ptn/ai-platform/baseline/main.bicep
```diff
@@ -117,55 +337,71 @@ module keyVault 'br/public:avm/res/key-vault/vault:0.6.2' = {
...
+          service: zone.value
+          subnetResourceId: subnetResourceId
+          privateDnsZoneResourceIds: [resourceId('Microsoft.Network/privateDnsZones', zone.key)]
+        })
+      : null
     roleAssignments: [
       {
         principalId: managedIdentity.properties.principalId
         roleDefinitionIdOrName: 'Contributor'
+        principalType: 'ServicePrincipal'
```
ross-p-smith, line 365: Don't want to stop this going through - but more of a question - why does it need Contributor - I would have thought the newer fine grained Storage x Contributor should do what we need

avm/ptn/ai-platform/baseline/README.md
cecheta: Yes, I wanted to use a different version than the default here to show that it's possible to supply a different image, however using the same version as the default is probably still good enough

avm/ptn/ai-platform/baseline/main.bicep
cecheta: I'm not sure, but it's recommended in the docs, and I'd be quite hesitant to change this. I imagine we would run into other issues.

avm/ptn/ai-platform/baseline/README.md
cecheta: Updated

ross-p-smith (approved):

avm/ptn/ai-platform/baseline/main.bicep
cecheta: @AlexanderSehr is this ok?

PR #2968, opened 2024-08-02
feat: Added utility to re-run failed jobs for failed workflows

AlexanderSehr:
## Description

As it happens from time to time that multiple workflows fail due to temporary issues (e.g. PSGallery not responding, throttling, etc.), this utility is intended to easy re-running the failed jobs for these workflows.
By default, the utility focuses on the `main` branch and once it ran, re-triggered all jobs, those jobs concluded, you can run the issue managing workflow to close the corresponding issues if the re-run fix the temporary issue.

Example excution
```pwsh
Invoke-FailedWorkflowsReRun
# VERBOSE: Fetching current GitHub workflows
# VERBOSE: Fetched [145] workflows
# VERBOSE: Runs to re-run failed jobs for [6/145]
# VERBOSE: Re-triggerung complete
```

## Type of Change

  - [ ] Update to documentation


avm/utilities/tools/Invoke-WorkflowsFailedJobsReRun.ps1
```diff
@@ -0,0 +1,357 @@
...
+            Headers = @{
+                Authorization = "Bearer $PersonalAccessToken"
+            }
+        }
+        $response = Invoke-RestMethod @requestInputObject
+    } else {
+        # Using GH API instead of 'gh workflow list' to get all results instead of just the first few
+        $requestInputObject = @(
+            '-H', 'Accept: application/vnd.github+json',
+            '-H', 'X-GitHub-Api-Version: 2022-11-28',
```
ChrisSidebotham, line 153: can we dynamically fetch the API Version incase we have a newer one publish?

AlexanderSehr: Hey @ChrisSidebotham, not to my knowledge. This is straight from the GH API docs.
But even if - I wonder if that would be a good call as there is a reasonable change the API would change with newer versions - and if we were using 'latest' instead of an explicit version it could break our code. Thoughts?

jtracey93 (approved):
LGTM 🚀

PR #3020, opened 2024-08-09
fix: PR logic was not working with draft PRs

rahalan:
PR label logic only kicked in, when PR was openend. If it was a draft PR, then the workflow failed (not taking into account, that no team was assigned) and also if the PR was later ready for review, the logic never was triggered again.

Two changes:
1. logic will not only be triggered when PR is openend, but also if a PR is moved into ready for review state
2. if PR is in draft state, the code will not fail, but simply no action will be taken


eriqua (approved):
🆗

PR #2763, opened 2024-07-22
feat: `avm/res/data-factory/factory` Added various UDTs

donheerschap:
## Description

Added UDTs for:

- Managed Private Endpoints
- Integration Runtimes
- Linked Services

For a better dev experience.

## Pipeline Reference

| Pipeline |
| -------- |
|    [avm.res.data-factory.factory](https://github.com/donheerschap/bicep-registry-modules/actions/workflows/avm.res.data-factory.factory.yml)      |

## Type of Change

- [ ] Update to CI Environment or utilities (Non-module affecting changes)
- [x] Azure Verified Module updates:
  - [ ] Bugfix containing backwards-compatible bug fixes, and I have NOT bumped the MAJOR or MINOR version in `version.json`:
    - [ ] Someone has opened a bug report issue, and I have included "Closes #{bug_report_issue_number}" in the PR description.
    - [ ] The bug was found by the module author, and no one has opened an issue to report it yet.
  - [x] Feature update backwards compatible feature updates, and I have bumped the MINOR version in `version.json`.
  - [ ] Breaking changes and I have bumped the MAJOR version in `version.json`.
  - [ ] Update to documentation

## Checklist

- [x] I'm sure there are no other open Pull Requests for the same update/change
- [x] I have run `Set-AVMModule` locally to generate the supporting module files.
- [x] My corresponding pipelines / checks run clean and green without any errors or warnings

PR #3117, opened 2024-08-29
feat: DataFactory - Added UDTs - `avm/res/data-factory/factory`

AlexanderSehr:
## Description

Follow-up to #2763

---
Quote: @donheerschap
Added UDTs for:

Managed Private Endpoints
Integration Runtimes
Linked Services
For a better dev experience.

> Edit: @AlexanderSehr also addressed linter warnings

## Pipeline Reference

| Pipeline |
| -------- |
|    [avm.res.data-factory.factory](https://github.com/Azure/bicep-registry-modules/actions/workflows/avm.res.data-factory.factory.yml)      |

## Type of Change

  - [ ] Update to documentation


avm/res/data-factory/factory/README.md
```diff
@@ -165,7 +165,7 @@ module factory 'br/public:avm/res/data-factory/factory:<version>' = {
         }
```
AlexanderSehr, line 165: This is just a blocking comment to ensure this PR is only merged after https://github.com/Azure/bicep-registry-modules/pull/3149.

AlexanderSehr:
Hey @clintgrove, please take a look when you get the chance :)

eriqua:
Hey @clintgrove would you have some bandwidth to review this PR? Thanks!

avm/res/data-factory/factory/main.bicep
```diff
@@ -586,3 +587,54 @@ type customerManagedKeyType = {
   @description('Optional. User assigned identity to use when fetching the customer managed key. Required if no system assigned identity is available for use.')
   userAssignedIdentityResourceId: string?
 }?
+
+type managedPrivateEndpointsType = {
```
eriqua, line 591: Sorry @clintgrove if I chime in on one point, I'll leave the rest of the PR review to you 🙂 

I think we are unofficially following a naming convention in singular form for types. If you agree I'd change this and others for consistency
```suggestion
type managedPrivateEndpointType = {
```

AlexanderSehr: Hey @eriqua, 
this would be new to be, but can be done. 
Personally, I handled it so far usually as per the following 'rules':
- If the type is singular, the name should be singular (e.g., if the PE type only discribes a single PE)
- If the type is plural (e.g., an array), the name should be plural (e.g., if the PE type describes an array of PEs).
Now in this case, the managed PE type is a an type that's automatically an array.

Maybe we should define this, what do you say @eriqua? For example, our spec-types are in fact all singular, even through they describe a plural instance.

As I'd anyways like to discuss this with you and the team (having types always be singular not only in name but also implementation), I'd suggest to have a chat some time and put if on the agenda if you'd agree? The good news is that the entire point is neither a minor nor major change and completely invisible to the consumer. So the stakes are low 😄

eriqua: Yes, the specs-types one is the reason why I'm so far unofficially following that convention.
I'd suggest going always with the singular one as easier to follow and comment on.

clintgrove: sorry guys. I am here now. I shall take a look today

avm/res/data-factory/factory/linked-service/main.bicep
```diff
@@ -41,6 +41,7 @@ resource linkedService 'Microsoft.DataFactory/factories/linkedservices@2018-06-0
           referenceName: integrationRuntimeName
           type: 'IntegrationRuntimeReference'
         }
+    #disable-next-line BCP225 // false-positive as 'type' is interpreted as a syntax value
```
clintgrove, line 44: Please help me to understand why this line of code is here?

avm/res/data-factory/factory/main.bicep
```diff
@@ -235,11 +235,9 @@ module dataFactory_integrationRuntimes 'integration-runtime/main.bicep' = [
       dataFactoryName: dataFactory.name
       name: integrationRuntime.name
       type: integrationRuntime.type
-      integrationRuntimeCustomDescription: integrationRuntime.?integrationRuntimeCustomDescription ?? 'Managed Integration Runtime created by avm-res-datafactory-factories'
-      managedVirtualNetworkName: contains(integrationRuntime, 'managedVirtualNetworkName')
-        ? integrationRuntime.managedVirtualNetworkName
-        : ''
-      typeProperties: contains(integrationRuntime, 'typeProperties') ? integrationRuntime.typeProperties : {}
+      integrationRuntimeCustomDescription: integrationRuntime.?integrationRuntimeCustomDescription
```
clintgrove, line 238: With these changes here, have you verified that when using the MAX bicep file that the ADF that is created has two IR's and a few linked services like SQL and Storage account as that is some of the changes I made previously. As long as functionality is not taken away with these proposed changes then its fine. But we must be able to add many IR's and many linked services

avm/res/data-factory/factory/main.bicep
clintgrove: I would agree to make is singular semantics

avm/res/data-factory/factory/main.json
```diff
@@ -522,21 +651,21 @@
       }
     },
     "managedPrivateEndpoints": {
-      "type": "array",
+      "$ref": "#/definitions/managedPrivateEndpointsType",
```
clintgrove, line 654: help me to understand this line of code please, never seen this before

AlexanderSehr: Hey @clintgrove, this is a reference to the 'new' `definitions: {}` section that was introduced to ARM templates following the introduction of User-defined types. What it is saying is: _The structure/definition of this parameter can be found in the `definition` map with the key `managedPrivateEndpointsType`:

https://github.com/Azure/bicep-registry-modules/blob/cc97c1df6320387193731fd13db3b7619879aa86/avm/res/data-factory/factory/main.json#L509-L544

avm/res/data-factory/factory/linked-service/main.bicep
AlexanderSehr: This snippet suppresses a warning Bicep shows you in VSCode and when building the template because it thinks it cannot resolve the value in time (presumably because it interprets a value `type` as something else than a simple parameter/variable: 
image

Just suppressing it here so that not every user of the module sees this warning they cannot do anything about.

avm/res/data-factory/factory/main.bicep
AlexanderSehr: done

avm/res/data-factory/factory/main.bicep
AlexanderSehr: It should continue to work as before without a problem. But let's dissect the individual lines to check what they should do:
- _238_: `integrationRuntimeCustomDescription: integrationRuntime.?integrationRuntimeCustomDescription`
  - IF the `integrationRuntimeCustomDescription` is provided, it will use it. Otherwise it will pass `null` which should cause Bicep to automatically pull the parameter's default `'Managed Integration Runtime created by avm-res-datafactory-factories'` from the referenced child-module, which is the same value as before the change   
- _239_: `managedVirtualNetworkName: integrationRuntime.?managedVirtualNetworkName`
  - IF the `managedVirtualNetworkName` is provided, it will use it. Otherwise it will pass `null` which should cause Bicep to automatically pull the parameter's default `''` from the referenced child-module, which is the same value as before the change   
- _240_: `integrationRuntime.?typeProperties`
  - IF the `typeProperties` is provided, it will use it. Otherwise it will pass `null` which should cause Bicep to automatically pull the parameter's default `{}` from the referenced child-module, which is the same value as before the change   

So the changes in these lines don't change anything really, but just upgrade the code to Bicep's capabilities. It's also further ppossible to deploy multiple integration runtimes etc. as the loop is encapsulating the entire `module` block which was not changed. The main change in this module are only the introduction of user-defined-types which provide intellisense to the module's user.

clintgrove (approved):
Thanks for the feedback on questions. I am ok to approve this now

PR #3317, opened 2024-09-17
fix: isHnsEnabled - `avm/res/storage/storage-account`

### LINKED ISSUE #1522: [Question/Feedback]: Ignore (unchanged) read-only properties in Bicep during deployment

Agazoth:
## Description

Fix Cannot redeploy if EnableHierargicalNamespace is false
Fixes #1522 

## Pipeline Reference

| Pipeline |
| -------- |
|     [avm.res.storage.storage-account](https://github.com/Agazoth/bicep-registry-modules/actions/workflows/avm.res.storage.storage-account.yml)     |

## Type of Change

  - [ ] Update to documentation

## Checklist


ktremain (dismissed):

bartcode:
I don't think this will fix the issue, since the issue was there if the `isHnsEnabled` property is part of the request regardless of its value - so even when it's `null`. I don't have a storage account I can test this on anymore, unfortuantely.

Agazoth:
It does. I run this version in my dev environment 2-3 times per day.

AlexanderSehr:
I'd also argue this may only fix the issue for certain storage accounts as in the following:
- If you deployed the storage account using the original implementation without setting the `enableHierarchicalNamespace` parameter, the storage account would be deployed with a value of `null` for that property
- If you deployed the storage account with the implementation suggested in this PR and not set the `enableHierarchicalNamespace` parameter, the storage account would be deployed with a value of `false` 
- If you deployed the storage account e.g., via the portal with setting the `enableHierarchicalNamespace` parameter, the value should not be set at all

I base these assumption on what was described in the original issue.

If the case, then all 3 options are not compatible with one another. @Agazoth you should be able to reproduce this by deploying a storage account via the portal and then via this implementation, or deploying an older version of the storage account module and then trying to idempotently deploy over it using this version.

The only solution that may work unfortunately (because of the way the PG handles `null` values, i.e., not ignoring them as they probably should) is to change the entire `properties: { ... }` block of the storage account resource deployment into a `properties: union( { <default> }, enableHierarchicalNamespace ? { enableHierarchicalNamespace: enableHierarchicalNamespace }, {})` implementation which would only pass the value to the deployment if `enableHierarchicalNamespace` is `true`. 
I'm writing 'unfortunately' because as soon as you introduce a `union()` function to a properties block like this, all the 'suggestions' the Bicep extension provides the module owner with in case any property is missing, goes out the window, making the maintainance a lot harder.
I hope to see the day when we see some sort of `undefined` value like [here](https://github.com/Azure/bicep/issues/387).

Agazoth:
@AlexanderSehr I have testet deploy and redeploy. These are my findings:

storage.bicep deployed with the storage-account folder with the fix in the same directory.
```bicep
// #3317 - changes line 416 in main.bicep from "isHnsEnabled: enableHierarchicalNamespace ? enableHierarchicalNamespace : null" to "isHnsEnabled: enableHierarchicalNamespace"

targetScope = 'subscription'

resource rg 'Microsoft.Resources/resourceGroups@2024-03-01' = {
  name: 'AxStorageTest'
  location: 'west europe'
}

// Deployed with no hns setting - fails on redeploy
// module storage1 'br/public:avm/res/storage/storage-account:0.13.3' = {
//   name: 'storage1'
//   scope: rg
//   params: {
//     name: 'axstoragetestnohns'
//     location: 'west europe'
//     kind: 'StorageV2'
//     skuName: 'Standard_LRS'
//   }
// }

// Deployed with hns setting true - can redeploy
module storage2 'br/public:avm/res/storage/storage-account:0.13.3' = {
  name: 'storage2'
  scope: rg
  params: {
    name: 'axstoragetestwithhnstrue'
    location: 'west europe'
    kind: 'StorageV2'
    skuName: 'Standard_LRS'
    enableHierarchicalNamespace: true
  }
}

// Deployed with hns setting false - fails on redeploy
// module storage3 'br/public:avm/res/storage/storage-account:0.13.3' = {
//   name: 'storage3'
//   scope: rg
//   params: {
//     name: 'axstoragetestwithhnsfals'
//     location: 'west europe'
//     kind: 'StorageV2'
//     skuName: 'Standard_LRS'
//     enableHierarchicalNamespace: false
//   }
// }

// Deployed with #3317 - can redeploy
module storage4 './storage-account/main.bicep'= {
  name: 'storage4'
  scope: rg
  params: {
    name: 'axfixhnsfalse'
    location: 'west europe'
    kind: 'StorageV2'
    skuName: 'Standard_LRS'
    enableHierarchicalNamespace: false
  }
}

// Deployed with #3317 - can redeploy
module storage5 './storage-account/main.bicep'= {
  name: 'storage5'
  scope: rg
  params: {
    name: 'axhnstrue'
    location: 'west europe'
    kind: 'StorageV2'
    skuName: 'Standard_LRS'
    enableHierarchicalNamespace: true
  }
}

// Deployed with #3317 - can redeploy
module storage6 './storage-account/main.bicep'= {
  name: 'storage6'
  scope: rg
  params: {
    name: 'axfixnohnsparam'
    location: 'west europe'
    kind: 'StorageV2'
    skuName: 'Standard_LRS'
  }
}

// Taking over storage account already deployed in the portal deployment
// Can redeploy - but had to add requireInfrastructureEncryption: false, since that is not set in the portal
module storage7 './storage-account/main.bicep'= {
  name: 'storage7'
  scope: rg
  params: {
    name: 'axportaldeploynoprimary'
    location: 'west europe'
    kind: 'StorageV2'
    skuName: 'Standard_LRS'
    requireInfrastructureEncryption: false
  }
}

```

Whether a resource is deployed from the portal or via the AVM module, it can redeploy with the fix.

Resources deployed with the current version 0.13.3 of the storage account module can only be redeployed if they originally had enableHierarchicalNamespace set to true.

It looks as if omitting the enableHierarchicalNamespace parameter or setting it to false removes it from the deployment when the template hits ARM. I do still believe, that the fix not only resolves the current broken state of 0.13.3, but also makes it unnecessary to do any union() of parameters.

kek-Sec:
Hey we are also facing this issue even with `enableHierarchicalNamespace` set to `false` , v `0.13.3`

AlexanderSehr:
Thanks for the thorough testing @Agazoth. I guess this cuts it then. @ktremain already approved, yet there is one conflict @Agazoth. If you could merge latest upstream main into your branch (make sure to disable actions first as new modules where merged since and would trigger in your fork) and regenerate the JSON file, that would be great 💪

Agazoth:
@AlexanderSehr should be ready to go now

AlexanderSehr (approved):

PR #2859, opened 2024-07-26
feat: Add the creation of `Microsoft.KeyVault/vaults/secrets` in Sql Server Module - `avm/res/sql/server`

jianingwang123:
## Description

Fixes #2608

Set `azure/login@v2` to use federated credential. (Due to some security policies, we are currently unable to use client secret for azure login. After completing all changes to this PR, we will restore the login method. This is currently only for testing the github pipeline.)

## Pipeline Reference

| Pipeline |
| -------- |
|    [avm.res.sql.server](https://github.com/jianingwang123/bicep-registry-modules/actions/workflows/avm.res.sql.server.yml)      |

## Type of Change

- [ ] Update to CI Environment or utilities (Non-module affecting changes)
- [x] Azure Verified Module updates:
  - [ ] Bugfix containing backwards-compatible bug fixes, and I have NOT bumped the MAJOR or MINOR version in `version.json`:
    - [ ] Someone has opened a bug report issue, and I have included "Closes #{bug_report_issue_number}" in the PR description.
    - [ ] The bug was found by the module author, and no one has opened an issue to report it yet.
  - [ ] Feature update backwards compatible feature updates, and I have bumped the MINOR version in `version.json`.
  - [ ] Breaking changes and I have bumped the MAJOR version in `version.json`.
  - [ ] Update to documentation

## Checklist

- [x] I'm sure there are no other open Pull Requests for the same update/change
- [x] I have run `Set-AVMModule` locally to generate the supporting module files.
- [x] My corresponding pipelines / checks run clean and green without any errors or warnings

@jongio - for notification.

PR #3325, opened 2024-09-18
feat: Update SQL Server API - `avm/res/sql/server`

peterbud:
* Add support for enabling IPv6 in SQL server
* Update allowed values for publicNetworkAccess
* Refactor parameter assignments to fix linter warnings

## Pipeline Reference

| Pipeline |
| -------- |
|     [avm.res.sql.server](https://github.com/Azure/bicep-registry-modules/actions/workflows/avm.res.sql.server.yml)     |

## Type of Change

  - [ ] Update to documentation

## Checklist

### DIFF
--- a/avm/res/sql/server/README.md
+++ b/avm/res/sql/server/README.md
@@ -471,7 +471,6 @@ module server 'br/public:avm/res/sql/server:<version>' = {
     ]
     elasticPools: [
       {
-        maintenanceConfigurationId: '<maintenanceConfigurationId>'
         name: 'sqlsmax-ep-001'
         skuCapacity: 10
         skuName: 'GP_Gen5'
@@ -641,7 +640,6 @@ module server 'br/public:avm/res/sql/server:<version>' = {
     "elasticPools": {
       "value": [
         {
-          "maintenanceConfigurationId": "<maintenanceConfigurationId>",
           "name": "sqlsmax-ep-001",
           "skuCapacity": 10,
           "skuName": "GP_Gen5",
@@ -829,7 +827,6 @@ param databases = [
 ]
 param elasticPools = [
   {
-    maintenanceConfigurationId: '<maintenanceConfigurationId>'
     name: 'sqlsmax-ep-001'
     skuCapacity: 10
     skuName: 'GP_Gen5'
--- a/avm/res/sql/server/README.md
+++ b/avm/res/sql/server/README.md
@@ -1299,7 +1299,6 @@ module server 'br/public:avm/res/sql/server:<version>' = {
     ]
     elasticPools: [
       {
-        maintenanceConfigurationId: '<maintenanceConfigurationId>'
         name: 'sqlswaf-ep-001'
         skuCapacity: 10
         skuName: 'GP_Gen5'
@@ -1432,7 +1431,6 @@ module server 'br/public:avm/res/sql/server:<version>' = {
     "elasticPools": {
       "value": [
         {
-          "maintenanceConfigurationId": "<maintenanceConfigurationId>",
           "name": "sqlswaf-ep-001",
           "skuCapacity": 10,
           "skuName": "GP_Gen5",
@@ -1579,7 +1577,6 @@ param databases = [
 ]
 param elasticPools = [
   {
-    maintenanceConfigurationId: '<maintenanceConfigurationId>'
     name: 'sqlswaf-ep-001'
     skuCapacity: 10
     skuName: 'GP_Gen5'
--- a/avm/res/sql/server/README.md
+++ b/avm/res/sql/server/README.md
@@ -1680,6 +1680,7 @@ param vulnerabilityAssessmentsObj = {
 | [`enableTelemetry`](#parameter-enabletelemetry) | bool | Enable/Disable usage telemetry for module. |
 | [`encryptionProtectorObj`](#parameter-encryptionprotectorobj) | object | The encryption protection configuration. |
 | [`firewallRules`](#parameter-firewallrules) | array | The firewall rules to create in the server. |
+| [`isIPv6Enabled`](#parameter-isipv6enabled) | string | Whether or not to enable IPv6 support for this server. |
 | [`keys`](#parameter-keys) | array | The keys to configure. |
 | [`location`](#parameter-location) | string | Location for all resources. |
 | [`lock`](#parameter-lock) | object | The lock settings of the service. |
@@ -1878,6 +1879,21 @@ The firewall rules to create in the server.
 - Type: array
 - Default: `[]`
 
+### Parameter: `isIPv6Enabled`
+
+Whether or not to enable IPv6 support for this server.
+
+- Required: No
+- Type: string
+- Default: `'Disabled'`
+- Allowed:
+  ```Bicep
+  [
+    'Disabled'
+    'Enabled'
+  ]
+  ```
+
 ### Parameter: `keys`
 
 The keys to configure.
@@ -1971,6 +1987,7 @@ Minimal TLS version allowed.
     '1.0'
     '1.1'
     '1.2'
+    '1.3'
   ]
   ```
 
@@ -2393,6 +2410,7 @@ Whether or not public network access is allowed for this resource. For security 
     ''
     'Disabled'
     'Enabled'
+    'SecuredByPerimeter'
   ]
   ```
 
--- a/avm/res/sql/server/README.md
+++ b/avm/res/sql/server/README.md
@@ -1296,6 +1296,7 @@ module server 'br/public:avm/res/sql/server:<version>' = {
     ]
     elasticPools: [
       {
+        maintenanceConfigurationId: '<maintenanceConfigurationId>'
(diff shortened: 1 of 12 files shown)


emichellecarter (dismissed):
Thanks so much!  Looks great!

avm/res/sql/server/README.md
```diff
@@ -1139,6 +1139,7 @@ module server 'br/public:avm/res/sql/server:<version>' = {
 | [`enableTelemetry`](#parameter-enabletelemetry) | bool | Enable/Disable usage telemetry for module. |
 | [`encryptionProtectorObj`](#parameter-encryptionprotectorobj) | object | The encryption protection configuration. |
 | [`firewallRules`](#parameter-firewallrules) | array | The firewall rules to create in the server. |
+| [`isIPv6Enabled`](#parameter-isipv6enabled) | string | Whether or not to enable IPv6 support for this server. |
```
eriqua, line 1142: Hey @peterbud thanks for this contribution, looks great!
Since new functionality is introduced I'd suggest to also update the minor version in the corresponding version.json file

peterbud: OK, I'll do that. I'm in conversation with @krbar on how to make the tests run more smoothly: currently the max tests cases are hitting regional limitations (zone redundancy). I hope that we can sort out in the coming days

AlexanderSehr: Also, an attached pipeline badge to validate that the static & deployment tests passed would be great :)

peterbud: > Hey @peterbud thanks for this contribution, looks great! Since new functionality is introduced I'd suggest to also update the minor version in the corresponding version.json file

I have bumped the version.

Also after discussing with @krbar, I have pinned the location for `max` and `waf` test to `uksouth`, to avoid issues with limited zone redundancy in some regions.

> Also, an attached pipeline badge to validate that the static & deployment tests passed would be great :) 

I can add that unfortunately only when my subscription will be re-enabled on the 10th Oct. One of the max test deployment was not deleted, and by the time I noticed I already reached the limit of my subscription.

eriqua: @krbar @AlexanderSehr in this case do you think we can have an exception and have the PR merged? Personally I'd go for it having a look at its run completion

avm/res/sql/server/elastic-pool/main.bicep
```diff
@@ -68,7 +68,7 @@ resource elasticPool 'Microsoft.Sql/servers/elasticPools@2023-08-01-preview' = {
   properties: {
     highAvailabilityReplicaCount: highAvailabilityReplicaCount
     licenseType: licenseType
-    maintenanceConfigurationId: maintenanceConfigurationId
+    maintenanceConfigurationId: !empty(maintenanceConfigurationId) ? maintenanceConfigurationId : null
```
AlexanderSehr, line 71:
```suggestion
    maintenanceConfigurationId: maintenanceConfigurationResourceId
```
Should contain `resourceId`. Also, I'd recommend to make the parameter nullable through which you don't need this logic as the default would be `null` anyways

avm/res/sql/server/README.md
AlexanderSehr: @peterbud, regarding the test resources, please enure you ran the 'toggle workflows' workflow once in your fork (it's called '.Platform: Toggle Workflows' or similar). It will disable all workflows and hence save you from a large amount of module workflows running due to their trigger on main, ultimately saving you costs. You can then enable the workflows you want to test individually (like the one for SQL-Sever).

peterbud: All workflows were/are disabled, it was my manual test which left this resource in my subscription, so it's my fault at the end.

peterbud: > @krbar @AlexanderSehr in this case do you think we can have an exception and have the PR merged? Personally I'd go for it having a look at its run completion

@eriqua Pls see the workflow run badge from @krbar, I have included that in the top entry.

ReneHezser:
@peterbud 
We have added examples for Bicep parameter files to the Readme. This has been applied to all published modules but needs to be done for PRs as well. Can you please run the Set-AVMModule utility as detailed [here](https://azure.github.io/Azure-Verified-Modules/contributing/bicep/bicep-contribution-flow/generate-bicep-module-files/). It is required for the validation pipeline to succeed and the contribution to be published.

Please reach out if any support is needed.

eriqua:
Hey @peterbud just double checking: is this PR ready for rereview?

AlexanderSehr:
I hope so, looks good to me :) 
Sidenote: Just added the team of the module owners (@peterbud & @emichellecarter) to the 'reviewers'.

avm/res/sql/server/elastic-pool/main.bicep
peterbud: @AlexanderSehr I have done further tests with this, and based on the tests this empty check IS necessary.

I have tested the following options:
* `param maintenanceConfigurationId string = ''`
* `param maintenanceConfigurationId string?` 

If there is no empty check (just passing on the incoming parameter), then both of these lead to the same error:

`Property id '' at path 'properties.maintenanceConfigurationId' is invalid. Expect fully qualified resource Id that
     | start with '/subscriptions/{subscriptionId}' or '/providers/{resourceProviderNamespace}/'.`

The only way to specify a deployment without a maintenance window is to check for empty parameter and pass the `null` explicitly:

`maintenanceConfigurationId: !empty(maintenanceConfigurationId) ? maintenanceConfigurationId : null`

So based on this I'd like to put this back.

AlexanderSehr: Hey @peterbud,
it should depend on how you want the user experience to be like. To explain myself:
- If you want you user to be able to specify an empty input value for an optional parameter you do need the input check, yes. In that case you can also have that default value `''` because your empty check will detect that no value was provided and pass `null` into to the RP on behalf of the user (this is the currently implemented behavior)
  ```bicep
  param aParameter string = ''
  
  resource dummy '(...)' = {
    properties: {
      aProperty: !empty(aParameter) ? aParameter : null
    }
  }
  ```
- The one I usually prefer (but it's absolutely up to you) is the one where I want the user to either specify a value, not specify anything at all, or if neither is an option, specify `null`. In this case, the parameter would be implemented as nullable, and the place where the parameter is used would not need an empty check, as the value allowed values are anyways only either `null` or `a proper value`.
  ```bicep
  param aParameter string?
  
  resource dummy '(...)' = {
    properties: {
      aProperty: aParameter
    }
  }
  ```

The only real difference of the two is that the user cannot specify `''` because this value would be passed through to the RP which sometimes does not support it.

Last but not least, I suspect the error you're seeing is actually not from this implementation but the way the parameter is passed in from the parent module: `maintenanceConfigurationId: elasticPool.?maintenanceConfigurationId ?? ''`. That that it really really wants to pass in `''` if the user did not specify the parameter. This should be removed and it should just say `maintenanceConfigurationId: elasticPool.?maintenanceConfigurationId` which will simply pass in `null` to the child module (or, if it would have a default there, would use that).
That last point is by the way something I've been slowly but surely cleaning up in the AVM modules I worked on. As pass null to a child means you assume the corresponding parameter's default value, the default in the parent I just called out is not needed anymore. This was only added because back in the days it was technically required to provide a default to make a parameter optional. Hence my advice: For all parameters that are passed to any other bicep file, check if the default the parent may specify via the `??` is the same as the default in the child. If so, you can safely remove the `?? (...)`  content in the parent.

> Sidenote (just some trivia): `empty(null)` evaluates to `true`

AlexanderSehr: Long story short, both implementations will work. Just wanted to give a recommendation which should work if you remove the default from the parent 😉

peterbud: Thanks, that has really helped I think. I'd also prefer the second, and I was wondering why actually has failed. Now I get it. Let me fix this.

peterbud:
> I hope so, looks good to me :) Sidenote: Just added the team of the module owners (@peterbud & @emichellecarter) to the 'reviewers'.

I have updated everything according to our earlier conversations. [Tests](https://github.com/peterbud/bicep-registry-modules/actions/runs/11306784306) are OK (although I don't have Maintenance Window available in my subscription)

Ready to merge. (I cannot approve/review my own PR)

peterbud:
> Hey @peterbud just double checking: is this PR ready for rereview?

Yes

AlexanderSehr:
Hey @peterbud,
as I just merged #2859 following your approval you have a small conflict at your hand. Please merge latest upstream in your branch and re-generate the generated files of this module via the `Set-AVMModule` utility 💪

peterbud:
> Hey @peterbud, as I just merged #2859 following your approval you have a small conflict at your hand. Please merge latest upstream in your branch and re-generate the generated files of this module via the `Set-AVMModule` utility 💪

Done

AlexanderSehr (approved):

PR #3497, opened 2024-10-09
fix: Container-App-Stack - Trigger publish

AlexanderSehr:
## Description

- Just a small change to ensure the module is published as a previous merge did not do the trick
- Also added missing passthru of `enableTelemetry` parameter to this and the insights-dashboard module


jtracey93 (approved):
LGTM 🚀

PR #3553, opened 2024-10-14
fix: AZD ML Project - Adjusted KeyVault access policies deployment

AlexanderSehr:
## Description

- Fixed AZD ML Project access policies deployment to use native Bicep code. 
- Using the AVM module in this case is not a good idea as it sets several default values (such as purge protection)

## Pipeline Reference

| Pipeline |
| -------- |
|   [avm.ptn.azd.ml-project](https://github.com/Azure/bicep-registry-modules/actions/workflows/avm.ptn.azd.ml-project.yml)       |

## Type of Change

  - [ ] Update to documentation


eriqua (commented):
@AlexanderSehr although the avm module sets secure values by default, it should always be possible to overwrite those values by explicitly passing them as input. 
Out of curiosity, why not doing that instead of using the native resource?

AlexanderSehr:
> @AlexanderSehr although the avm module sets secure values by default, it should always be possible to overwrite those values by explicitly passing them as input. Out of curiosity, why not doing that instead of using the native resource?

Great question. I'd advice against it because the unexpected overwrite of the `purgeProtection` that is the original reason I've created this PR may only be one of the properties that's overwritten. There are a lot more defaults in the Key Vault module that I'd need to guess even though the creator actually only wanted to deploy the `accessPolicies` child module. I hope that makes sense :)

eriqua:
> d this PR may only be one of the properties that's overwritten. There are a lot more defaults in the Key Vault module that I'd need to guess even though the creator actually only wanted to deploy the `accessPolicies` child module. I hope that makes sense :)

Fair for me. I guess this can be updated if we get to the child module publishing, right?

eriqua (approved):
🆗

AlexanderSehr:
> > d this PR may only be one of the properties that's overwritten. There are a lot more defaults in the Key Vault module that I'd need to guess even though the creator actually only wanted to deploy the `accessPolicies` child module. I hope that makes sense :)
> 
> Fair for me. I guess this can be updated if we get to the child module publishing, right?

Definitely 💪

PR #3563, opened 2024-10-15
fix: Common-Types - Fixed workflow configuration

AlexanderSehr:
## Description

- Fixed workflow configuration
- Implemented publish-triggering module change

## Pipeline Reference

| Pipeline |
| -------- |
|    [avm.utl.types.avm-common-types](https://github.com/Azure/bicep-registry-modules/actions/workflows/avm.utl.types.avm-common-types.yml)      |

## Type of Change

  - [ ] Update to documentation

### DIFF
--- a/avm/utl/types/avm-common-types/main.bicep
+++ b/avm/utl/types/avm-common-types/main.bicep
@@ -4,6 +4,7 @@ This module provides you with all common variants for AVM interfaces to be used 
 
 Details for how to implement these interfaces can be found in the AVM documentation [here](https://azure.github.io/Azure-Verified-Modules/specs/shared/interfaces).
 '''
+metadata owner = 'Azure/module-maintainers'
 
 //  ====================== //
 //   Diagnostic Settings   //
--- a/avm/utl/types/avm-common-types/main.bicep
+++ b/avm/utl/types/avm-common-types/main.bicep
@@ -12,7 +12,7 @@ Details for how to implement these interfaces can be found in the AVM documentat
 // Type with all properties available
 @export()
 type diagnosticSettingFullType = {
-  @description('Optional. The name of diagnostic setting.')
+  @description('Optional. The name of the diagnostic setting.')
   name: string?
 
   @description('Optional. The name of logs that will be streamed. "allLogs" includes all possible logs for the resource. Set to `[]` to disable log collection.')
--- a/avm/utl/types/avm-common-types/main.bicep
+++ b/avm/utl/types/avm-common-types/main.bicep
@@ -12,6 +12,7 @@ metadata owner = 'Azure/module-maintainers'
 
 // Type with all properties available
 @export()
+@description('An AVM-aligned type for a diagnostic setting. To be used if both logs & metrics are supported by the resource provider.')
 type diagnosticSettingFullType = {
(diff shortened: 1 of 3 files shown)


eriqua (dismissed):
🆗

avm/utl/types/avm-common-types/main.bicep
```diff
@@ -12,7 +13,7 @@ Details for how to implement these interfaces can be found in the AVM documentat
 // Type with all properties available
 @export()
 type diagnosticSettingFullType = {
-  @description('Optional. The name of diagnostic setting.')
+  @description('Optional. The name of the diagnostic setting.')
```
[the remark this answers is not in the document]
AlexanderSehr: If the types would be part of the readme - sure 😏 
But they're only added to the readme if there's a parameter that references a type. Unfortuantely, as the module does not have any parameters, there is not `parameters` section.
I already noted down somewhere that we may want to rethink how to handle UDTs in the readme - especially if they're exported.

eriqua (dismissed):
🆗

avm/utl/types/avm-common-types/main.bicep
```diff
@@ -12,7 +13,7 @@ Details for how to implement these interfaces can be found in the AVM documentat
...
   name: string?
 
   @description('Optional. The name of logs that will be streamed. "allLogs" includes all possible logs for the resource. Set to `[]` to disable log collection.')
```
AlexanderSehr, line 19: Testing something real quick for another module...

avm/utl/types/avm-common-types/main.bicep
```diff
@@ -55,8 +57,8 @@ type diagnosticSettingFullType = {
   marketplacePartnerResourceId: string?
 }
 
-// Type with only metrics available
 @export()
+@description('An AVM-aligned type for a diagnostic setting. To be used if only metrics are supported by the resource provider.')
```
eriqua, line 61: I don't know why but this additional description made me wonder if discriminators may become handy for types which are mutually exclusive. For example the discriminator type for managedIdentity may be 'systemAssigned'/'userAssigned'/'systemUserAssigned'. I'd like to have a quick look at that

AlexanderSehr: Hey @eriqua,
discriminators rely on one property being the same across all types (the discriminator). Unfortunately, this is not the case for any of the UDTs as they don't differ in property value, but the properties that are available themselves. 
Of course we could add an artifictial value, but then a user of a module (e.g., KeyVault) would need to know which of the types to choose, each time. How it would be intended right now, the module owner would need to know that, once, when creating the module (/ implementing the UDT). The module user would then not have a choice and could only pass the values the module supports.

eriqua: I meant the artificial property of course,  added on top of existing properties exactly for that purpose. An additional "type" or "mode" that the user can leverage to then access only the properties they want. 
Hence why I wanted to investigate on it, mainly to always import the same "wrapper" type, including the others.
However as you correctly said, in that case the effort of choosing the right type based on the discriminator would be on the end user rather than on the module owners.
Approving.

eriqua (approved):
🆗

Issue #1141, opened 2024-02-23
[AVM Question/Feedback]: `avm/res/sql/server` additional InteliSense for database options 


Ralph101:
### Check for previous/existing GitHub issues

- [X] I have checked for previous/existing GitHub issues

### Description

Is it possible to add some InteliSense if you are adding a database to the server. Default it wil not show any option you can add, so you need to dive into the GitHub docs how to a database to a sql server. Also when you add a Standard S1 the to maximum disk storage is to high. Maybe it's possible to add some condition based on sku type?

matebarabas:
@bryansan-msft, @Azure/avm-res-sql-server-module-owners-bicep, @Azure/avm-res-sql-server-module-contributors-bicep-bicep, this issue was just transferred from the AVM repo.

bryansan-msft:
Hello @Ralph101, I have been internally discussing with the team how we can approach this, and I will be making the necessary changes to improve the user experience when specifying parameters for the module in the next days

peterbud:
Closing as #3582 is adding the UDTs and IntelliSense for the most important objects

PR #3582, opened 2024-10-16
feat: Add strong types to `avm/res/sql/server` and it's children 

peterbud:
## Description

This PR is aiming to add strong types to parameters for `avm/res/sq/server` and it's children.

This PR contains typing for the `server`, `database` and `elastic-pool`. Other modules will be tackled later, as I'd like to see first that this is a good direction and approach adding these.

* Types are exported from children and imported into the main module
* To align with the latest API there is an SKU object for databases and elastic pools, instead of individual `skuName`, `skuTier` etc.
* Default values are provided in the child module, and the parent is not duplicating those default value assignment
* Also I have checked the latest API, and turned out that some of the new parameters were missing, so I went ahead and added them to have a complete coverage
* Updated `minCapacity` to enable fractional values both at database and elastic-pool modules
* As a consequence version number has been increased
* Updated tests where it was necessary. All tests are successfully completed

### Questions

* What is the best way to expose the types of the parameters of the child module with the least repetition? Currently I need to define a given module parameter for the module (with it's description), and then add the same to the type again with the same description. Is there a better  DRY way to do this?
* Also is there a better way to assign the values in the parent module to the parameters to the child module? Currently every parameter is assigned individually, maybe there is a better and shorter way to do that?

Contributes to #1141

## Pipeline Reference

| Pipeline |
| -------- |
|    [avm.res.sql.server](https://github.com/peterbud/bicep-registry-modules/actions/workflows/avm.res.sql.server.yml)      |

Pls note that in my subscription Maintenance Window feature is not available, hence the WAF test is failing

## Type of Change

  - [ ] Update to documentation

## Checklist

### DIFF
--- a/avm/res/sql/server/audit-settings/main.bicep
+++ b/avm/res/sql/server/audit-settings/main.bicep
@@ -16,7 +16,7 @@ param serverName string
 param state string = 'Enabled'
 
 @description('Optional. Specifies the Actions-Groups and Actions to audit.')
-param auditActionsAndGroups array = [
+param auditActionsAndGroups string[] = [
   'BATCH_COMPLETED_GROUP'
   'SUCCESSFUL_DATABASE_AUTHENTICATION_GROUP'
   'FAILED_DATABASE_AUTHENTICATION_GROUP'
--- a/avm/res/sql/server/audit-settings/main.bicep
+++ b/avm/res/sql/server/audit-settings/main.bicep
@@ -8,7 +8,7 @@ param name string
 @description('Conditional. The Name of SQL Server. Required if the template is used in a standalone deployment.')
 param serverName string
 
-@description('Optional. The resource group of the SQL Server. Required if the template is used in a standalone deployment.')
+@description('Optional. Specifies the state of the audit. If state is Enabled, storageEndpoint or isAzureMonitorTargetEnabled are required.')
 @allowed([
   'Enabled'
   'Disabled'
--- a/avm/res/sql/server/audit-settings/main.bicep
+++ b/avm/res/sql/server/audit-settings/main.bicep
@@ -8,33 +8,37 @@ param name string
 @description('Conditional. The Name of SQL Server. Required if the template is used in a standalone deployment.')
 param serverName string
 
-@description('Required. The resource group of the SQL Server. Required if the template is used in a standalone deployment.')
+@description('Optional. The resource group of the SQL Server. Required if the template is used in a standalone deployment.')
 @allowed([
   'Enabled'
   'Disabled'
 ])
-param state string
+param state string = 'Enabled'
 
 @description('Optional. Specifies the Actions-Groups and Actions to audit.')
-param auditActionsAndGroups array?
+param auditActionsAndGroups array = [
+  'BATCH_COMPLETED_GROUP'
+  'SUCCESSFUL_DATABASE_AUTHENTICATION_GROUP'
+  'FAILED_DATABASE_AUTHENTICATION_GROUP'
+]
 
 @description('Optional. Specifies whether audit events are sent to Azure Monitor.')
-param isAzureMonitorTargetEnabled bool = false
+param isAzureMonitorTargetEnabled bool = true
 
 @description('Optional. Specifies the state of devops audit. If state is Enabled, devops logs will be sent to Azure Monitor.')
-param isDevopsAuditEnabled bool?
+param isDevopsAuditEnabled bool = false
 
 @description('Optional. Specifies whether Managed Identity is used to access blob storage.')
 param isManagedIdentityInUse bool = false
 
 @description('Optional. Specifies whether storageAccountAccessKey value is the storage\'s secondary key.')
-param isStorageSecondaryKeyInUse bool?
+param isStorageSecondaryKeyInUse bool = false
 
 @description('Optional. Specifies the amount of time in milliseconds that can elapse before audit actions are forced to be processed.')
-param queueDelayMs int?
+param queueDelayMs int = 1000
 
 @description('Optional. Specifies the number of days to keep in the audit logs in the storage account.')
-param retentionDays int?
+param retentionDays int = 90
 
 @description('Optional. A blob storage to hold the auditing storage account.')
 param storageAccountResourceId string = ''
--- a/avm/res/sql/server/database/backup-long-term-retention-policy/main.bicep
+++ b/avm/res/sql/server/database/backup-long-term-retention-policy/main.bicep
@@ -14,10 +14,10 @@ param backupStorageAccessTier 'Archive' | 'Hot'?
 @description('Optional. The setting whether to make LTR backups immutable.')
 param makeBackupsImmutable bool?
 
-@description('Optional. Weekly retention in ISO 8601 duration format.')
+@description('Optional. Monthly retention in ISO 8601 duration format.')
 param monthlyRetention string?
 
-@description('Optional. Monthly retention in ISO 8601 duration format.')
+@description('Optional. Weekly retention in ISO 8601 duration format.')
 param weeklyRetention string?
 
 @description('Optional. Week of year backup to keep for yearly retention.')
--- a/avm/res/sql/server/database/backup-long-term-retention-policy/main.bicep
+++ b/avm/res/sql/server/database/backup-long-term-retention-policy/main.bicep
@@ -8,17 +8,23 @@ param serverName string
 @description('Required. The name of the parent database.')
 param databaseName string
 
-@description('Optional. Monthly retention in ISO 8601 duration format.')
-param weeklyRetention string = ''
+@description('Optional. The BackupStorageAccessTier for the LTR backups.')
+param backupStorageAccessTier 'Archive' | 'Hot'?
+
+@description('Optional. The setting whether to make LTR backups immutable.')
+param makeBackupsImmutable bool?
 
 @description('Optional. Weekly retention in ISO 8601 duration format.')
-param monthlyRetention string = ''
+param monthlyRetention string?
+
+@description('Optional. Monthly retention in ISO 8601 duration format.')
+param weeklyRetention string?
 
 @description('Optional. Week of year backup to keep for yearly retention.')
 param weekOfYear int = 1
 
 @description('Optional. Yearly retention in ISO 8601 duration format.')
-param yearlyRetention string = ''
+param yearlyRetention string?
 
 resource server 'Microsoft.Sql/servers@2023-08-01-preview' existing = {
   name: serverName
@@ -28,10 +34,12 @@ resource server 'Microsoft.Sql/servers@2023-08-01-preview' existing = {
   }
 }
 
-resource backupLongTermRetentionPolicy 'Microsoft.Sql/servers/databases/backupLongTermRetentionPolicies@2023-08-01-preview' = {
+resource backupLongTermRetentionPolicy 'Microsoft.Sql/servers/databases/backupLongTermRetentionPolicies@2023-05-01-preview' = {
   name: 'default'
   parent: server::database
   properties: {
+    backupStorageAccessTier: backupStorageAccessTier
+    makeBackupsImmutable: makeBackupsImmutable
     monthlyRetention: monthlyRetention
     weeklyRetention: weeklyRetention
     weekOfYear: weekOfYear
--- a/avm/res/sql/server/database/main.bicep
+++ b/avm/res/sql/server/database/main.bicep
@@ -42,7 +42,7 @@ param createMode
   | 'Secondary' = 'Default'
 
 @description('Optional. The resource ID of the elastic pool containing this database.')
-param elasticPoolId string?
+param elasticPoolResourceId string?
 
 @description('Optional. The azure key vault URI of the database if it\'s configured with per Database Customer Managed Keys.')
 param encryptionProtector string?
@@ -92,16 +92,16 @@ param preferredEnclaveType 'Default' | 'VBS'?
 param readScale 'Enabled' | 'Disabled' = 'Disabled'
 
 @description('Optional. The resource identifier of the recoverable database associated with create operation of this database.')
-param recoverableDatabaseId string?
+param recoverableDatabaseResourceId string?
 
 @description('Optional. The resource identifier of the recovery point associated with create operation of this database.')
-param recoveryServicesRecoveryPointId string?
+param recoveryServicesRecoveryPointResourceId string?
 
 @description('Optional. The storage account type to be used to store backups for this database.')
 param requestedBackupStorageRedundancy 'Geo' | 'GeoZone' | 'Local' | 'Zone' = 'Local'
 
 @description('Optional. The resource identifier of the restorable dropped database associated with create operation of this database.')
-param restorableDroppedDatabaseId string?
+param restorableDroppedDatabaseResourceId string?
 
 @description('Optional. Point in time (ISO8601 format) of the source database to restore when createMode set to Restore or PointInTimeRestore.')
 param restorePointInTime string?
@@ -116,7 +116,7 @@ param secondaryType 'Geo' | 'Named' | 'Standby'?
 param sourceDatabaseDeletionDate string?
 
 @description('Optional. The resource identifier of the source database associated with create operation of this database.')
-param sourceDatabaseId string?
+param sourceDatabaseResourceId string?
 
 @description('Optional. The resource identifier of the source associated with the create operation of this database.')
 param sourceResourceId string?
@@ -160,7 +160,7 @@ resource database 'Microsoft.Sql/servers/databases@2023-08-01-preview' = {
     catalogCollation: catalogCollation
     collation: collation
     createMode: createMode
-    elasticPoolId: elasticPoolId
+    elasticPoolId: elasticPoolResourceId
     encryptionProtector: encryptionProtector
     encryptionProtectorAutoRotation: encryptionProtectorAutoRotation
     federatedClientId: federatedClientId
@@ -177,15 +177,15 @@ resource database 'Microsoft.Sql/servers/databases@2023-08-01-preview' = {
     performCutover: performCutover
     preferredEnclaveType: preferredEnclaveType
     readScale: readScale
-    recoverableDatabaseId: recoverableDatabaseId
-    recoveryServicesRecoveryPointId: recoveryServicesRecoveryPointId
+    recoverableDatabaseId: recoverableDatabaseResourceId
+    recoveryServicesRecoveryPointId: recoveryServicesRecoveryPointResourceId
     requestedBackupStorageRedundancy: requestedBackupStorageRedundancy
-    restorableDroppedDatabaseId: restorableDroppedDatabaseId
+    restorableDroppedDatabaseId: restorableDroppedDatabaseResourceId
     restorePointInTime: restorePointInTime
     sampleName: sampleName
     secondaryType: secondaryType
     sourceDatabaseDeletionDate: sourceDatabaseDeletionDate
-    sourceDatabaseId: sourceDatabaseId
+    sourceDatabaseId: sourceDatabaseResourceId
     sourceResourceId: sourceResourceId
     useFreeLimit: useFreeLimit
     zoneRedundant: zoneRedundant
@@ -263,6 +263,7 @@ output location string = database.location
 //   Definitions   //
 // =============== //
 
+@export()
 type diagnosticSettingType = {
   @description('Optional. The name of diagnostic setting.')
   name: string?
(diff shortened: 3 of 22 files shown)


AlexanderSehr:
Hey @peterbud,
to your questions: @eriqua and I happened to discuss this today and will likely propose a spec to provide some guidelines. In general, this is what I'd suggest to do
- Start from the leafs of your module hierachy (i.e., the most nested child modules). There, you should have a parameter for each property of the implemented resource type, of which some may be objects & arrays of objects. For those you should implement a User Defined Type
- When you implement the User Defined Types, you should add an `@export()` annotation to make them available to calling modules. For now this is important for any of the parent modules, but eventually this will be a nice perk once we are able to published child modules too
- Now, going to the parents of those leaf-child modules: There you should have either an object or array of objects parameter for the child that is deployed (e.g., the blobServices module has an array (of objects) parameter for the containers to be deployed into the blobServies resource). As before, you should have a User Defined Type for those - but - while you should have the properties of the child being added to that user-defined-type, you should re-use the user defined types you added the `@export()` annotation for in the corresponding child. 
- If you do this in the tree all the way up, you will only need to repepat each child-modules parameters once - which should keep the types quite managable (size-wise).

There are a few smaller recommdantions we'd add to the above, such as
- ideally, a UDT should only describe how one single intance of an object looks like (even if it is for an array parameter). Whether a parameter is an array or nullable should be defined in the parameter, not the type (simply to make it less confusing). For example, rather have a parameter like `param roleAssignments roleAssignmentsType[]?` where the type is implemented like `type roleAssignmentType = {}` instead of `param roleAssignments roleAssignmentsType` where the type would be implemented like `type roleAssignmentType = {}[]?`
- Re-use types from the `types/avm-common-types` module (we just published) for the AVM specs to reduce the number of UDTs in your module - be it only for the AVM specs. But hey, it's something 😄 

Now to your question regarding 'passing each property/parameter down individually'. I have a strong guess you're referring to examples like
https://github.com/Azure/bicep-registry-modules/blob/644e64f6cf91040cf594ded1cb67c2d50d4e36e0/avm/res/sql/server/main.bicep#L248-L259

The simple answer is - no 😄 There is no 'easier' way by default. Bicep always wants you to declare the handover explicitely. However, you can shorten it slightly, which brings me to another piece of advice:
- If you pass values down to any referenced module (child or otherwise) you should check of those parameters have a default in the referenced template. In 9 out of 10 cases, the default in the parent (behind the `??`) will be a duplicate of the default in the child. In the above example that would, for example, be the case for the `skuTier` parameter.
- What you should do, is remove those defaults from the parent. In this case to change `skuTier: database.?skuTier ?? 'GeneralPurpose'` to `skuTier: database.?skuTier`. Why? Well because since nullable types were introduced, Bicep will automatically assume a referenced module's defaut values if you pass null into them. So those explicit defaults in the calling module are not needed anymore (unless you want to specify something else that the default in the referenced module, of course)
- Back when we started, there was no other way then to duplicate those values. Now, we just did not get to 'clean' all the modules up
- This won't save you from specifying the properties individually, but at least it will be a simple pass through

I hope to above provided some guidance.
@eriqua, feel free to add to it if something comes to mind.

avm/res/sql/server/elastic-pool/main.bicep
```diff
@@ -5,7 +5,7 @@ metadata owner = 'Azure/module-maintainers'
 @description('Required. The name of the Elastic Pool.')
 param name string
 
-@description('Conditional. The name of the parent SQL Server. Required if the template is used in a standalone deployment.')
+@description('Required. The name of the parent SQL Server. Required if the template is used in a standalone deployment.')
```
AlexanderSehr, line 8: Please note, this parameter is only required if using this module in a standalone deployment. If you use it via its parent, the value would actually be optional, as the `server/main.bicep` module would pass its name through. Hence we specified those parameters (that is, parent resource names) as 'Conditional'. With UDT's we could actually change that - but I'd suggest to do this 'systematically' for all modules instead of just individual ones.

avm/res/sql/server/elastic-pool/main.bicep
```diff
@@ -90,3 +102,94 @@ output resourceGroupName string = resourceGroup().name
...
+    | 'BC_DC'
+    | 'HS_Gen5'
+    | 'HS_PRMS'
+    | 'HS_MOPRMS'
+    | 'ServerlessPool'
+
+  @description('Optional. Size of the particular SKU.')
+  size: string?
+
+  @description('Required. The tier or edition of the particular SKU, e.g. Basic, Premium.')
```
AlexanderSehr, line 151: Defined as optional

avm/res/sql/server/elastic-pool/main.bicep
```diff
@@ -90,3 +102,94 @@ output resourceGroupName string = resourceGroup().name
...
+  tier: string?
+}
+
+@export()
+type elasticPoolPropertyType = {
```
AlexanderSehr, line 156: I'd recommend to move this type (including the export, don't remove that 😄) into the parent module

avm/res/sql/server/database/main.bicep
```diff
@@ -2,146 +2,148 @@ metadata name = 'SQL Server Database'
...
+  | 'PointInTimeRestore'
+  | 'Recovery'
+  | 'Restore'
+  | 'RestoreExternalBackup'
+  | 'RestoreExternalBackupSecondary'
+  | 'RestoreLongTermRetentionBackup'
+  | 'Secondary' = 'Default'
+
+@description('Optional. The resource ID of the elastic pool containing this database.')
+param elasticPoolId string?
```
AlexanderSehr, line 45:
```suggestion
param elasticPoolResourceId string?
```

as per [spec](https://azure.github.io/Azure-Verified-Modules/specs/shared/#id-snfr22---category-inputs---parametersvariables-for-resource-ids).

avm/res/sql/server/database/main.bicep
```diff
@@ -2,146 +2,148 @@ metadata name = 'SQL Server Database'
...
+@description('Optional. Whether or not customer controlled manual cutover needs to be done during Update Database operation to Hyperscale tier.')
+param manualCutover bool?
 
 @description('Optional. The max size of the database expressed in bytes.')
 param maxSizeBytes int = 34359738368
 
-@description('Optional. The name of the sample schema to apply when creating this database.')
-param sampleName string = ''
+@description('Optional. Minimal capacity that database will always have allocated.')
+param minCapacity string = '0'
```
AlexanderSehr, line 83:
```suggestion
param minCapacity int = 0
```
as per [spec](https://azure.github.io/Azure-Verified-Modules/specs/shared/#id-snfr14---category-inputs---data-types). We both know the Resource Provider won't like that, so the casting must happen where the value is used (e.g., `'${minCapacity}'`)

avm/res/sql/server/database/main.bicep
```diff
@@ -2,146 +2,148 @@ metadata name = 'SQL Server Database'
...
-param readScale string = 'Disabled'
-
-@description('Optional. The number of readonly secondary replicas associated with the database.')
-param highAvailabilityReplicaCount int = 0
+param readScale 'Enabled' | 'Disabled' = 'Disabled'
 
-@description('Optional. Minimal capacity that database will always have allocated.')
-param minCapacity string = ''
+@description('Optional. The resource identifier of the recoverable database associated with create operation of this database.')
+param recoverableDatabaseId string?
```
AlexanderSehr, line 95: Is that a resource ID? If so, it should contain `resourceId` in the name

avm/res/sql/server/database/main.bicep
```diff
@@ -2,146 +2,148 @@ metadata name = 'SQL Server Database'
...
 
-@description('Optional. Time in minutes after which database is automatically paused. A value of -1 means that automatic pause is disabled.')
-param autoPauseDelay int = 0
+@description('Optional. The resource identifier of the recovery point associated with create operation of this database.')
+param recoveryServicesRecoveryPointId string?
```
AlexanderSehr, line 98: same as previous comments about resource Ids (if applicable)

avm/res/sql/server/database/main.bicep
```diff
@@ -2,146 +2,148 @@ metadata name = 'SQL Server Database'
...
+param secondaryType 'Geo' | 'Named' | 'Standby'?
 
 @description('Optional. The time that the database was deleted when restoring a deleted database.')
-param sourceDatabaseDeletionDate string = ''
+param sourceDatabaseDeletionDate string?
 
-@description('Optional. Resource ID of backup if createMode set to RestoreLongTermRetentionBackup.')
-param recoveryServicesRecoveryPointResourceId string = ''
+@description('Optional. The resource identifier of the source database associated with create operation of this database.')
+param sourceDatabaseId string?
```
AlexanderSehr, line 119: same comment as with the other resource Ids

avm/res/sql/server/database/main.bicep
```diff
@@ -2,146 +2,148 @@ metadata name = 'SQL Server Database'
...
 
-@description('Optional. Point in time (ISO8601 format) of the source database to restore when createMode set to Restore or PointInTimeRestore.')
-param restorePointInTime string = ''
+@description('Optional. The resource identifier of the source associated with the create operation of this database.')
+param sourceResourceId string?
```
AlexanderSehr, line 122: and again - same comment as with the other resource IDs 😄

avm/res/sql/server/database/main.bicep
```diff
@@ -2,146 +2,148 @@ metadata name = 'SQL Server Database'
...
-  ''
-])
-param requestedBackupStorageRedundancy string = ''
+@description('Optional. Whether or not the database uses free monthly limits. Allowed on one database in a subscription.')
+param useFreeLimit bool?
 
-@description('Optional. Whether or not this database is a ledger database, which means all tables in the database are ledger tables. Note: the value of this property cannot be changed after the database has been created.')
-param isLedgerOn bool = false
+@description('Optional. Whether or not this database is zone redundant.')
+param zoneRedundant bool?
```
AlexanderSehr, line 128: I don't thing it had `true` default by accident. Was likely implemented this way to be WAF aligned 😉

AlexanderSehr (changes requested):
Hey @peterbud, I now added quite a few comments that will apply to all files you touched. I'll hence stop here for a moment and wait for you to respond to the current comments :)

avm/res/sql/server/database/main.bicep
peterbud: I'd argue that default zone redundancy should be off for the following reasons:

* At the Azure Portal zone redundancy off is the default setting (you can check at the Azure portal when creating a database)
* Zone redundancy is **only** available at [selected regions](https://microsoft.sharepoint.com/:f:/r/teams/csscm/Shared%20Documents/CapacityManagementTool/[KB](https://microsoft.sharepoint.com/:f:/r/teams/csscm/Shared%20Documents/CapacityManagementTool/KB?csf=1&web=1&e=SbrEU9)?csf=1&web=1&e=SbrEU9)
* Also zone redundancy is not a security feature but a _high availability_ feature.

I agree that for a WAF deployment zone redundancy is necessary, but for "plain" deployments turning it on be by default is leading to unnecessary deployment issues (not available everywhere).

peterbud:
@AlexanderSehr Thanks for the quick review, I'll make the appropriate changes.

avm/res/sql/server/database/main.bicep
peterbud: `int` would not work. If you define as `int`, you cannot pass in fractional values, like 0.5 or 0.25, which is needed for this param.

Even the [API definition](https://learn.microsoft.com/en-us/azure/templates/microsoft.sql/servers/databases?pivots=deployment-language-bicep) is saying: `minCapacity: json('decimal-as-string')`

peterbud:
I have marked the items as resolved, where I could resolve them, so PR ready for further review

peterbud:
I have also added the types for backup sub-modules, with them at least the current database and elastic-pool modules are complete

avm/res/sql/server/database/main.bicep
AlexanderSehr: Hey @peterbud, thanks for elaburating. Then that's my bad - I was under the impression it is a full number, not a decimal. Resolving

AlexanderSehr:
> > * If you do this in the tree all the way up, you will only need to repepat each child-modules parameters once - which should keep the types quite managable (size-wise).
> 
> I'm not sure I follow you here. Let's take an example: a child module called "child" and a "parent" module.
> 
> If a child has a parameter called "age", I define it in the child module, like:
> 
> ```bicep
> @description('Required. The age of the child.')
> param age int
> ```
> 
> So far so good.
> 
> But if in the parent I have a collection of children, I need to define a UDT, if I understand your suggestion. However I need to define here **again** the same property:
> 
> ```bicep
> ...
> @description ('Optional. The children of the parent')
> param children: childrenPropType[]?
> 
> @export()
> @description('The children type')
> type childrenPropType = {
>   @description('Required. The age of the child.')
>   age: int
> 
> ... //other props
> }
> ```
> 
> So this is the "repeating only once" solution, right? I was wondering how to avoid defining the same stuff twice, but as you describe its not feasible right now?

Hey @peterbud,
that's correct. Please excuse the presumably confusing explanation. Without 'showing' it's not an easy concept to convey 😄 

Under consideration of Bicep's current features, we have to do it like you describe and repeat the property once. What I was try to get to was to say that 'at least' the parent of the second module you show above does not need to describe `age: int` yet again, as it would instead import the `childPropType` type.

peterbud:
Thanks, you now it's clear!

jikuja (commented):
Really nice additions. This makes usage of the module much more easier. Few comments on review. Nothing serious.

avm/res/sql/server/database/backup-long-term-retention-policy/main.bicep
```diff
@@ -8,17 +8,23 @@ param serverName string
...
-param weeklyRetention string = ''
+@description('Optional. The BackupStorageAccessTier for the LTR backups.')
+param backupStorageAccessTier 'Archive' | 'Hot'?
+
+@description('Optional. The setting whether to make LTR backups immutable.')
+param makeBackupsImmutable bool?
 
 @description('Optional. Weekly retention in ISO 8601 duration format.')
-param monthlyRetention string = ''
+param monthlyRetention string?
```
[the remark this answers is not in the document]
jikuja: Missed the correct lines:
```
@description('Optional. Weekly retention in ISO 8601 duration format.')
param monthlyRetention string?

@description('Optional. Monthly retention in ISO 8601 duration format.')
param weeklyRetention string?
```

Looks like descriptions should be swapped.

avm/res/sql/server/audit-settings/main.bicep
```diff
@@ -8,33 +8,37 @@ param name string
 @description('Conditional. The Name of SQL Server. Required if the template is used in a standalone deployment.')
 param serverName string
 
-@description('Required. The resource group of the SQL Server. Required if the template is used in a standalone deployment.')
```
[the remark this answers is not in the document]
peterbud: Indeed, thank you!

avm/res/sql/server/database/backup-long-term-retention-policy/main.bicep
```diff
@@ -28,10 +34,12 @@ resource server 'Microsoft.Sql/servers@2023-08-01-preview' existing = {
   }
 }
 
-resource backupLongTermRetentionPolicy 'Microsoft.Sql/servers/databases/backupLongTermRetentionPolicies@2023-08-01-preview' = {
+resource backupLongTermRetentionPolicy 'Microsoft.Sql/servers/databases/backupLongTermRetentionPolicies@2023-05-01-preview' = {
```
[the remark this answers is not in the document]
peterbud: Because the latest API version has an issue, and does not recognize `backupStorageAccessTier` and `makeBackupsImmutable` properties
See the [2024-08-01 version](https://github.com/Azure/bicep-types-az/blob/main/generated/sql/microsoft.sql/2023-08-01-preview/types.md#properties-133) (wrong)
Vs the [2024-05-01 version](https://github.com/Azure/bicep-types-az/blob/main/generated/sql/microsoft.sql/2023-05-01-preview/types.md#longtermretentionpolicyproperties) (which is good)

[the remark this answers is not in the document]
jikuja: Good to know. 

For this kind of implementation decision it could be a good idea to write ADR as a comment.

avm/res/sql/server/database/main.bicep
```diff
@@ -2,145 +2,147 @@ metadata name = 'SQL Server Database'
...
 param serverName string
 
+@description('Optional. The database SKU.')
+param sku databaseSkuType = {
+  name: 'GP_Gen5_2'
+  tier: 'GeneralPurpose'
+}
+
+@description('Optional. Time in minutes after which database is automatically paused. A value of -1 means that automatic pause is disabled.')
+param autoPauseDelay int = 0
```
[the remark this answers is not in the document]
peterbud: Good catch, probably -1 would be a better default. I'll update

avm/res/sql/server/database/main.bicep
AlexanderSehr: Hey @peterbud,
I get your point, but the so-called 'reliability' pillar of WAF is one of the enforced ones for AVM. I'm pretty sure sooner or later there'll also be a failing PSRule test if we don't enable the setting. For this reason we usually aways enabled the setting for AVM's resource modules.
Lastly, the Portal is (unfortunately) not a reliable source for settings.

Anyways, let me loop @jtracey93 & @eriqua in here to collect their thoughts as well.

avm/res/sql/server/audit-settings/main.bicep
```diff
@@ -8,33 +8,37 @@ param name string
...
 @allowed([
   'Enabled'
   'Disabled'
 ])
-param state string
+param state string = 'Enabled'
 
 @description('Optional. Specifies the Actions-Groups and Actions to audit.')
-param auditActionsAndGroups array?
+param auditActionsAndGroups array = [
```
AlexanderSehr, line 19:
```suggestion
param auditActionsAndGroups string[] = [
```

avm/res/sql/server/elastic-pool/main.bicep
```diff
@@ -39,17 +43,22 @@ param maintenanceConfigurationId string?
 @description('Optional. The storage limit for the database elastic pool in bytes.')
 param maxSizeBytes int = 34359738368
 
+// TODO: Only for Serverless
```
AlexanderSehr, line 46: Leftover?

avm/res/sql/server/tests/e2e/elasticPool/main.test.bicep
```diff
@@ -0,0 +1,73 @@
...
+
+@description('Optional. The name of the resource group to deploy for testing purposes.')
+@maxLength(90)
+param resourceGroupName string = 'dep-${namePrefix}-sql.servers-${serviceShort}-rg'
+
+@description('Optional. The location to deploy resources to.')
+param resourceLocation string = deployment().location
+
+@description('Optional. A short identifier for the kind of deployment. Should be kept short to not run into resource-name length-constraints.')
+param serviceShort string = 'epool'
```
AlexanderSehr, line 18:
```suggestion
param serviceShort string = 'ssep'
```

avm/res/sql/server/database/main.bicep
eriqua: Hi @peterbud, echoing what @AlexanderSehr mentioned above.

You can reference this FAQ page to know more about [What AVM means by WAF aligned](https://azure.github.io/Azure-Verified-Modules/faq/#what-does-avm-mean-by-waf-aligned)
In particular, the further note mentioning how both reliability and security are prioritized.

Let me add that your point about zone redundancy not being available for all locations is totally valid. 

However, as mentioned by the same note, defaulted values can be anyway altered via input parameters if required. 

This means zone redundancy should be explicitly disabled if using a non supporting location. Generally, the intention is to make it more difficult rather than easier to deploy a non WAF aligned resource.

@jtracey93 happy to hear your thoughts as well

peterbud: OK, no worries, then let's keep `zoneRedundant` by default on.

avm/res/sql/server/audit-settings/main.bicep
peterbud: Good idea

avm/res/sql/server/elastic-pool/main.bicep
peterbud: I was not sure why the description specifically mentions the serverless pool. I had the impression that for elastic pool you need a provisioned server. Anyhow, let's figure out later.

avm/res/sql/server/main.bicep
```diff
@@ -249,36 +250,53 @@ module server_databases 'database/main.bicep' = [
...
+      sourceDatabaseDeletionDate: database.?sourceDatabaseDeletionDate
+      sourceDatabaseResourceId: database.?sourceDatabaseResourceId
+      sourceResourceId: database.?sourceResourceId
+      useFreeLimit: database.?useFreeLimit
+      zoneRedundant: database.?zoneRedundant
+
+      // diagnosticSettings: database.?diagnosticSettings
+      // sampleName: database.?sampleName ?? ''
+      // backupShortTermRetentionPolicy: database.?backupShortTermRetentionPolicy ?? {}
+      // backupLongTermRetentionPolicy: database.?backupLongTermRetentionPolicy ?? {}
```
AlexanderSehr, line 299: Leftover

avm/res/sql/server/main.bicep
```diff
@@ -685,7 +704,7 @@ type auditSettingsType = {
   retentionDays: int?
 
   @description('Required. Specifies the state of the audit. If state is Enabled, storageEndpoint or isAzureMonitorTargetEnabled are required.')
-  state: 'Enabled' | 'Disabled'
+  state: 'Enabled' | 'Disabled'?
```
AlexanderSehr, line 707: Optional/Required?

AlexanderSehr (commented):
Nothing more to add from my side :) @emichellecarter, do you want to give the PR a final look as the co-owner?

jikuja:
I was wondering new default settings for avm/res/sql/server/audit-settings/main.bicep:

* Is there any performance degraration when `param state string = 'Enabled'` but logging targets are disabled
* Is there any perfromance degraration when `param state string = 'Enabled'` and `isAzureMonitorTargetEnabled bool = true` but no diagnostics settings are defined
* Should avm/res/sql/server/main.bicep or avm/res/sql/server/audit-settings/main.bicep  somewhere mention that auditing module settings does not turn on any log persist methods?

I personally would like to see server module to deploy `Microsoft.Sql/servers/auditingSettings` resource with good default settings to maintain sane state of the resource:

* if user does not supply `auditSettings` => deploy resource with `state = 'Disabled'` (Here is a good set of discussions of logging should be turned on and how much configuration is needed for WAF. At the moment logging requires extra config and turning on logging without configuring any persistancy IMO gives false sense of security)
* if user removes `auditSettings` => deploy resource with `state = 'Disabled'`
* if user adds `auditSettings` => deploy resource with users' config

----
In  one of the earlier tickets there was short mention of the database logging configs. That is a good to keep in mind when re-writing server level defaults and server level logging types/documentation.

avm/res/sql/server/main.bicep
jikuja: IMO required. It user makes conscious choise to configure audit logging then they should also select if logging is enabled or disabled.

avm/res/sql/server/main.bicep
peterbud: Actually I forgot to uncomment them after defining the objects. Good catch.

avm/res/sql/server/main.bicep
peterbud: Optional, default is 'Enabled'

peterbud:
Hi @jikuja thanks for the feedback

> I was wondering new default settings for avm/res/sql/server/audit-settings/main.bicep:
> 
> * Is there any performance degraration when `param state string = 'Enabled'` but logging targets are disabled

I'm not aware of any performance degradation. If you have specific know case I'd love to hear that.

> * Is there any perfromance degraration when `param state string = 'Enabled'` and `isAzureMonitorTargetEnabled bool = true` but no diagnostics settings are defined

Same as above: I'm not aware of any performance degradation

> I personally would like to see server module to deploy `Microsoft.Sql/servers/auditingSettings` resource with good default settings to maintain sane state of the resource:
> 
> * if user does not supply `auditSettings` => deploy resource with `state = 'Disabled'` (Here is a good set of discussions of logging should be turned on and how much configuration is needed for WAF. At the moment logging requires extra config and turning on logging without configuring any persistancy IMO gives false sense of security)
> * if user removes `auditSettings` => deploy resource with `state = 'Disabled'`
> * if user adds `auditSettings` => deploy resource with users' config
> 
> In one of the earlier tickets there was short mention of the database logging configs. That is a good to keep in mind when re-writing server level defaults and server level logging types/documentation.

Auditing should be turned ON by default with the recommended settings we have in this PR. Here are the reasons:

* Look at the current pipeline runs: pre-flight failed rules (before this PR): Auditing rule is failing as no audit settings was configured by default,
* This Azure rule is pretty clear that at least a basic auditing should be in place by default: https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.SQL.Auditing/ - and that is what this PR is putting into practice

In practice this means:
* If user does not specify any `auditingSettings`, the default will deploy a basic audit settings as the above mentioned rule requires (and pre-flight will e successful): 
* If the user specifies `auditingSettings`: {state: 'Disabled'} auditing can be turned off
* If the user specifies an `auditingSettings` with Log persistence etc. that will be obviously honored.

jikuja:
The work with parameter typing is really important and missing typing was one the biggest issue when I used module. Good work!


> Auditing should be turned ON by default with the recommended settings we have in this PR. Here are the reasons:
> 
> * Look at the current pipeline runs: pre-flight failed rules (before this PR): Auditing rule is failing as no audit settings was configured by default,
> * This Azure rule is pretty clear that at least a basic auditing should be in place by default: https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.SQL.Auditing/ - and that is what this PR is putting into practice

**Saying this first: examples are easy to add afterwards. Things I write should not cause any delays with the merge.**

I personally feel that template or documentation should do more than just satisfy PsRule: at least there should be documentation how to enable logging with diagnosticsSetting. 

I till remember how hard it was to do first IaC template for Azure SQL because the only documentation for proper diagnostics setting creation was available on on quickstarts and/or by checking what portal does when enabling  auditiong.

Because Azure SQL has more settings on two levels both should be handled by template  or by the documentation to make module usage easier. It is really good to see that if `allLogs` is used by default if categories are not supplied by the user.

AlexanderSehr:
> The work with parameter typing is really important and missing typing was one the biggest issue when I used module. Good work!
> 
> > Auditing should be turned ON by default with the recommended settings we have in this PR. Here are the reasons:
> > 
> > * Look at the current pipeline runs: pre-flight failed rules (before this PR): Auditing rule is failing as no audit settings was configured by default,
> > * This Azure rule is pretty clear that at least a basic auditing should be in place by default: https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.SQL.Auditing/ - and that is what this PR is putting into practice
> 
> **Saying this first: examples are easy to add afterwards. Things I write should not cause any delays with the merge.**
> 
> I personally feel that template or documentation should do more than just satisfy PsRule: at least there should be documentation how to enable logging with diagnosticsSetting.
> 
> I till remember how hard it was to do first IaC template for Azure SQL because the only documentation for proper diagnostics setting creation was available on on quickstarts and/or by checking what portal does when enabling auditiong.
> 
> Because Azure SQL has more settings on two levels both should be handled by template or by the documentation to make module usage easier. It is really good to see that if `allLogs` is used by default if categories are not supplied by the user.

Hey @jikuja,
absolutely fair points. Ideally, there should be a few places that can help
- There should be a User-Defined-Type for every object / object-array with the parameters & metadata to explain how to configure xzy setting
- The `Usage Examples` in the `readme.md` file pull the examples from the test cases. I.e., if e.g., a [max test](https://github.com/Azure/bicep-registry-modules/tree/main/avm/res/sql/server#example-5-using-large-parameter-set) implement diagnostic settings (to test the feature), the readme should show the same. In an ideal case, this is also mentioned in the test's metadata description, which is also pulled in. 
One could of course argue there should also be a code example that just shows the diagnostic settings (or any other feature for that matter). The challenge with those is usually that one has to maintain them. In the worst case you end up with outdated/stale code examples if somebody forgets to update them with every change. Hence we chose the approach of pulled the examples from the deployment tests.

peterbud:
> * The `Usage Examples` in the `readme.md` file pull the examples from the test cases. I.e., if e.g., a [max test](https://github.com/Azure/bicep-registry-modules/tree/main/avm/res/sql/server#example-5-using-large-parameter-set) implement diagnostic settings (to test the feature), the readme should show the same. In an ideal case, this is also mentioned in the test's metadata description, which is also pulled in.

We can add later more test files / examples to address different use cases, admittedly there is a high number of different setups which might be interesting for the users.

AlexanderSehr:
> > * The `Usage Examples` in the `readme.md` file pull the examples from the test cases. I.e., if e.g., a [max test](https://github.com/Azure/bicep-registry-modules/tree/main/avm/res/sql/server#example-5-using-large-parameter-set) implement diagnostic settings (to test the feature), the readme should show the same. In an ideal case, this is also mentioned in the test's metadata description, which is also pulled in.
> 
> We can add later more test files / examples to address different use cases, admittedly there is a high number of different setups which might be interesting for the users.

True. I guess the one thing to always wheigh against is that each test means another deployment which requires capacity & agents. For most modules it's not an issue, while for some there are quite specific restrictions

peterbud:
> True. I guess the one thing to always wheigh against is that each test means another deployment which requires capacity & agents. For most modules it's not an issue, while for some there are quite specific restrictions

Maybe it would be a nice feature to be able to add more examples to the generated README.md without adding more test cases. I'm not sure how difficult that would be though, as never looked at the README generation.

peterbud:
Is there anything outstanding to this PR, or could it be merged?

AlexanderSehr:
> > True. I guess the one thing to always wheigh against is that each test means another deployment which requires capacity & agents. For most modules it's not an issue, while for some there are quite specific restrictions
> 
> Maybe it would be a nice feature to be able to add more examples to the generated README.md without adding more test cases. I'm not sure how difficult that would be though, as never looked at the README generation.

It would be doable. But not advisable as this means we'd have a lot of code that isn't tested and may turn stale in no time. Some sort of validation would always be preferred. :)

AlexanderSehr:
> Is there anything outstanding to this PR, or could it be merged?

I don't think there's anything more from my side. Ideally @emichellecarter would give it her blessing as the co-owner. But if she is not available, the repo maintainers can help out.

peterbud:
AFAIK @emichellecarter does not have a bandwidth ATM, so I don't think she would be able to review this PR

AlexanderSehr (dismissed):

AlexanderSehr:
> AFAIK @emichellecarter does not have a bandwidth ATM, so I don't think she would be able to review this PR

Gotcha @peterbud, thank you.
Please note that there is a conflict with the `main.json`. Please resolve it and (**important**) run the `Set-AVMModule` utility again after to make sure the template hash etc. is really up to date after the merge.

peterbud:
> Please note that there is a conflict with the `main.json`. Please resolve it and (**important**) run the `Set-AVMModule` utility again after to make sure the template hash etc. is really up to date after the merge.

Done.

AlexanderSehr (approved):
Thank you @peterbud,
off we go and fingers crossed. I'll keep an eye on the workflow.

peterbud:
Looks like a new static validation test failed. I'll look into that

peterbud:
I have seen that you went ahead and created the fix. Thx

AlexanderSehr:
That's right. The PR is ready to go 💪

PR #3606, opened 2024-10-22
fix: Private Endpoint updated NIC resourceID & groupID output

AlexanderSehr:
## Description

- Fixed the NIC output which was claiming to return the ID while it did return the entire object
- Updated the group ID to use the `tryGet()` function (and return `null` if not set)
- Updated UDTs to latest specs

## Pipeline Reference

| Pipeline |
| -------- |
|     [avm.res.network.private-endpoint](https://github.com/AlexanderSehr/bicep-registry-modules/actions/workflows/avm.res.network.private-endpoint.yml)     |

## Type of Change

  - [ ] Update to documentation


avm/res/network/private-endpoint/main.bicep
```diff
@@ -15,31 +15,33 @@ param applicationSecurityGroupResourceIds array?
 param customNetworkInterfaceName string?
 
 @description('Optional. A list of IP configurations of the private endpoint. This will be used to map to the First Party Service endpoints.')
-param ipConfigurations ipConfigurationsType
+param ipConfigurations ipConfigurationsType[]?
```
eriqua, line 18: Did we eventually discuss on the option of making type names singular? This comment applies to all non common types below

avm/res/network/private-endpoint/tests/e2e/max/main.test.bicep
```diff
@@ -39,7 +36,7 @@ module nestedDependencies 'dependencies.bicep' = {
   name: '${uniqueString(deployment().name, resourceLocation)}-nestedDependencies'
   params: {
     virtualNetworkName: 'dep-${namePrefix}-vnet-${serviceShort}'
-    keyVaultName: 'dep-${namePrefix}-kv-${serviceShort}-${substring(uniqueString(baseTime), 0, 3)}'
+    keyVaultName: 'dep-${namePrefix}-kv-${serviceShort}'
```
eriqua, line 39: Why this change?

avm/res/network/private-endpoint/main.bicep
AlexanderSehr: We talked about, but did not release a spec for it yet. I can update the wording for sure. In that moment my main focus was on definitely moving the array & nullable indicators 😄

AlexanderSehr: Updated @eriqua

avm/res/network/private-endpoint/main.bicep
```diff
@@ -9,37 +9,39 @@ param name string
...
-param ipConfigurations ipConfigurationsType
+param ipConfigurations ipConfigurationType[]?
 
 @description('Optional. The private DNS zone group to configure for the private endpoint.')
 param privateDnsZoneGroup privateDnsZoneGroupType?
 
 @description('Optional. Location for all Resources.')
 param location string = resourceGroup().location
 
+import { lockType } from 'br/public:avm/utl/types/avm-common-types:0.1.0'
```
ReneHezser, line 26: For easier maintenance, would it make sense to have only one import line with multiple types?

AlexanderSehr: Yes and no. So far I'm torn on this one. The benefit of having all in one is, like you said, to well - have it all in one. But the question then is where to put it. 
- Anywhere? I guess that would negate the maintanence benefit.
- Then maybe in the `// Definitions //` section. But what if a module doesn't have any other definitions. Should we then add such as section? And if we do, does that mean we have to define a spec that all type imports always have to be in a definitions section...

For now, I opted to put it right next to the parameter where it is used. Aside from the strong coupling, it has the one benefit that you can have the option to vary the version you want to use per type, instead of having to use one module version for all types.

eriqua: @ReneHezser @AlexanderSehr Good question and discussion. I'd also be more inclined towards having one import for each parameter using it. I'd actually see it more convenient from a maintenance perspective as well.

avm/res/network/private-endpoint/tests/e2e/max/main.test.bicep
eriqua: If no purge protection needed then sure!

avm/res/network/private-endpoint/main.bicep
```diff
@@ -287,12 +255,12 @@ type privateLinkServiceConnectionsType = {
...
+}
 
 type customDnsConfigType = {
   @description('Required. Fqdn that resolves to private endpoint IP address.')
   fqdn: string
 
   @description('Required. A list of private IP addresses of the private endpoint.')
   ipAddresses: string[]
-}[]?
+}
```
eriqua, line 266: Should this top level type also be exported?

AlexanderSehr: Well ... yes

AlexanderSehr: Updated. Thanks

eriqua (approved):
🆗

Issue #3378, opened 2024-09-29
[AVM Module Issue]: avm/res/operational-insights/workspace does not support deploy of SQLAuditing solution


jikuja:
### Check for previous/existing GitHub issues

- [X] I have checked for previous/existing GitHub issues

### Issue Type?

Bug

### Module Name

avm/res/operations-management/solution

### Description

Following code fails:

```bicep
module law 'br/public:avm/res/operational-insights/workspace:0.7.0' = {
  name: logAnalyticsWorspaceName
  params: {
    name: logAnalyticsWorspaceName
    enableTelemetry: false
    gallerySolutions: [
      {
        name: 'SQLAuditing'
        product: 'SQLAuditing'
        publisher: 'Microsoft'
      }
    ]
  }
}
```

Error message:

```
Solution product name cannot start with 'OMSGallery/' as it is reserved for Microsoft first party solutions. Operation Id: '8aab36af321b604584069fe60e602148' (Code: InvalidParameter, Target: plan.product)
```

Solution can be added by using solutions resource:

```bicep
var solutionName = 'SQLAuditing(${logAnalyticsWorspaceName})'
resource solution 'Microsoft.OperationsManagement/solutions@2015-11-01-preview' = {
  name: solutionName
  location: location
  properties: {
    workspaceResourceId: law.outputs.resourceId
  }
  plan: {
    name: solutionName
    promotionCode: ''
    product: 'SQLAuditing'
    publisher: 'Microsoft'
  }
}
```

-------

For me it looks like SQLAudit solution is not available on OMSGallery namespace,

https://github.com/Azure/bicep-registry-modules/blob/main/avm/res/operations-management/solution/main.bicep#L48:

```bicep
var solutionProduct = publisher == 'Microsoft' ? 'OMSGallery/${name}' : product
```

--------

For a reference Portal creates following resource(not the call but resulting stete of the resource) when turning on auditing to LAW on Azure SQL:

```json
{
  "plan": {
    "name": "SQLAuditing[law-law]",
    "publisher": "Microsoft",
    "promotionCode": "",
    "product": "SQLAuditing",
    "version": "1.0"
  },
  "properties": {
    "workspaceResourceId": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/RgName/providers/Microsoft.OperationalInsights/workspaces/law-law",
    "provisioningState": "Succeeded",
    "creationTime": "Thu, 26 Sep 2024 14:55:20 GMT",
[...]

PR #3671, opened 2024-10-30
fix: remove parameter composing logic from `avm/res/operations-management/solution` 

krbar:
## Description

This PR addresses the issue with some solutions, which don't follow the naming patterns described in the [template reference](https://learn.microsoft.com/en-us/azure/templates/microsoft.operationsmanagement/solutions?pivots=deployment-language-bicep). To address this, the module passes the parameters to the resource provider without attempting to compose the names. It is the responsibility of the user to provide parameters expected by the resource provider. See parameter descriptions for more details. 
Please note that this is a breaking change, the parameters of the existing deployments have to be updated to match the new format.

Related to #3378 (first part of the fix).

## Pipeline Reference

| Pipeline |
| -------- |
|       [avm.res.operations-management.solution](https://github.com/krbar/bicep-registry-modules/actions/workflows/avm.res.operations-management.solution.yml)   |

## Type of Change

  - [ ] Update to documentation

## Checklist

### DIFF
--- a/avm/res/operations-management/solution/main.bicep
+++ b/avm/res/operations-management/solution/main.bicep
@@ -2,15 +2,9 @@ metadata name = 'Operations Management Solutions'
 metadata description = 'This module deploys an Operations Management Solution.'
 metadata owner = 'Azure/module-maintainers'
 
-@metadata({
-  example: '''
-  - AntiMalware(contoso-Logs)
-  - MyThirdPartySolution[contoso-Logs]
-  '''
-})
 @description('''Required. Name of the solution.
-For solutions authored by Microsoft, the name must be in the pattern: `SolutionType(WorkspaceName)`.
-For solutions authored by third parties, the name should be in the pattern: `SolutionType[WorkspaceName]`.
+For solutions authored by Microsoft, the name must be in the pattern: `SolutionType(WorkspaceName)`, for example: `AntiMalware(contoso-Logs)`.
+For solutions authored by third parties, the name should be in the pattern: `SolutionType[WorkspaceName]`, for example `MySolution[contoso-Logs]`.
 The solution type is case-sensitive.''')
 param name string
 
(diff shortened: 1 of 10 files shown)


avm/res/operations-management/solution/main.bicep
```diff
@@ -2,7 +2,16 @@ metadata name = 'Operations Management Solutions'
 metadata description = 'This module deploys an Operations Management Solution.'
 metadata owner = 'Azure/module-maintainers'
 
-@description('Required. Name of the solution. For Microsoft published gallery solution the target solution resource name will be composed as `{name}({logAnalyticsWorkspaceName})`.')
+@metadata({
+  example: '''
```
AlexanderSehr, line 6: Why not add the example to the description so that it's part of the readme?

AlexanderSehr: Nevermind, I see it is also in the description - which raises the opposite question: Why? 😄

avm/res/operations-management/solution/main.bicep
```diff
@@ -11,8 +20,11 @@ param logAnalyticsWorkspaceName string
 @description('Optional. Location for all resources.')
 param location string = resourceGroup().location
 
-@description('Optional. The product of the deployed solution. For Microsoft published gallery solution it should be `OMSGallery` and the target solution resource product will be composed as `OMSGallery/{name}`. For third party solution, it can be anything. This is case sensitive.')
-param product string = 'OMSGallery'
```
AlexanderSehr, line 15: If I see this correctly, the logic seemed to have been 'If `publisher != 'Microsoft'` the default 'OMSGallery' should be used for the `product` property. Would it make sense to keep the default, or did that default never really make much sense?

krbar: In fact, it didn't make sense. If `publisher != 'Microsoft'`, the value must not start with `'OMSGallery'` and `publisher == 'Microsoft'`, the value was always converted to `OMSGallery/{name}`. So the default was actually never passed to the resource provider...

avm/res/operations-management/solution/main.bicep
krbar: I tried the example metadata but I couldn't locate the example in the readme. Let me check again, maybe I overlooked something...

AlexanderSehr: That's my point. It's not a supported property 😄 I'd just keep the example in the description and remove the other

krbar: Now it should be fine! :)

AlexanderSehr (approved):

PR #3832, opened 2024-11-22
fix: update readme to pass static test

rahalan:
## Description

Fixes #3829 

## Pipeline Reference

| Pipeline |
| -------- |
|     [avm.res.compute.virtual-machine-scale-set](https://github.com/rahalan/bicep-registry-modules/actions/workflows/avm.res.compute.virtual-machine-scale-set.yml)     |

## Type of Change

- [ ] Update to CI Environment or utilities (Non-module affecting changes)
- [ ] Azure Verified Module updates:
  - [x] Bugfix containing backwards-compatible bug fixes, and I have NOT bumped the MAJOR or MINOR version in `version.json`:
    - [ ] Someone has opened a bug report issue, and I have included "Closes #{bug_report_issue_number}" in the PR description.
    - [ ] The bug was found by the module author, and no one has opened an issue to report it yet.
  - [ ] Feature update backwards compatible feature updates, and I have bumped the MINOR version in `version.json`.
  - [ ] Breaking changes and I have bumped the MAJOR version in `version.json`.
  - [ ] Update to documentation

## Checklist

- [x] I'm sure there are no other open Pull Requests for the same update/change
- [x] I have run `Set-AVMModule` locally to generate the supporting module files.
- [x] My corresponding pipelines / checks run clean and green without any errors or warnings

PR #3834, opened 2024-11-22
fix: update bicep version -> new json

rahalan:
## Description

Missed the extension json. This needed also a regeneration after the new Bicep version was released.

## Pipeline Reference

| Pipeline |
| -------- |
|       [avm.res.compute.virtual-machine](https://github.com/rahalan/bicep-registry-modules/actions/workflows/avm.res.compute.virtual-machine.yml)  |

## Type of Change

  - [ ] Update to documentation

## Checklist

### DIFF
--- a/avm/res/compute/virtual-machine/extension/main.json
+++ b/avm/res/compute/virtual-machine/extension/main.json
@@ -5,8 +5,8 @@
   "metadata": {
     "_generator": {
       "name": "bicep",
-      "version": "0.31.34.60546",
-      "templateHash": "1194243367873711347"
+      "version": "0.31.92.45157",
+      "templateHash": "688718350646227538"
     },
     "name": "Virtual Machine Extensions",
     "description": "This module deploys a Virtual Machine Extension.",
@@ -121,7 +121,10 @@
         "settings": "[if(not(empty(parameters('settings'))), parameters('settings'), null())]",
         "protectedSettings": "[if(not(empty(parameters('protectedSettings'))), parameters('protectedSettings'), null())]",
         "suppressFailures": "[parameters('supressFailures')]"
-      }
+      },
+      "dependsOn": [
+        "virtualMachine"
+      ]
     }
   },
   "outputs": {


ReneHezser (approved):
✅

avm/res/compute/virtual-machine/extension/main.json
```diff
@@ -5,8 +5,8 @@
   "metadata": {
     "_generator": {
       "name": "bicep",
-      "version": "0.31.34.60546",
-      "templateHash": "1194243367873711347"
+      "version": "0.31.92.45157",
```
eriqua, line 8: This only will not qualify for publishing

rahalan: @eriqua can't follow. What is needed in your opinion? The whole update is only, because Bicep released a new version. Nothing changed in the module.

eriqua: If no version to be released is stuck then it's fine, but it's important we're all aligned on how publishing works.

- If main.json is updated then a new version is published.
- If the pipeline fails, the version is not published until another PR is open, where the main.json is updated again.

So my comment here is, since avm/res/compute/virtual-machine/main.json is not updated in this PR, but it was in the previous one, which failed, the corresponding new version will not be published with this PR.

This is similar to the comment I added to #3832, but in that case is more relevant.

For this PR, if we want a new version to be published, we need a small update to avm/res/compute/virtual-machine/main.json as well. If that is not needed, then we're good to go and you can resolve this comment.

AlexanderSehr: Wow - I prepared a beautiful comment to +1 on @eriqua's point and instead of clicking on comment, I accidently selected 'resolve conversation'. Shame on me.

Also a shame that my text is gone - so let me start over in a short version:
- Your next action should be determined based on whether should should (have) lead to a new version being published or not
- If the PR was just added to fix an issue with the template with new intention of publishing a new version, there was no added action needed.
- If you actually wanted a new version (e.g., because previous changes were not published due to this template issue), then a dummy change (outside the module's metadata) would've been required. For example in the a module parameter's description.

@eriqua, I'm terribly sorry. @rahalan, please create or not create a new PR based on whether you wanted to have a new published version 😉

PR #3911, opened 2024-12-10
chore: lower case kind value to avoid diff everytime

### LINKED ISSUE #3909: [AVM Module Issue]: Change @allowed values in "kind" to lowercase to match actual values

### Issue Type?

Feature Request

### Module Name

avm/res/web/serverfarm

### (Optional) Module Version

_No response_

### Description

Current values allowed are using upper-case letters, which is causing a modify to happen at every run.
 ```
@description('Optional. Kind of server OS.')
@allowed([
  'App'
  'Elastic'
  'FunctionApp'
  'Windows'
  'Linux'
])
param kind string = 'App'
 ```
What-if output: 
 ```
  ~ Microsoft.Web/serverfarms/asp-xxx-xxx [2022-09-01]
    ~ kind:                                 "linux" => "Linux"
 ```
The ``allowed`` decorator is case-sensitive, so values must be fixed in the module
```
@description('Optional. Kind of server OS.')
@allowed([
  'app'
  'elastic'
  'functionapp'
  'windows'
  'linux'
])
param kind string = 'app'
```

It's not causing a functional issue, it's just kind of annoying always seeing this in every pipeline run.

### (Optional) Correlation Id

_No response_

pankajagrawal16:
## Description

Closes #3909 

## Pipeline Reference

| Pipeline |
| -------- |
|      [avm.res.web.serverfarm](https://github.com/pankajagrawal16/bicep-registry-modules/actions/workflows/avm.res.web.serverfarm.yml)    |

## Type of Change

  - [ ] Update to documentation

## Checklist

### DIFF
--- a/avm/res/web/serverfarm/main.bicep
+++ b/avm/res/web/serverfarm/main.bicep
@@ -27,16 +27,16 @@ param location string = resourceGroup().location
 
 @description('Optional. Kind of server OS.')
 @allowed([
-  'App'
-  'Elastic'
-  'FunctionApp'
-  'Windows'
-  'Linux'
+  'app'
+  'elastic'
+  'functionApp'
+  'windows'
+  'linux'
 ])
-param kind string = 'App'
+param kind string = 'app'
 
 @description('Conditional. Defaults to false when creating Windows/app App Service Plan. Required if creating a Linux App Service Plan and must be set to true.')
-param reserved bool = (kind == 'Linux')
+param reserved bool = (kind == 'linux')
 
 @description('Optional. The Resource ID of the App Service Environment to use for the App Service Plan.')
 param appServiceEnvironmentId string = ''
--- a/avm/res/web/serverfarm/README.md
+++ b/avm/res/web/serverfarm/README.md
@@ -125,7 +125,7 @@ module serverfarm 'br/public:avm/res/web/serverfarm:<version>' = {
         workspaceResourceId: '<workspaceResourceId>'
       }
     ]
-    kind: 'App'
+    kind: 'app'
     location: '<location>'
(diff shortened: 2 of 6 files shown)


avm/res/web/serverfarm/main.bicep
```diff
@@ -136,7 +136,7 @@ resource avmTelemetry 'Microsoft.Resources/deployments@2024-03-01' = if (enableT
 
 resource appServicePlan 'Microsoft.Web/serverfarms@2022-09-01' = {
   name: name
-  kind: kind
+  kind: toLower(kind)
```
AlexanderSehr, line 139: Wonder if it would be better to just update the allowed set at the parameter :)

pankajagrawal16: In order to keep things backward compatible, this seemed like better option?

AlexanderSehr: As we're not yet in version 1.0 and because this is very very easy to adjust on the consumer side, I'd still vote to rather 'fix' the underlying issue. Thanks to the versioning we can do those changes 😉

pankajagrawal16: Ok then i will update accordingly and bump up the version

pankajagrawal16: @AlexanderSehr Have a look now!

tsc-buddy (approved):

avm/res/web/serverfarm/main.bicep
AlexanderSehr: Looks great. Thanks for the thorough work.

PR #4033, opened 2024-12-20
fix: Expose Application gateway through https

### LINKED ISSUE #3794: [Failed pipeline] avm.ptn.aca-lza.hosting-environment
Failed run: https://github.com/Azure/bicep-registry-modules/actions/runs/11854905184

kpantos:
## Description
The release enables the exposing of the container apps environment through an Application Gateway with secure https protocol endpoint thus addressing and fixing the following issue:
- Fixes #3794 

## Pipeline Reference

[avm.ptn.aca-lza.hosting-environment](https://github.com/kpantos/bicep-registry-modules/actions/workflows/avm.ptn.aca-lza.hosting-environment.yml)

## Type of Change

- [X] Azure Verified Module updates:
  - [X] Feature update backwards compatible feature updates, and I have bumped the MINOR version in `version.json`.
  - [ ] Update to documentation

## Checklist


avm/ptn/aca-lza/hosting-environment/tests/configuration/acahello.demoapp.com.key
AlexanderSehr, line 1: Hey @kpantos,
unfrotauntely this won't happen, even though these files are liklely only here for testing. 

If needed during testing, I highly recommend using a KeyVault in combination with either one of the scripts in the [e2e-template-assets](https://github.com/Azure/bicep-registry-modules/blob/main/utilities/e2e-template-assets/scripts) folder like [Set-PfxCertificateInKeyVault](https://github.com/Azure/bicep-registry-modules/blob/main/utilities/e2e-template-assets/scripts/Set-PfxCertificateInKeyVault.ps1), or create a custom script.

In general, we should not store any secrets/certificates/keys in our source code but generate them on the fly. cc: @jtracey93

kpantos: Hey @AlexanderSehr 
Can you please provide the reasoning behind why there shouldn’t be a test certificate that’s used only for testing, only binded to an https listener (not used for any authentication) which is not going to be part of the released module anyway?
In order to generate at runtime and store at keyvault for testing it will require standing up 3 new test resource groups, key vaults and identities to manage secrets which will increase the already long testing period and test complexity.
Cc @jtracey93

AlexanderSehr: The reasoning is that it's a bad practice, gives a bad example and in general is not aligned with the SFI. Mind you that folks have already been after us for public keys that were only used for testing. And truth be told I'm a bit suprised GitHub did not block non-password-protected certificates from being uploaded to source control.
Anyhow, if you can convince the other maintainers, be my guest - but as far as I'm concerned this should not be done this way and it would be preferred to align with similar scenarios implemented in other modules.

kpantos: Well it depends, it is bad practice only when a production cert / secret is used not when a temporary self-signed one is used for testing an environment which gets deleted anyway. It has nothing to do with the SFI as again the test certificate is not in the released module nor used anywhere but during the testing of the module for binding to ssl. This pattern exists in other SFI audited repositories without flagging it. Can you please point me to the module specification that refers to this requirement?

Having said that, I was under the impression that you were interested in growing the AVM list by fostering and enabling the development of modules, but it looks like that instead you are introducing more blockers, bugs, policies, and checks focusing on over-engineering them. 

So, going forward, I'm going to implement this last recommendation but as my time is valuable, I'm going to be seriously considering whether it is worth investing more in the continuous development for the AVM program. After all we can always fall back to our original plan of having our own CSU LZA module registry which we can control.

cc: @jtracey93 @matebarabas

rahalan: Sorry, but I don't understand the discussion here. @kpantos under no circumstances are we should store a secret in source control.

kpantos: @rahalan thanks but not adding any value to the conversation especially as you don't understand it.

Closing the pull request.

ReneHezser: We are generating certificates for other modules on the fly with the Set-PfxCertificateInKeyVault.ps1
Could that be an option for you?

kpantos: Thanks Rene, but the challenge is not with how to implement this - I have actually already done so. 
The real challenge is that in order to implement this in this pattern module these additional services have to be deployed: 
- A new subnet with container instance delegation
- A ZRS Storage account 
- A private endpoint for the storage account 
- A private DNS zone for the storage account
- A deployment script with vnet integration (container instance)
- A managed identity with write access to the keyvault
And all these... three times (one for each test performed)
So, besides the increased testing time (which is already too long) there's also a cost associated with this, which module users will have to incur just for a demo certificate.
This module is not a simple resource module, it's probably the largest one currently in the registry and adding a bunch of resources just to generate a simple demo cert for SSL binding increases the maintenance surface beyond the initial intended scope.

PR #4037, opened 2024-12-20
feat: Allow disabling audit settings in 'avm/res/sql/server'

### LINKED ISSUE #2307: `[AVM Module Issue]:  SQL Server auditing support for user assigned managed identities`

### Issue Type?

Feature Request

### Module Name

avm/res/sql/server

### (Optional) Module Version

_No response_

### Description

The current version of the module does not support user assigned managed identities for SQL Auditing. 

https://github.com/Azure/bicep-registry-modules/blob/305f9a5684f60817df4b93b63eb6b3f942d38e77/avm/res/sql/server/audit-settings/modules/nested_storageRoleAssignment.bicep#L18

Please update the module to different principal types. 

### (Optional) Correlation Id

_No response_

### LINKED ISSUE #3888: [AVM Module Issue]: can no longer provision a sqlserver without audit settings always being configured

### Issue Type?

Bug

### Module Name

avm/res/sql/server

### (Optional) Module Version

latest (0.11.1)

### Description

As of this commit: https://github.com/Azure/bicep-registry-modules/commit/d26fc144f5df0652b0cb954a7b05ad7c81648df8 to `avm/res/sql/server`, it's no longer possible to provision an sqlserver _without_ auditing settings also being configured.

The best you can do is to provision with audit settings in a disabled state.

Take for example: NOT supplying `auditSettings` or using `auditSettings: null` in the module's params list results in auditing settings being configured with a default state of enabled

The only option is to provision the server with auditing configured, but in with a state of disabled. EG `auditSettings: { state: 'Disabled' }`

### (Optional) Correlation Id

_No response_

peterbud:
Also make user assigned identities work for audit storage.

Fixes #3888
Fixes #2307

## Pipeline Reference

| Pipeline |
| -------- |
|       [avm.res.sql.server](https://github.com/peterbud/bicep-registry-modules/actions/workflows/avm.res.sql.server.yml)   |

## Type of Change

  - [ ] Update to documentation

## Checklist

### DIFF
--- a/avm/res/sql/server/main.bicep
+++ b/avm/res/sql/server/main.bicep
@@ -126,6 +126,9 @@ param vulnerabilityAssessmentsObj vulnerabilityAssessmentType?
 @description('Optional. The audit settings configuration.')
 param auditSettings auditSettingsType = {} //Use the defaults from the child module
 
+@description('Optional. Set to false to disable audit settings completely. Default is true to enable audit settings.')
+param enableAuditSettings bool = true
+
 @description('Optional. Key vault reference and secret settings for the module\'s secrets export.')
 param secretsExportConfiguration secretsExportConfigurationType?
 
@@ -472,7 +475,7 @@ module server_encryptionProtector 'encryption-protector/main.bicep' = if (encryp
   ]
 }
 
-module server_audit_settings 'audit-settings/main.bicep' = if (auditSettings != null) {
+module server_audit_settings 'audit-settings/main.bicep' = if (auditSettings != null && enableAuditSettings) {
   name: '${uniqueString(deployment().name, location)}-Sql-AuditSettings'
   params: {
     serverName: server.name
--- a/avm/res/sql/server/main.bicep
+++ b/avm/res/sql/server/main.bicep
@@ -123,11 +123,10 @@ param encryptionProtectorObj encryptionProtectorType?
 @description('Optional. The vulnerability assessment configuration.')
 param vulnerabilityAssessmentsObj vulnerabilityAssessmentType?
 
-@description('Optional. The audit settings configuration.')
-param auditSettings auditSettingsType = {} //Use the defaults from the child module
-
-@description('Optional. Set to false to disable audit settings completely. Default is true to enable audit settings.')
-param enableAuditSettings bool = true
+@description('Optional. The audit settings configuration. If you want to disable auditing, set the parmaeter to an empty object.')
+param auditSettings auditSettingsType = {
+  state: 'Enabled'
+}
 
 @description('Optional. Key vault reference and secret settings for the module\'s secrets export.')
 param secretsExportConfiguration secretsExportConfigurationType?
@@ -475,7 +474,7 @@ module server_encryptionProtector 'encryption-protector/main.bicep' = if (encryp
   ]
 }
 
-module server_audit_settings 'audit-settings/main.bicep' = if (auditSettings != null && enableAuditSettings) {
+module server_audit_settings 'audit-settings/main.bicep' = if (!empty(auditSettings)) {
   name: '${uniqueString(deployment().name, location)}-Sql-AuditSettings'
   params: {
     serverName: server.name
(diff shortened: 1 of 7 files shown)


avm/res/sql/server/main.bicep
```diff
@@ -472,7 +475,7 @@ module server_encryptionProtector 'encryption-protector/main.bicep' = if (encryp
   ]
 }
 
-module server_audit_settings 'audit-settings/main.bicep' = if (auditSettings != null) {
+module server_audit_settings 'audit-settings/main.bicep' = if (auditSettings != null && enableAuditSettings) {
```
AlexanderSehr, line 478: I'm not sure I fully grasp the logic here. 
- If the user wants to deploy audit settings, he/she would set `enableAuditSettings` to `true`
- If the user does not want to deploy audit settings, he/she would set `enableAuditSettings` to `false`
- If the user originally enabled the audit settings but wants to disable them now, he/she would need to set `enableAuditSettings` to `true` and also the `state` to 'Disabled'? This seems rather counter intuitive.

Would it not e better to solve it like this:
- Change the `auditSettings` parameter to `param auditSettings auditSettingsType?`
- Change the condition here to `module server_audit_settings 'audit-settings/main.bicep' = if (!empty(auditSettings)) {`

This would mean, if the user does not specify the audit settings, they're not applied. However, if he/she does they would by default be enabled (via the default value of the state) and if one would then want to disable it again, the state would just need to be changed to `disabled`.

peterbud: The requirements are the following:
1. if the user does not specify anything, a default audit settings should be deployed
2. if the user specifies any specific audit settings, that will be deployed 
3. if the user want, he/she should be able to disable the deployment of audit settings completely

The current behavior of the modules satisfies 1 and 2, but not 3.
Your proposed changes would satisfy 2 and 3, but not 1.
With this PR all three requirements could be satisfied.

BTW this PR does something similar like the **telemetry** module: by default if deploys the telemetry, but the user can opt-out.

If you look at the original #3888, the problem is that having a deployed `auditSettings` with disabled state is **NOT** the same as not having any `auditSettings`. Functionally they are similar, but not the same.

I hope this makes sense, let me know your POV.

AlexanderSehr: Hey @peterbud,
thanks for the write-up. I'm just concerned that it would be very counterintuitive to disable the setting itself (as an 'enabled' needs to be specified to set a 'disabled'). 

To Satisfy #1 it would be enourgh to change the default to
```bicep
param auditSettings auditSettingsType = {
  state = 'Enabled'
}
```
To avoid the setting alltogether a user would need to either set `auditSettings` to `nulll` or `{}`.

The angle where I'm coming from is that this is the usual pattern in AVM, that is, that child modules either have an object parameter, or array parameter in their parent, in the child itself is deployed if the parameter is not null or empty. 
What **is** a bit special here is requriement #1, but also not a unicorn as e.g. implemented in the VM module with an extra `enabled` flag (as a `state` property is not available for VM extensions where this was used) 

https://github.com/Azure/bicep-registry-modules/blob/f6527fa82f9161f2fb156691fc08be26a06d39e4/avm/res/compute/virtual-machine/main.bicep#L175-L178

https://github.com/Azure/bicep-registry-modules/blob/f6527fa82f9161f2fb156691fc08be26a06d39e4/avm/res/compute/virtual-machine/main.bicep#L162-L167

https://github.com/Azure/bicep-registry-modules/blob/f6527fa82f9161f2fb156691fc08be26a06d39e4/avm/res/compute/virtual-machine/main.bicep#L153-L155

https://github.com/Azure/bicep-registry-modules/blob/f6527fa82f9161f2fb156691fc08be26a06d39e4/avm/res/compute/virtual-machine/main.bicep#L696

Ultimately, I'll leave this up to you as the module owner. It's just a suggestion from my end :)

peterbud: Hi @AlexanderSehr, thanks for the quick reply. TIL: passing `null` and passing `{}` is NOT the same from the perspective of the default values!

I have tried earlier the method you proposed, but when I have tested with passing in `null`, the default value was used and I could not disable the deployment of the audit settings module. So I tried to use a different way. But today I have tested and realized, that passing in an empty object param `{}` actually works!

Let me update quickly the code

peterbud: OK, pls have a look, I have refactored it.

AlexanderSehr: Dang, ok. Thanks for testing and aligning it with the VM module. Will approve

AlexanderSehr (approved):

PR #4096, opened 2024-12-29
feat: Updated Network Manager module to support Routing Configuration feature

ahmadabdalla:
## Description

Azure Virtual Network Manager (AVNM) recently introduced a new feature that allows users to configure routing settings for their virtual networks. See [concept-user-defined-route](https://learn.microsoft.com/en-us/azure/virtual-network-manager/concept-user-defined-route) documentation.

ℹ️ Routing Configuration is Generally Available in many Azure regions: See [UDR Management - General Availability](https://learn.microsoft.com/en-us/azure/virtual-network-manager/concept-user-defined-route#general-availability). Given the optional usage of this feature, This should not impact existing usage of AVNM.

While enabling this feature, the following had to be updated on the module:
- Added new child modules for `routing-configuration`, `rule-collection` and `rule`.
- Updated the API to `2024-05-01` as per (https://learn.microsoft.com/en-us/azure/templates/microsoft.network/networkmanagers?pivots=deployment-language-bicep)
- Updated module version.
- Use of the common types for `roleAssignments` and `locks`.
- Update the scope access to become optional as AVNM now enables a `None` feature mode to support capabilities like IPAM and Virtual Network Verifier.
- Updated built in role name from `IPAM Pool Contributor` to `IPAM Pool User`, Still keeps the same GUID.
- Added support for `memberType` for Subnets for Network Groups to support the new routing feature. However, set the default to virtual network as it is the common scenario.
- Updated the default test case to not use any features.
- Expanded on the max test case coverage to include the routing feature and added different network groups of subnet types.
- Expanded on usage of the `import` functionality to reduce code duplication and import types from child module to parents.
- Updated the `description` parameter to default to `''` instead of `null`. Setting this to `null` causes an error 400 on the API.

## Pipeline Reference

| Pipeline |
| -------- |
|   [avm.res.network.network-manager](https://github.com/ahmadabdalla/bicep-registry-modules/actions/workflows/avm.res.network.network-manager.yml)    |

## Type of Change

  - [ ] Update to documentation

## Checklist


avm/res/network/network-manager/main.bicep
```diff
@@ -214,58 +233,22 @@ output location string = networkManager.location
...
 
-    @sys.description('Required. Resource ID of the virtual network.')
-    resourceId: string
-  }[]?
+  @sys.description('Optional. Static Members to create for the network group. Contains virtual networks to add to the network group.')
+  staticMembers: staticMembersType?
 }[]?
 
-type networkManagerScopeAccessType = ('Connectivity' | 'SecurityAdmin')[]
+type networkManagerScopeAccessType = ('Connectivity' | 'SecurityAdmin' | 'Routing')[]?
```
AlexanderSehr, line 251:
```suggestion
type networkManagerScopeAccessType = ('Connectivity' | 'SecurityAdmin' | 'Routing')
```
As per soon to be released specs, both the `array` and `nullable` indicators should be set at the parameter, not the UDT-side. The reasoning is simply to make it easier to understand a parameter/property by looking at it.

So this change should go along with updating the parameter to
```bicep
networkManagerScopeAccesses networkManagerScopeAccessType[]?
```

AlexanderSehr: Note: The same idea applies to all other types

avm/res/network/network-manager/main.bicep
```diff
@@ -171,6 +177,19 @@ module networkManager_securityAdminConfigurations 'security-admin-configuration/
   }
 ]
 
+module networkManager_routingConfigurations 'routing-configuration/main.bicep' = [
+  for (routingConfiguration, index) in routingConfigurations ?? []: {
+    name: '${uniqueString(deployment().name, location)}-NetworkManager-RoutingConfigurations-${index}'
+    params: {
+      name: routingConfiguration.name
+      networkManagerName: networkManager.name
+      description: routingConfiguration.?description ?? ''
```
AlexanderSehr, line 186: Are the fallback defaults here intended? If not provided for a nullable parameter they just won't be used. In other words: I'd only recommend setting a fallback default here if you definitely want to not pass `null`.

avm/res/network/network-manager/routing-configuration/main.bicep
```diff
@@ -0,0 +1,77 @@
...
+module routingConfigurations_ruleCollections 'rule-collection/main.bicep' = [
+  for (ruleCollection, index) in ruleCollections ?? []: {
+    name: '${uniqueString(deployment().name)}-RoutingConfigurations-RuleCollections-${index}'
+    params: {
+      networkManagerName: networkManager.name
+      routingConfigurationName: routingConfigurations.name
+      name: ruleCollection.name
+      description: ruleCollection.?description ?? ''
+      appliesTo: ruleCollection.appliesTo
+      disableBgpRoutePropagation: ruleCollection.?disableBgpRoutePropagation ?? true
```
AlexanderSehr, line 41:
```suggestion
      disableBgpRoutePropagation: ruleCollection.?disableBgpRoutePropagation
```
Redundant as the deployment will default to the parameters default (`true`) regardless if not provided.

AlexanderSehr (commented):
Great work @ahmadabdalla, as always. Really only have an ask regarding the declaration of `arrays` & `nullable` for UDTs if you can make room for it :)

avm/res/network/network-manager/main.bicep
ahmadabdalla: ouch. Let me look into that. Thanks @AlexanderSehr

avm/res/network/network-manager/main.bicep
ahmadabdalla: I've set this as a default value on the param directly instead and allowed nullable here.

avm/res/network/network-manager/main.bicep
ahmadabdalla: @AlexanderSehr this does require a lot of updates to the module. May I suggest we roll in this new feature please as its a feature request from internal MS eng. Then I can dedicate another PR for this new spec update?

AlexanderSehr: Sure thing @ahmadabdalla. Just wanted to mention it as the spec (and test) will come eventually :) Will resolve

avm/res/network/network-manager/main.bicep
AlexanderSehr: Neat, thanks. I think this will make it a bit clearer (also because now the user can see the default in the readme). Much appreciated, thanks 💪

AlexanderSehr (approved):

PR #4286, opened 2025-01-24
feat: Update `avm/res/web/static-site` - Add additional param for `publicNetworkAccess`

### LINKED ISSUE #4283: [AVM Module Issue]: Static Web App missing `publicNetworkAccess` parameter due to outdated API

### Issue Type?

Security Bug

### Module Name

avm/res/web/static-site

### (Optional) Module Version

_No response_

### Description

Due to the usage of an old API ([2021-03-01](https://github.com/Azure/bicep-registry-modules/blob/9a0b5afb7f5b2016ffa3b631c17aa83472cfb260/avm/res/web/static-site/main.bicep#L158)), this module is lacking the ability to disable public access to the Static WebApp. 

This parameter is available in any of the next versions, such as [2022-03-01](https://learn.microsoft.com/en-us/azure/templates/microsoft.web/2022-03-01/staticsites?pivots=deployment-language-bicep).

Is it possible to update the API used?

### (Optional) Correlation Id

_No response_

ChrisSidebotham:
## Description

#Closes #4283

## Pipeline Reference

| Pipeline |
| -------- |
|     [avm.res.web.static-site](https://github.com/ChrisSidebotham/bicep-registry-modules/actions/workflows/avm.res.web.static-site.yml)     |

## Type of Change

  - [ ] Update to documentation

## Checklist


avm/res/web/static-site/main.bicep
```diff
@@ -88,6 +88,9 @@ param functionAppSettings object = {}
 @description('Optional. The custom domains associated with this static site. The deployment will fail as long as the validation records are not present.')
 param customDomains array = []
 
+@description('Optional. The public network access settings for the static site. `Disabled` is configured by default.')
+param publicNetworkAccess string = 'Disabled'
```
AlexanderSehr, line 92:
```suggestion
@allowed([
  ''
  'Enabled'
  'Disabled'
])
param publicNetworkAccess string = 'Disabled'
```

Requires a regeneration

avm/res/web/static-site/main.bicep
```diff
@@ -88,6 +88,9 @@ param functionAppSettings object = {}
 @description('Optional. The custom domains associated with this static site. The deployment will fail as long as the validation records are not present.')
 param customDomains array = []
 
+@description('Optional. The public network access settings for the static site. `Disabled` is configured by default.')
```
eriqua, line 91: Leveraging the PR to trigger a discussion about this setting in general and current implementation across the library. 
I see the inconsistency is strong with this one. 
In CARML we started disabling publicNetworkAccess within the resource implementation instead of through its default value, and only in case private endpoint is enabled. This is still the current implementation for multiple modules, such as for example [storage-account](https://github.com/Azure/bicep-registry-modules/blob/main/avm/res/storage/storage-account/main.bicep#L444)
The reasoning behind was that it won't be useful to disable public network access by default if private endpoint is not configured. In these cases the default value is set to empty `''`.
Meanwhile, many other resource modules are now disabling publicNetworkAccess by default via default values in AVM. An example is [RSV](https://github.com/Azure/bicep-registry-modules/blob/main/avm/res/recovery-services/vault/main.bicep#L63).  In these cases the default value is set to `Disabled`.
Needless to say, we also have cases where the default value is set to `Enabled`.
I think it would be nice to align the behavior.

AlexanderSehr: I could imagine that some resources may have PSRule tests that request one or the other. Don't have an example at hand though (for RSV, for example, this test does not exist, so I guess the person implementing the parameter simply decided that it would be a good solution).
One added caveat is that aside from PEs, one could also navigate to a resource like a storage account using service endpoints, making the matter even harder to define a default for.

ChrisSidebotham: I am about t commit another push to this which updates it to the same configuration as we use for PaaS Services, but I am removing `networkacl` as this is not supportted for this resource

avm/res/web/static-site/main.bicep
ChrisSidebotham: applied

avm/res/web/static-site/main.bicep
```diff
@@ -88,6 +88,14 @@ param functionAppSettings object = {}
 @description('Optional. The custom domains associated with this static site. The deployment will fail as long as the validation records are not present.')
 param customDomains array = []
 
+@description('Optional. Whether or not public network access is allowed for this resource. For security reasons it should be disabled. If not specified, it will be disabled by default if private endpoints are set and networkAcls are not set.')
```
eriqua, line 91:
```suggestion
@description('Optional. Whether or not public network access is allowed for this resource. For security reasons it should be disabled. If not specified, it will be disabled by default if private endpoints are set.')
```

eriqua:
@ChrisSidebotham ping 🤌

avm/res/web/static-site/main.bicep
ChrisSidebotham: Updated and pushed - please review and resolve @eriqua - Apologies I have been on leave

eriqua (approved):
🆗

Issue #4344, opened 2025-01-29
[AVM Module Issue]: IPPrefix error while deploying to regions that don't support availability zones (ie. - australiacentral)


juancj:
### Check for previous/existing GitHub issues

- [x] I have checked for previous/existing GitHub issues

### Issue Type?

Bug

### Module Name

avm/res/network/public-ip-prefix

### Description

When trying to deploy IPPrefix to a region that doesn't support availability zones I get the following error:
 
The resource 'Microsoft.Network/publicIPAddresses/TSE-PROD-AUC-AEGIS-PIP-01' does not support availability zones at location 'australiacentral'

ericscheffler:
Linking propsed PR #4356

juancj:
Passing the parameter as empty allows deploying the module to regions that don't support availability zones. Closing the issue.

krbar:
@juancj thanks for the feedback. Closing the issue.

PR #4356, opened 2025-01-30
feat: Public IP Prefix - Removing default for `zones` to support more regions by default

juancj:
## Description
 Removed the default values from the zones declaration in the `avm.res.network.public-ip-prefix` module. The change now allows to deploy PIP Prefixes in regions that don't support availability zones. Refer to issue #4344 for more details

## Pipeline Reference

[avm.res.network.public-ip-prefix](https://github.com/juancj/bicep-registry-modules/actions/workflows/avm.res.network.public-ip-prefix.yml)

## Type of Change

  - [ ] Update to documentation

## Checklist

### DIFF
--- a/avm/res/network/public-ip-prefix/main.bicep
+++ b/avm/res/network/public-ip-prefix/main.bicep
@@ -50,11 +50,7 @@ param ipTags ipTagType[]?
   2
   3
 ])
-param zones int[] = [
-  1
-  2
-  3
-]
+param zones int[] = []
 
 @description('Optional. Enable/Disable usage telemetry for module.')
 param enableTelemetry bool = true
--- a/avm/res/network/public-ip-prefix/README.md
+++ b/avm/res/network/public-ip-prefix/README.md
@@ -771,14 +771,7 @@ A list of availability zones denoting the IP allocated for the resource needs to
 
 - Required: No
 - Type: array
-- Default:
-  ```Bicep
-  [
-    1
-    2
-    3
-  ]
-  ```
+- Default: `[]`
 - Allowed:
   ```Bicep
   [
--- a/avm/res/network/public-ip-prefix/main.json
+++ b/avm/res/network/public-ip-prefix/main.json
@@ -5,8 +5,8 @@
   "metadata": {
     "_generator": {
       "name": "bicep",
-      "version": "0.32.4.45862",
-      "templateHash": "1837206327927302660"
+      "version": "0.33.13.18514",
+      "templateHash": "7290386318086898070"
     },
     "name": "Public IP Prefixes",
     "description": "This module deploys a Public IP Prefix."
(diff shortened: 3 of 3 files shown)


avm/res/network/public-ip-prefix/main.bicep
```diff
@@ -50,11 +50,7 @@ param ipTags ipTagType[]?
   2
   3
 ])
-param zones int[] = [
-  1
-  2
-  3
-]
+param zones int[] = []
```
AlexanderSehr, line 53: This unfortunately goes against the AVM spec [SFR5](https://azure.github.io/Azure-Verified-Modules/specs/bcp/res/#:~:text=ID%3A%20SFR5%20%2D%20Category%3A%20Composition%20%2D%20Availability%20Zones)

Instead of removing the default, would a workaround for unsupported regions be to just provide an empty array as a parameter, like in your suggested update?

cc: @jtracey93

juancj: Hi Alexander, 
Just tested it sending the array param as empty and that seems to work fine. Thanks for the help.

ericscheffler: @AlexanderSehr Roger that; this does create somewhat of an issue however where Azure regions just don't have availability zones, where customers will get errors they won't see until deployment time. It felt kind of weird to me to submit an empty array, but if that's the standard, so be it :-) Thanks for the feedback!

AlexanderSehr: At the end of the day, both approaches work and have their merits (i.e., I fully understand where you're both coming from). 
While I was not a core part of the conversation that lead to this spec, I understand that deploying a resource, by default, as zonal (even if it may not work in some regions) was considered 'better' as it's more WAF-aligned and can be worked around relatively easily (be it a bit akward, as you highlighted @ericscheffler). 

@jtracey93, maybe you want to also add your five cents if you have a moment :)

If the elaboration makes sense I'd recommend to close the PR, but before we do, let's wait for that reponse as we can refer to it again later :)

jtracey93: Yup the decision was made with engineering to default to Zone-redundant configs in all modules to align to the reliability pillar in WAF but also to provide flexibility to override via the parameter inputs, as discussed here.

You can see more on this in the AVM FAQ here: https://azure.github.io/Azure-Verified-Modules/resources/faq/#what-does-avm-mean-by-waf-aligned

ericscheffler: Fair enough 👍 ... if that's the ruling on the field I'm fine with that, thanks for the clarification.

AlexanderSehr: Thanks @ericscheffler :) Always good to discuss anything that may seem odd. 
@juancj, what's your take? Can we close the PR? :)

PR #4427, opened 2025-02-11
feat: APIM - Removed workaround for secrets

### LINKED ISSUE #2070: [AVM Question/Feedback]: replace `secret` property with custom types

### Description

Some modules that are using a property `secureList` within the secure parameter. An example is the container-app-job module here: https://github.com/Azure/bicep-registry-modules/blob/7f9c0893518f8df23c72733d9c8b64660c226a74/avm/res/app/job/tests/e2e/max/main.test.bicep#L74

This looks like a workaround which was needed in the past. We can now declare custom types and decorate a property with the `@secure()` tag to protect its value, instead of the whole security object.
Something like this (I skipped the descriptions for readability):

```bicep
type secretsType = {
  identity: string

  keyVaultUrl: string

  name: string

  @secure()
  value: string
}[]
```

> **This can be a breaking change, as the property for the module is then changed!**

I see we have 5 modules that use a secureList like this and suggest they introduce a custom type instead of the secureList workaround.

- [ ] api-management/service
- [ ] app/container-app
- [ ] app/job
- [ ] resources/deployment-script
- [ ] storage/storage-account/local-user
- [x] key-vault/vault

AlexanderSehr:
## Description

Replaced workaround of secureList with UDT

## Pipeline Reference

| Pipeline |
| -------- |
|     [avm.res.api-management.service](https://github.com/Azure/bicep-registry-modules/actions/workflows/avm.res.api-management.service.yml)     |

## Type of Change

  - [ ] Update to documentation


AlexanderSehr:
Hey @tony-box,
do you have capacity to review this PR? Otherwise I'd reach out to the @Azure/avm-module-reviewers-bicep team for support

tony-box (approved):
@AlexanderSehr looks good, thank you!

PR #4469, opened 2025-02-18
fix: Update `avm/res/net-app/net-app-account` - fix optional zone for volumes and specify `protocolTypes` options

thecmdradama:
## Description
This PR resolves the following issues in the `avm/res/net-app/net-app-account` module
* On the NetApp Volume submodule, `zone` has been renamed to `availabilityZone`.  
* The type for `availabilityZone` has been changed to `string` from `int[]` and is now optional with null as the default. Previously the module would try and deploy a volume into all three zones if no value is defined.
* Changed: `protocolTypes` is now an optional array of strings with three possible options `'NFSv3', 'NFSv4.1', and 'CIFS'`. If no value is provided, the default value of `['NFSv3']` is used.
* Added to the description for `protocolTypes` indicating that the user will need to specify `['NFSv3','CIFS']` or `['NFSv4.1','CIFS']` if creating dual-stack volumes.
* Added in additional submodule README.md and main.json files generated by Set-AVMModule.

## Pipeline Reference

| Pipeline |
| -------- |
|    [avm.res.net-app.net-app-account](https://github.com/thecmdradama/bicep-registry-modules/actions/workflows/avm.res.net-app.net-app-account.yml)    |

## Type of Change

  - [x] Update to documentation

## Checklist


avm/res/net-app/net-app-account/capacity-pool/volume/main.bicep
```diff
@@ -32,7 +32,12 @@ param keyVaultPrivateEndpointResourceId string?
 param volumeType string?
 
 @description('Optional. Zone where the volume will be placed.')
-param zones int[] = [1, 2, 3]
+@allowed([
+  '1'
+  '2'
+  '3'
+])
+param availabilityZone string?
```
AlexanderSehr, line 40: Please undo. This makes this interface non compliant to the AVM specs. In this case
- [ID: SFR5 - Category: Composition - Availability Zones
](https://azure.github.io/Azure-Verified-Modules/specs/bcp/res/#:~:text=ID%3A%20SFR5%20%2D%20Category%3A%20Composition%20%2D%20Availability%20Zones) and
- [ID: BCPNFR17 - Category: Composition - Code Styling - Type casting](https://azure.github.io/Azure-Verified-Modules/specs/bcp/res/#:~:text=ID%3A%20BCPNFR17%20%2D%20Category%3A%20Composition%20%2D%20Code%20Styling%20%2D%20Type%20casting)

Changing it to a string (which is unfortuantely how the API expects it) makes the module not only harder to use, but also allows you to specify only one zone.

Now, this is where it gets a bit more interesting. Even though the API calls for 'zones` (plural) one could argue it can only be deployed into one zone as it's not e.g., a scale set or similar. This would explain also you change to a singular value.
For these cases however, the specs depand to make the parameter still an int, but a mandatory value as we cannot assume in which zone the resource should be deployed into. However, you're correct in the user must be able to not deploy into any zone add all. One approach that could be chosen is the one implemented by the disks module [here](https://github.com/Azure/bicep-registry-modules/blob/57e651e9b40b95cd8656441cc3bb626b4a0dfba6/avm/res/compute/disk/main.bicep#L126-L133)

thecmdradama: Thanks @AlexanderSehr That makes perfect sense. Yeah it is a little frustrating but I get the point for changing it back to an int from a string.

AlexanderSehr: Much appreciated, thanks @thecmdradama.
While looking into this, I reviewed some other modules and found a variety of implementations (sometimes with very provider-specific values). Anyhow - to align the approaches entirely, I defined now an [interface](https://github.com/Azure/Azure-Verified-Modules/pull/1870) that we will discuss in the core team soon.

AlexanderSehr:
Hey @fbinotto,
please review at your earliest convenience :)

thecmdradama:
Just had a thought around another change I would like to make. Being that the capacity pool size must be set in 1TiB chunks. Is there any value in defining a user defined function to handle the Tibibyte to byte conversion? Would make consuming the module considerably easier. Thoughts @AlexanderSehr and @fbinotto?

AlexanderSehr:
> Just had a thought around another change I would like to make. Being that the capacity pool size must be set in 1TiB chunks. Is there any value in a user defined function to handle the Tebibyte to byte conversion? Would make consuming the module considerably easier. Thoughts @AlexanderSehr and @fbinotto?

Hey @thecmdradama,
not to my knowledge and also none is to be found when looking at the [available Bicep functions](https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/bicep-functions).
However, it is possible to [define functions](https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/user-defined-functions) that could do this on the user's behalf.

thecmdradama:
> Hey @thecmdradama, not to my knowledge and also none is to be found when looking at the [available Bicep functions](https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/bicep-functions). However, it is possible to [define functions](https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/user-defined-functions) that could do this on the user's behalf.

Thanks @AlexanderSehr, a user defined function was what I was thinking of implementing to then simplify the capacity pool size parameter. 
From what I can see, there's only one pattern (subResourceWrapper) and no resources that currently implement UDF's. So not really anything currently to try and create a standard from.

I've thrown a quick example together on how it could look, in my last commit. Maybe a topic of discussion for the AVM Core team on an agreed approach for utilising UDF's in modules and patterns?

Thanks,

AlexanderSehr:
> > Hey @thecmdradama, not to my knowledge and also none is to be found when looking at the [available Bicep functions](https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/bicep-functions). However, it is possible to [define functions](https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/user-defined-functions) that could do this on the user's behalf.
> 
> Thanks @AlexanderSehr, a user defined function was what I was thinking of implementing to then simplify the capacity pool size parameter. From what I can see, there's only one pattern (subResourceWrapper) and no resources that currently implement UDF's. So not really anything currently to try and create a standard from.
> 
> I've thrown a quick example together on how it could look, in my last commit. Maybe a topic of discussion for the AVM Core team on an agreed approach for utilising UDF's in modules and patterns?
> 
> Thanks,

ah well, no standard means no restrictions 😉 
My only thought was that if there were multiple such functions it might make sense to publish them as a `utl` (utility) module. But for just one function, this would be a bit of an overkill.
Looking at your commit, I'd argue it looks great 💪

fbinotto (approved):
LGTM

PR #4693, opened 2025-03-07
fix: `avm/res/web/static-site` fixes for pdns & Identity Var

### LINKED ISSUE #4345: [AVM Module Issue]: Unknown (privatelink) DNS zone upon Static Site creation.

### Issue Type?

Feature Request

### Module Name

avm/res/web/static-site

### (Optional) Module Version

_No response_

### Description

At present, Static WebApps are deployed under a random generated site identifier and the `azurestaticapps.net` suffix. This suffix can contain a partition ID, and you can have a URL like `{identifier}.azurestaticapps.net` or `{identifier}.{partitionId}.azurestaticapps.net`. This is documented in both [Private Endpoint documentation](https://github.com/MicrosoftDocs/azure-docs/blob/228ddc7b1231993d31fd13fd343ecff514c48ea4/articles/private-link/private-endpoint-dns.md?plain=1#L213) as well as in the [Static Web App documentation](https://github.com/MicrosoftDocs/azure-docs/blob/228ddc7b1231993d31fd13fd343ecff514c48ea4/articles/static-web-apps/private-endpoint.md?plain=1#L69) on using private endpoints.

The privatelink domains that you need to create a DNS record in, als need to provide this partition id. So, if you app is located on `{identifier}.4.azurestaticapps.net` and you want to create a private endpoint, the corresponding A record must be created in the `privatelink.4.azurestaticapps.net` DNS zone. 

The name of the private DNS zone is part of its `resourceId`. This `resourceId` is input of the private endpoint parameters of this module:

```bicep
module staticSite 'br/public:avm/res/web/static-site:<version>' = {
  name: 'staticSiteDeployment'
  params: {
    name: 'mysite'
    privateEndpoints: [
      {
        privateDnsZoneGroup: {
          privateDnsZoneGroupConfigs: [
            {
              privateDnsZoneResourceId: '<privateDnsZoneResourceId>'
            }
          ]
        }
        subnetResourceId: '<subnetResourceId>'
      }
    ] 
}
```
The `privateDnsZoneResourceId` can't be known. There are two solutions to this problem:

1. Deploy twice: first without a private endpoint, see what domain partition id you are dealt, and add the endpoint with the correct `privateDnsZoneResourceId` later.
2. Create a module that defines the static site resource and private endpoint separately, and use the output of the former to construct the `privateDnsZoneResourceId` later. 

The second approach is (in my mind) generalizable across users.

The feature request is for the module to be able to amend the `privateDnsZoneResourceId` so it fixes the actual resource part. 
So when one provided the following `privateDnsZoneResourceId`:
```text
/subscriptions/{guid}/resourceGroups/{rg_name}/providers/Microsoft.Network/privateDnsZones/privatelink.azurestaticapps.net
```
the module could automatically amend the suffix by adding the `{partitionId}`. This information is retrievable by getting the `DefaultHost` property of the staticSite resource and replace the identifier with `privatelink`. For example:

```bicep
resource staticSite 'Microsoft.Web/staticSites@2021-03-01' = {
  name: name
  location: location
  sku: {
    name: sku
    tier: sku
  }
  properties: {
    // some properties
  }

var identifier = split(staticSite.properties.defaultHostname, '.')[0]
var privateLinkZone = replace(staticSite.properties.defaultHostname, identifier, 'privatelink')
```

Then the variable `privateLinkZone` can be used and the private endpoint is correctly created.

### Alternatives
Amending parameters in a nontransparent manner feels iffy. As alternative there could also be an additional boolean parameter on the `privateEndpoints` parameter that would steer this behavior. 

I checked other implementations as well:
- Aure Data Explorer, AKS, Batch, Container Apps (and probably others) have variable domain names, but only have `{region}` as part of the domain zone, which is predictable at deployment time.
- SQL Managed Instance do have unpredictable elements in the domain zone for privatelink, but it is unclear how you can create a private endpoint through Bicep.

It seems that Azure Static WebApps is unique in this situation. 

### (Optional) Correlation Id

_No response_

### LINKED ISSUE #4701: [AVM Module Issue]: var identity is constructed incorrectly on several templates

### Issue Type?

Bug

### Module Name

avm/res/web/static-site

### (Optional) Module Version

_No response_

### Description

The issue is opened for Static Web App because that is where I have noticed and tested it, but the same bug seems to exist in multiple templates:
- avm/res/app-configuration/configuration-store
- avm/res/compute/dis-encryption-set
- avm/res/document-db/database-account
- avm/res/machine-learning-services/workspace/compute
- avm/res/search/search-service
- avm/res/signal-r-service/signal-r
- avm/res/sql/managed-instance
- avm/res/sql/server
- avm/res/web/hosting-environment - ([Microsoft.Web/hostingEnvironments](https://learn.microsoft.com/en-us/azure/templates/microsoft.web/hostingenvironments?pivots=deployment-language-bicep) does not support the identity property, so should be removed I believe)
- avm/res/web/static-site

I would like to request the fix of all bugs in this request - hope that I do not need to open an issue for each module separately.

#### Describe the bug
When deploying the template with system-assigned managed identity explicitly set to false while not providing any user-assigned managed resource id, the deployment validation fails.

#### Error message 
New-AzResourceGroupDeployment: 10:09:05 - Error: Code=InvalidTemplate; Message=Deployment template parse failed: 'Required property 'type' expects a value but got null. Path ''.'.

The search/search-service gives a different error:
New-AzResourceGroupDeployment: 10:09:05 - Error: Code=InvalidTemplate; Message=Deployment template parse failed: 'Error converting value "" to type 'Azure.Deployments.Core.Definitions.Resources.ResourceIdentityType'. Path 'type'.'.

#### To Reproduce
Create deployment invoking the module as below. 
```bicep
module staticwebapp 'br/public:avm/res/web/static-site:0.8.2' = {
  name: 'deploy-static-webapp'
  params: {
    name: 'nmfstaticwebapp'
    managedIdentities: {
      systemAssigned: false
      userAssignedResourceIds: []
    }
  }
}
```

#### Root cause and suggested fix
The `identity` property accepts only the following values: `'None' | 'SystemAssigned' | 'SystemAssigned, UserAssigned' | 'UserAssigned'` .

The bicep code in mentioned modules sets the type to `null` or `''` instead of None at the following section:
```bicep
var identity = !empty(managedIdentities)
  ? {
      type: (managedIdentities.?systemAssigned ?? false)
        ? (!empty(managedIdentities.?userAssignedResourceIds ?? {}) ? 'SystemAssigned, UserAssigned' : 'SystemAssigned')
        : (!empty(managedIdentities.?userAssignedResourceIds ?? {}) ? 'UserAssigned' : null)
      userAssignedIdentities: !empty(formattedUserAssignedIdentities) ? formattedUserAssignedIdentities : null
    }
  : null
```

This can be adjusted to the below to resolve the bug (already implemented the same way in multiple templates, e.g. data-factory/factory).
```bicep
var identity = !empty(managedIdentities)
  ? {
      type: (managedIdentities.?systemAssigned ?? false)
        ? (!empty(managedIdentities.?userAssignedResourceIds ?? {}) ? 'SystemAssigned, UserAssigned' : 'SystemAssigned')
        : (!empty(managedIdentities.?userAssignedResourceIds ?? {}) ? 'UserAssigned' : 'None')
      userAssignedIdentities: !empty(formattedUserAssignedIdentities) ? formattedUserAssignedIdentities : null
    }
  : null
```

#### Additional context
This is not a blocking issue because the desired result can be achieved by simply leaving the the managedIdentities parameter out of the deployment. This might lower the priority to fix it, but should happen regardless.

### (Optional) Correlation Id

_No response_

ChrisSidebotham:
## Description

Closes #4345 
Closes #4701 

Configured DNS Record in test: 
image

New DNS Zone:
image

## Pipeline Reference

| Pipeline |
| -------- |
|    [avm.res.web.static-site](https://github.com/ChrisSidebotham/bicep-registry-modules/actions/workflows/avm.res.web.static-site.yml)      |

## Type of Change

  - [ ] Update to documentation

## Checklist


avm/res/web/static-site/main.bicep
```diff
@@ -96,6 +96,16 @@ param customDomains array = []
 ])
 param publicNetworkAccess string = ''
 
+@description('Condiitonal. Due the nature of Azure Static Apps, a partition ID is added to the app URL upon creation. Enabling the creation of the private DNS Zone will provision a DNS Zone with the correct partition ID, this is required for private endpoint connectivity to be enabled. You can choose to disable this option and create your own private DNS Zone by leveraging the output of the partitionId within this module. Default is `Enabled`, However the Private Dns Zone will only be created following if a `privateEndpoint` configuration is supplied.')
```
AlexanderSehr, line 99:
```suggestion
@description('Conditional. Due the nature of Azure Static Apps, a partition ID is added to the app URL upon creation. Enabling the creation of the private DNS Zone will provision a DNS Zone with the correct partition ID, this is required for private endpoint connectivity to be enabled. You can choose to disable this option and create your own private DNS Zone by leveraging the output of the partitionId within this module. Default is `Enabled`, However the Private DNS Zone will only be created following if a `privateEndpoint` configuration is supplied.')
```

avm/res/web/static-site/main.bicep
```diff
@@ -301,7 +319,23 @@ module staticSite_privateEndpoints 'br/public:avm/res/network/private-endpoint:0
         'Full'
       ).location
       lock: privateEndpoint.?lock ?? lock
-      privateDnsZoneGroup: privateEndpoint.?privateDnsZoneGroup
+      // privateDnsZoneGroup: privateEndpoint.?privateDnsZoneGroup
```
AlexanderSehr, line 322: This looks like the change is still tested. Should the PR be in draft?

avm/res/web/static-site/main.bicep
```diff
@@ -257,6 +267,14 @@ resource staticSite_roleAssignments 'Microsoft.Authorization/roleAssignments@202
   }
 ]
 
+module staticSite_privateDnsZone 'br/public:avm/res/network/private-dns-zone:0.7.0' = if (!empty(privateEndpoints) && createPrivateDnsZone == 'Enabled') {
```
AlexanderSehr, line 270: Considering the amount of properties one is able to configure in the module I wonder if we may need to introduce a parameterObject for them - like done e.g. for the `nicConfigurations` parameter in the VM module. 

Just thinking out loud here. If you cannot configure it to you needs you're essentially required to idempotently redeploy the DNS Zone after the deployment with another deployment. Maybe that's ok and intended - but wanted to raise the thought :)

avm/res/web/static-site/main.bicep
ChrisSidebotham: No, I just hadn't removed this, my bad

avm/res/web/static-site/main.bicep
ChrisSidebotham: So on this thought, for a VM the NIC is required, I guess here we do follow a smiliar pattern except the PDNS Zone is only needed for private endpoints to work. Typically the DNS Zone should sit with the rest of the DNS Zone and be configured with a vnet link.

In this current implemntation the app will still not be routable as the DNS Zone has not been linked to a Virtual Network, I need to update this so it accepts some properties. My view is this should be the bare minimal requirements to ensure PE connectivity is established and routable but any extensive configuration should be done in a dedicated module for the staticSite zone. il make an update to this today

avm/res/web/static-site/main.bicep
ChrisSidebotham: Updated, no longer relevant

avm/res/web/static-site/main.bicep
ChrisSidebotham: udpated

avm/res/web/static-site/main.bicep
ChrisSidebotham: Additional param added for virtualNetworkResourceId

avm/res/web/static-site/main.bicep
```diff
@@ -367,3 +401,83 @@ type privateEndpointOutputType = {
   @description('The IDs of the network interfaces associated with the private endpoint.')
   networkInterfaceResourceIds: string[]
 }
+
+// @description('An AVM-aligned type for a private endpoint. To be used if the private endpoint\'s default service / groupId can be assumed (i.e., for services that only have one Private Endpoint type like \'vault\' for key vault).')
+// type privateEndpointSingleServiceType = {
```
AlexanderSehr, line 406: Leftover?

avm/res/web/static-site/main.bicep
AlexanderSehr: I wonder if that's something we should chat about some time - just to ensure we're all on the same page. The next maintainer call would come to mind 💪

avm/res/web/static-site/main.bicep
ChrisSidebotham: removed

avm/res/web/static-site/main.bicep
ChrisSidebotham: can we let this go in the meantime?

AlexanderSehr: Define "Let go"? Should we not find together the best solution instead of potentially introducing multiple breaking changes? Hence the suggestion with the Maintainer call where these design questions would usually be brought up.

avm/res/web/static-site/main.bicep
```diff
@@ -301,7 +327,15 @@ module staticSite_privateEndpoints 'br/public:avm/res/network/private-endpoint:0
         'Full'
       ).location
       lock: privateEndpoint.?lock ?? lock
-      privateDnsZoneGroup: privateEndpoint.?privateDnsZoneGroup
+      privateDnsZoneGroup: {
```
AlexanderSehr, line 330: Please note. the Interface still allows the user to define a custom `privateDnsZoneGroup` even though the module would ignore it.
Either this module should implement its on PE interface and essentially not allow the user to set the `privateDnsZoneGroup` - or - this could be changed to something like
```bicep
privateDnsZoneGroup: privateEndpoint.?privateDnsZoneGroup ?? {
```
To, in theory still allow the user to bring a private DNS Zone. Otherwise it would appear quite confusing that the parameter is there, but its value would be thrown away if you catch my drift 😄

AlexanderSehr: Come to think of it, it would probably be a good call to call this also out in the description of the `privateEndpoints` parameter

avm/res/web/static-site/main.bicep
ChrisSidebotham: #4345 has been open since Jan 29th, here we are talking about a new resource call which is being updated. I would rather us push this change to resolve the current and very limited issue then look at refinements on a wider basis following this. I am unsure what we have to discuss around this, it should be a simply implementation unless the user brings there own DNS Zone?

avm/res/web/static-site/main.bicep
ChrisSidebotham: This is the intended behaviour, in most cases DNS Zones are stored centrally so the user will want to BYODnsZone ...

However I have updated the PE Description now

avm/res/web/static-site/main.bicep
```diff
@@ -63,7 +63,7 @@ import { lockType } from 'br/public:avm/utl/types/avm-common-types:0.5.1'
 param lock lockType?
 
 import { privateEndpointSingleServiceType } from 'br/public:avm/utl/types/avm-common-types:0.5.1'
-@description('Optional. Configuration details for private endpoints. For security reasons, it is recommended to use private endpoints whenever possible. Note, requires the \'sku\' to be \'Standard\'.')
+@description('Optional. Configuration details for private endpoints. For security reasons, it is recommended to use private endpoints whenever possible. Note, requires the \'sku\' to be \'Standard\'. Supplying a `privateDnsZoneGroup` Configuration will allow the private endpoint to point to a already craeted zone group, without this is `createPrivateDnsZone` is enabled it will use the output of the newly created zone.')
```
AlexanderSehr, line 66:
```suggestion
@description('Optional. Configuration details for private endpoints. For security reasons, it is recommended to use private endpoints whenever possible. Note, requires the \'sku\' to be \'Standard\'. Supplying the `privateDnsZoneGroup` configuration will allow the private endpoint to point to an already existing zone group. If not provided and `createPrivateDnsZone` being enabled, it will create and use the newly created zone group.')
```

AlexanderSehr: I hope that's correct. I wasn't entirely sure I understood it correctly.

avm/res/web/static-site/main.bicep
ChrisSidebotham: Fixed as discussed

avm/res/web/static-site/main.bicep
AlexanderSehr: Thanks for the adjustments @ChrisSidebotham, resolving

AlexanderSehr (approved):

PR #4971, opened 2025-04-02
test: Make  'avm/res/sql/server' tests idempotent

peterbud:
## Description

I have reviewed the SQL Server tests, and made the following changes:

* Added `init` & `idem` for tests where they were missing to ensure that tests are running in idempotent way
* Removed unnecessary dependencies

I left the `max` and `waf` tests as single, as they are already running 20+ mins, if we would add the `init` and `idem` for them (which I believe would be possible), it would take even longer to run them in CI. LMK if you would like me to add them as well.

## Pipeline Reference

| Pipeline |
| -------- |
|     [avm.res.sql.server](https://github.com/peterbud/bicep-registry-modules/actions/workflows/avm.res.sql.server.yml)     |

## Type of Change

  - [ ] Update to documentation

## Checklist

### DIFF
--- a/avm/res/sql/server/tests/e2e/admin/main.test.bicep
+++ b/avm/res/sql/server/tests/e2e/admin/main.test.bicep
@@ -43,21 +43,20 @@ module nestedDependencies 'dependencies.bicep' = {
 // ============== //
 // Test Execution //
 // ============== //
-
-module testDeployment '../../../main.bicep' = {
-  scope: resourceGroup
-  name: '${uniqueString(deployment().name, resourceLocation)}-test-${serviceShort}'
-  params: {
-    name: '${namePrefix}-${serviceShort}'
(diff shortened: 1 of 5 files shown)


krbar:
Hi @peterbud, thank you for your contribution! I think the duration of the `max` and `waf-aligned` tests would not be an issue. We have modules which deploy way longer than that and it is likely that the `idem` test will be shorter than the `init` test. So if there are no technical reasons to skip the idempotency test, I would suggest to add it also in these two cases. What do you think?

peterbud:
@krbar , thanks I have added also `init` and `idem` for the `max` and `waf` tests, and indeed they are not so long running. Pls see the pipeline runs.

krbar (approved):
Thanks @peterbud, looks good! 👍

PR #4973, opened 2025-04-02
fix: fixes CI tests `ptn/deployment-script/import-image-to-acr`

### LINKED ISSUE #4918: [Failed pipeline] avm.ptn.deployment-script.import-image-to-acr
Failed run: https://github.com/Azure/bicep-registry-modules/actions/runs/14110854113

### LINKED ISSUE #4943: [Failed pipeline] avm.ptn.deployment-script.import-image-to-acr
Failed run: https://github.com/Azure/bicep-registry-modules/actions/runs/14110854113

ReneHezser:
## Description

The names for Azure resources in the test cases are now calculated and should not generate "name in use" exceptions. 
For the max test to pass, we are waiting for [Resources in nested template not shown if reference is used in the parameter values](https://github.com/Azure/arm-template-whatif/issues/157)

Closes #4943 
Closes #4918

## Pipeline Reference

| Pipeline |
| -------- |
|  [avm.ptn.deployment-script.import-image-to-acr](https://github.com/ReneHezser/bicep-registry-modules/actions/workflows/avm.ptn.deployment-script.import-image-to-acr.yml) |

## Type of Change

  - [ ] Update to documentation

## Checklist


avm/ptn/deployment-script/import-image-to-acr/tests/e2e/waf-aligned/main.test.bicep
```diff
@@ -28,14 +28,14 @@ module dependencies 'dependencies.bicep' = {
   name: 'dependencies'
   scope: resourceGroup
   params: {
-    acrName: 'dep${namePrefix}acr${serviceShort}'
+    acrName: uniqueString('dep${namePrefix}acr${serviceShort}', subscription().subscriptionId, resourceGroupName)
```
AlexanderSehr, line 31: Did we not agree this is a bad call?

AlexanderSehr: Some for all other `uniqueString` usages

avm/ptn/deployment-script/import-image-to-acr/tests/e2e/max/main.test.bicep
```diff
@@ -47,10 +51,11 @@ resource resourceGroup 'Microsoft.Resources/resourceGroups@2021-04-01' = {
 // Test Execution //
 // ============== //
 
-resource keyVault 'Microsoft.KeyVault/vaults@2023-07-01' existing = {
-  name: last(split(dependencies.outputs.keyVaultResourceId, '/'))
-  scope: resourceGroup
-}
+// commented out, as the user is not available in the test environment
```
AlexanderSehr, line 54: If not used, why is it here? 😄

utilities/pipelines/staticValidation/psrule/.ps-rule/containerregistry-contenttrust-suppress.Rule.yaml
AlexanderSehr, line 1: Hey @eriqua, as the master of PSRule, I leave this to you. Makes sense to me what I'm seeing here and also the reason that @ReneHezser has given, but I'm not fluent in PSRule and the setup 😄  

I sense however that we'd want to disable the workflows before merging this PR.

avm/ptn/deployment-script/import-image-to-acr/tests/e2e/max/main.test.bicep
ReneHezser: We discussed this before ;-)
For documentation purposes, the usage of Key Vault for storing secrets (line 76) is shown.

avm/ptn/deployment-script/import-image-to-acr/tests/e2e/waf-aligned/main.test.bicep
ReneHezser: We didn't want to use tags. The likeliness of our "manually" generated names to be not unique is higher than this approach.

AlexanderSehr: Ah I think that was a conversation in a call I wasn't able to join. Pardon

We absolutely should not use tags. We did in the very beginning of AVM's predecessor's CARML's predecessor IaCS and has many issues one may not think off when approaching it. E.g., that they're not supported by all resources, nor propagated immediately.

The 'manually' generated names are fine as the `namePrefix` is unique to your fork and the `serviceShort` unique to each test case. Plus it can be consistently applied to all scopes. Using e.g. the `subscription()` function does not work for any scope above. Same for resource group etc. The other problem is that it becomes impossible to trace the name back to its origin. 

If you have a conflict, there are several other **subjectively** better options (which may be objectively wrong 😄) 
- Change the `namePrefix`
- Add a number suffix to the resource name
- Use the `uniqueString()`, but only for a suffix including a `take()` function to ensure the name is not getting too long

I'd probably argue the 3rd option is probably the best and could be rolled out to each test. Using e.g. `uniqueString(subscription().subscriptionId)` would ensure the name is **always** unique to each user and the entire `namePrefix` would not be needed. For mgmtGroup we could I guess use it's ID. Something like that. Thoughts?

cc: @eriqua (as you may be interested too)

eriqua: I'm not a big fan of updating all test cases for an issue which seems to be very limited. We have multiple name chunks decreasing the likelihood of name conflicts, plus not all resources have a globally unique name. 
As mentioned in our call I'd suggest the number suffix as the easiest approach. Option 3 Alex suggested is also fine, but I'd suggest to keep it to the owner to change if needed and only for globally unique resource names

utilities/pipelines/staticValidation/psrule/.ps-rule/containerregistry-contenttrust-suppress.Rule.yaml
eriqua: Hey @ReneHezser, @AlexanderSehr

If any issue with the resource not supporting the rule, I'd suggest to reach out to Bernie to fix the root cause if possible or at least track this. If lucky we may be able to remove the exception in a short time frame.

If an issue with our test file I guess this would be similar to what we have with vm images pattern (a config valid in general but not for specific use cases). 

What is the blocker in this case? And is this a dependency or a resource deployed by the template?

avm/ptn/deployment-script/import-image-to-acr/tests/e2e/max/main.test.bicep
AlexanderSehr: Ok, I trust you on this. Though I find it odd to keep the comment around given that it does not provide any additional value but showing a Bicep-native capability that's unrelated to this module itself.

avm/ptn/deployment-script/import-image-to-acr/tests/e2e/waf-aligned/main.test.bicep
AlexanderSehr: Is anything still happening here?

ReneHezser: Jep, just now. Sorry for the delay.
@eriqua since Alex is out, could you give your blessings?

avm/ptn/deployment-script/import-image-to-acr/CHANGELOG.md
```diff
@@ -0,0 +1,17 @@
+# Changelog
```
eriqua, line 1: I guess the world is not yet ready for this 😄

utilities/pipelines/staticValidation/psrule/.ps-rule/containerregistry-contenttrust-suppress.Rule.yaml
```diff
@@ -0,0 +1,13 @@
...
+kind: SuppressionGroup
+metadata:
+  name: "SuppressNSParameters"
+spec:
+  rule:
+    - Azure.ACR.ContentTrust
+  if:
+    source: Template
+    withinPath:
+      - avm/ptn/deployment-script/import-image-to-acr # Container Registry images should be signed, which is not feasible for the test-case.
```
eriqua: Is this the case in general or only for this ptn module? isn't the rule failing for the acr module?

avm/ptn/deployment-script/import-image-to-acr/tests/e2e/max/main.test.bicep
eriqua: I also would have suggested to remove the commented code. Just double checking if you both eventually agreed to keep it (sorry I missed the conversation)

avm/ptn/deployment-script/import-image-to-acr/CHANGELOG.md
ReneHezser: It should be, as the [Spec already requires each new version to be in the Changelog](https://azure.github.io/Azure-Verified-Modules/spec/BCPNRF22/#id-bcpnrf22---category-publishing---changelog).

utilities/pipelines/staticValidation/psrule/.ps-rule/containerregistry-contenttrust-suppress.Rule.yaml
ReneHezser: It would be, but the module disabled content trust in the properties. I've adjusted both of the failing tests to do the same. That way, we get around of adding the PSRule exclusion.

avm/ptn/deployment-script/import-image-to-acr/tests/e2e/max/main.test.bicep
ReneHezser: All right. When both of you would like to have it removed, I added it :-P
The test is now complete and not commented out.

avm/ptn/deployment-script/import-image-to-acr/CHANGELOG.md
eriqua: Should we add a paragraph with agreed initial version text for 0.4.3? I guess with the changelog test we'll have to add an initial version for all versioned modules. This would differ with every other changelog file as it starts directly with the changes of this 0.4.4 version.

ReneHezser: taking this offline

avm/ptn/deployment-script/import-image-to-acr/CHANGELOG.md
```diff
@@ -0,0 +1,28 @@
...
+## 0.4.4
+
+### Changes
+
+- Updated resource provider versions
+- In the test cases, the Azure Container Registry, storage account and Key Vault names are now unique to prevent deployment failures due to duplicate names
+- Updated az cli version in the deployment script
+- Added min and max length to the ACR name property
+- Exclude Content Trues psrule `Azure.ACR.ContentTrust` for this module
+- Disabled contentTrust in the 'max' and 'waf-aligned' tests, as creating signed images is a big overhead for the CI deployment testing and raise an error in the PSRule `Azure.ACR.ContentTrust`
```
eriqua: I guess this is not the case anymore. Anyway I would not add psrule exclusions in the changelog as that is a CI specific change, not part of the module and may be confusing for module consumers, especially if they don't know what PSRule is

```suggestion
```

avm/ptn/deployment-script/import-image-to-acr/CHANGELOG.md
```diff
@@ -0,0 +1,28 @@
+# Changelog
+
+The latest version of the changelog can be found [here](/Azure/bicep-registry-modules/blob/main/avm/ptn/deployment-script/import-image-to-acr/CHANGELOG.md).
+
+## 0.4.4
+
+### Changes
+
+- Updated resource provider versions
+- In the test cases, the Azure Container Registry, storage account and Key Vault names are now unique to prevent deployment failures due to duplicate names
```
eriqua: Don't you think we should only add here changes to main template only? i.e. the ones qualifying for publishing? Not sure if module consumers would need to know about test file updates and I'd see changelogs useful for module consumers mainly.

```suggestion
```

ReneHezser: Good point. Haven't thought about that.

avm/ptn/deployment-script/import-image-to-acr/CHANGELOG.md
ReneHezser: Thanks, this point is not true anymore. Additionally, the second point is about the test cases. See your comment below.

eriqua (approved):
🆗

PR #5006, opened 2025-04-08
test: Update cb-waf-security.Rule.yaml

jeanchg:
## Description
added the following Security PSRules:
      - Azure.SQL.AADOnly
      - Azure.SQL.MinTLS
      - Azure.SQL.VAScan

  - [ ] Update to documentation

## Checklist

### DIFF
--- a/utilities/pipelines/staticValidation/psrule/.ps-rule/cb-waf-security.Rule.yaml
+++ b/utilities/pipelines/staticValidation/psrule/.ps-rule/cb-waf-security.Rule.yaml
@@ -39,6 +39,9 @@ spec:
       - Azure.Firewall.PolicyMode
       - Azure.Redis.MinTLS
       - Azure.Redis.NonSslPort
+      - Azure.SQL.AADOnly
+      - Azure.SQL.MinTLS
+      - Azure.SQL.VAScan
       - Azure.Storage.DefenderCloud
       - Azure.Storage.Defender.MalwareScan
       - Azure.Storage.SecureTransfer


rahalan (approved):

rahalan (approved):

utilities/pipelines/staticValidation/psrule/.ps-rule/cb-waf-security.Rule.yaml
```diff
@@ -39,6 +39,9 @@ spec:
       - Azure.Firewall.PolicyMode
       - Azure.Redis.MinTLS
       - Azure.Redis.NonSslPort
+      - Azure.SQL.AADOnly
```
AlexanderSehr, line 42: Beware - if merged - this will trigger **all** workflows. If that is **not** the inteded behavior, I'd highly recommend to disable the actions first, merge the PR, re-enable the actions again and from the looks of it, run the SQL Namespace modules in the main branch to validate if they still pass.

Would you agree @eriqua?

jeanchg: it makes sense and would probably a generic process to use, as we add more PSRules in our custom baselines, right?

Here it's just an example where the SQL Server Module Owner mentioned that a new PSRule had recently been created for his module, I wanted to add it and I also added a couple that existed but were missing in the baseline. The SQL Server Module Owner is expecting this change but we don't need to trigger all workflows for that.

rahalan: @AlexanderSehr which action would that be? Isn't the affected yaml just part of the static code testing?

AlexanderSehr: Hey @rahalan,
great question. It's not an action, but the triggers that catch on:
```yaml
      - "utilities/pipelines/**"
      - "!utilities/pipelines/platform/**"
```

AlexanderSehr: Ah it's pretty straight forward
1. Disable actions in the repo's settings (just make sure to inform the other maintainers beforehand)
2. Approve & merge the PR
3. Enable the actions again
4. Run the SQL-Server module from the main branch (which will then either pass or fail the 'new' PSRule tests, configured by @jeanchg)

Mind you, this is and also should not be the standard way of merging PRs - but - for PRs that only impact 1 module but would trigger 150 workflows and exception is usually worth it

eriqua: Agree with the above approach. @jeanchg let me know if I can help with that, not sure if everyone has the permissions to perform it

rahalan: @eriqua , unfortunately neither @jeanchg or myself have the permission to disable actions. Can you help there? Thanks!

AlexanderSehr: Recommending to merge this in right after https://github.com/Azure/bicep-registry-modules/pull/5128 to account for potentially failing tests.
Also @peterbud a heads up, as these new WAF rules may fail for the Sql server and would the need to be addressed in the module or tests. Or maybe they don't fail 🤞

peterbud: Thanks @AlexanderSehr  for the heads-up.

Maybe it's a naive question, what would be the difference after this PR merged in terms of running the deployment pipeline? As I see these rules are _already_ running for WAF successfully:

image
 
 What is going to change?

peterbud: Thanks @AlexanderSehr  - I think now I understand better.

So if this change will _enforce_ the additional rules only for WAF, then we are good to go, as those rules are already passing for the WAF test case, and only the `Azure.Deployment.SecureParameter` rule fails as we have a `keys` param which is actually not a secure param by design.

peterbud (approved):

AlexanderSehr (approved):

peterbud:
I'm not sure it was intentional, but as I see now the sql deployment pipeline fails, as NOT only the `waf` tests are enforced/failing, but also the `default` test are as well, at least for the "AVM Custom Baseline [defaults]". My understanding is/was that only the WAF test will be failing.

Our 'default' test case was never meant to have AAD authentication, and also by default we don't want to turn on vulnerability assessment, so I believe enforcement of rules should be only applied to WAS tests. 

Any thoughts?

AlexanderSehr:
> I'm not sure it was intentional, but as I see now the sql deployment pipeline fails, as NOT only the `waf` tests are enforced/failing, but also the `default` test are as well, at least for the "AVM Custom Baseline [defaults]". My understanding is/was that only the WAF test will be failing.
> 
> Our 'default' test case was never meant to have AAD authentication, and also by default we don't want to turn on vulnerability assessment, so I believe enforcement of rules should be only applied to WAS tests.
> 
> Any thoughts?

Hey @peterbud,
that was/is by design. The idea is the following
- If a user only provides the bare minimum of parameters, the defaults of the module should still be as WAF-aligned as possible (example: Local Authentication should be disabled by default by setting a correponding boolean's default accordingly). 
- However, as you cannot set all values in the default test as some require active user input (e.g., a subnet for private endpoints) the WAF test goes all in to show how to deploy it properly with all WAF-aligned options which are not already covered by defaults.

No other test is validated like this. For the failing defaults test that means that we'd need to check 2 things
1. Are any one of them failing because of a default that we can change to a WAF-aligned value
2. OR - if not - we can disable that test via the corresponding PSRule configuration file for the defaults test (which should always only be the 2nd option).

I hope the above makes sense. I'm sure there's some great documentation somewhere, but with a bit of luck it was good enough a summary.

peterbud:
Thanks @AlexanderSehr for the explanation.

I have asked earlier in this thread which scenarios (`default` | `waf`) are impacted by this enforcement, and I apparently there is a third "flavour" the "AVM Custom Baseline" which I was not aware of. maybe it is something a more recent addition?

Anyhow, I'm not sure how this can be solved:
* `Azure.SQL.AADOnly`: this requires the existence of a service principal (dependent on that) so I'm wondering how that can be added to the default? Do we have such AVM modules which tackle such dependencies at the default test case?
* `Azure.SQL.VAScan`: same story, it requires a storage account to use as the repository for the vulnerability scans.

I can add dependencies to the default test case if that is the expectation, just it seems strange that the default, bare minimum test will require now a handful of dependencies.

AlexanderSehr:
Hey @peterbud,
what we could do is the following 
- `Azure.SQL.AADOnly`
  - Change the defaults inputs to
     ```bicep
     params: {
       administrators: {
         principalType: 'Application'
         azureADOnlyAuthentication: true
         sid: deployer().objectId
         login: deployer().objectId
       }
     }
     ```  
   - So to replace the current password-based login with the alternative - using the current deployment principal (usually the MSI)
   - As either parameter (normal login, or admin login) is conditionally required I'd argue this is good enough. If we don't want to rely on the `deployer()` bin an application, we could I guess alternatively deploy an MSI as via a `dependencies.bicep` deployment.
-  `Azure.SQL.VAScan`
   - Add this rule to [min-suppress.Rule.yaml](https://github.com/Azure/bicep-registry-modules/blob/main/utilities/pipelines/staticValidation/psrule/.ps-rule/min-suppress.Rule.yaml) below the SQL-MI exlcusions as 
    ```yaml
    # SQL Server
    - Azure.SQL.VAScan
    ```

peterbud:
OK, thanks @AlexanderSehr, I'll submit a PR with these changes soon

PR #5117, opened 2025-04-27
fix: SQL-Server - Fixed minor bug of key rotation

### LINKED ISSUE #2369: [Feature Request]: Add CMK implementation to SQL-Server
### Description

Add CMK implementation to SQL-Server (ref Azure/bicep-registry-modules#2514)

AlexanderSehr:
## Description

The way the SQL server resource handels CMK is a bit different from most other services. 
- The Key itself **must** be provided with a key version (i.e., it is **not** optional)
- However, auto-rotation is still possible, but it uses a dedicated resource for it called 'encryption-protector'. That, in turn, allows you to configure CMK rotation

Other
- Cleaned up CMK test (unused key).
- Added descriptions to types
- Aligned zonal implementation with AVM spec
- Aligned several properties with AVM specs
- Aligned some of the type names

Test: 
1. Deploy default deployment
   image
2. Generated a new version of the `keyServerEncryptionKey` and waited for a short time
3. Result
   image
 

## Pipeline Reference

| Pipeline |
| -------- |
|    [avm.res.sql.server](https://github.com/AlexanderSehr/bicep-registry-modules/actions/workflows/avm.res.sql.server.yml)      |

## Type of Change

  - [ ] Update to documentation


avm/res/sql/server/tests/e2e/cmk/dependencies.bicep
peterbud, line 1: It's OK to remove the dbKey, we can put it back once we get to the implementation of database level CMK

peterbud:
That's a lot of changes - thank you!

Question:
Theoretically the user still can configure the CMK without the `keyVersion`, right? I could not figure out what happens in the following cases:
  * `keyVersion` provided, `autoRotation` set to true: that's where we have a test set up, so that works
  * `keyVersion` NOT provided, `autoRotation` set to true
  * `keyVersion` provided, `autoRotation` set to false
  * `keyVersion` NOT provided, `autoRotation` set to false

AlexanderSehr:
> That's a lot of changes - thank you!
> 
> Question: Theoretically the user still can configure the CMK without the `keyVersion`, right? I could not figure out what happens in the following cases:
> 
> * `keyVersion` provided, `autoRotation` set to true: that's where we have a test set up, so that works
> * `keyVersion` NOT provided, `autoRotation` set to true
> * `keyVersion` provided, `autoRotation` set to false
> * `keyVersion` NOT provided, `autoRotation` set to false

Hey @peterbud,
it is indeed an interesting case as the resource has this kinda odd `encryption-protector` resource. For your cases, my understanding the following:
 - `keyVersion` provided, `autoRotation` set to true: that's where we have a test set up, so that works
   -  In this case, the key version is used for the initial version. If it were to change in the key vault via a new version (see screenshots), the auto-rotation will pick up on it a few minutes later
 - `keyVersion` NOT provided, `autoRotation` set to true
   - In this case, the same applies as in the previous case, just that the template will catch the latest version at the deployment time for the user. Again, if that key is updated in the key vault, the auto-ration will pick up on it. The key version remains mandatory for the RP - but - providing it is optional for the user because the template automatically gets the latest for him/her.  
 - `keyVersion` provided, `autoRotation` set to false
   - In this case, the key version specified will be used. An update to the same in the key vault will **not** be picked up by the RP and it will continue to use the version specified at deployment time. 
 - `keyVersion` NOT provided, `autoRotation` set to false
   - Finally, in this case, the template again fetches the latest version at deployment time on behalf of the user. The RP will **not** pick up on any updates to that key in the keyvault and continue to use the original version.

I hope that makes sense :)

avm/res/sql/server/tests/e2e/cmk/dependencies.bicep
AlexanderSehr: Correct. It's currently not used and at least was a bit confusing for me, when troubleshooting 😄 Let's add it when we need it.

peterbud (approved):
Thanks once again!

PR #5207, opened 2025-05-13
fix: Set proper object defaults to fix Logic Apps Designer not loading #5124

### LINKED ISSUE #5124: [AVM Module Issue]: Logic Apps (workflow) Designer not loading

### Issue Type?

I'm not sure

### Module Name

avm/res/logic/workflow

### (Optional) Module Version

_No response_

### Description

When deploying the module with **defaults** the Logic App designer in the Azure portal won't load properly. It just loads forever.

The "solution" that I found is to replace it with a template or to add the following in the bicep (minimum). 

```bicep
module logicapp 'br/public:avm/res/logic/workflow:0.5.1' = {
  name: 'DeployLogicApp'
  scope: resourceGroup(rg.name)
  params: {
    name: 'deployment>
    location: <location>
    state: 'Enabled'
    workflowTriggers: {}
    workflowActions:  {}
    workflowOutputs:  {}
  }
}
```

Anybody else that can confirm this?

### (Optional) Correlation Id

_No response_

lsnoddy:
## Description
Set workflowActions, workflowOutputs, and workflowTriggers to default to empty objects to correct an issue where the logic app designer would fail to load if these parameters were left as null.

Closes #5124

## Pipeline Reference

| Pipeline |
| -------- |
[avm.res.logic.workflow](https://github.com/lsnoddy/bicep-registry-modules/actions/workflows/avm.res.logic.workflow.yml)
|          |

## Type of Change

  - [ ] Update to documentation

## Checklist


avm/res/logic/workflow/main.bicep
```diff
@@ -62,7 +62,7 @@ param tags object?
 param triggersAccessControlConfiguration object?
 
 @description('Optional. The definitions for one or more actions to execute at workflow runtime.')
-param workflowActions object?
+param workflowActions object = {}
```
AlexanderSehr, line 65: Given this comment : https://github.com/Azure/bicep-registry-modules/issues/5124#issuecomment-2865694034

I'm not sure we want that to happen.

There is another alternative if both `null` and `{}` are not good ideas, and that is to implement it in a way where the property is not passed in at all if empty. This naturally has the caveat that you can not every remove all steps from the Logic App, as an empty object would just be ignored, but I'm not the expert here.

If you'd want to do that, you'd need to implement each property that should behave this way using the `shallowMerge` feature. 

For example
```bicep
workflowActions: ...(!empty(workflowActions) ? { actions: workflowActions } : {})
```

lsnoddy: @AlexanderSehr, I like the suggestion, however in testing it, if the property is not passed in, the logic app designer will not load. As such, the property must exist with an empty bracket {} if nothing is specified in the parameter. Isn't that expected behavior though? If I specify one of the properties as null, it should overwrite settings that may have been made in the portal? The deployment should reflect what is specified in the template.

AlexanderSehr: There are always two philosophies, but as you already validated it and one one works it makes the discussion easy 😄

AlexanderSehr (approved):

PR #5274, opened 2025-05-16
refactor: Refresh `avm/res/document-db/database-account` module to `0.15`

seesharprun:
## Description

Performed general housekeeping of the `avm/res/document-db/database-account` module that includes:

- Update the descriptions and metadata for the core account 
  - I have intentionally not updated the metadata for child modules yet
  - There are multiple issues open to improve the UDTs for child modules and I can update their metadata at that point
- Added a zone redundancy parameter for single-region accounts
  - I'm disabling zone redundancy for many unit tests to speed up testing and work in more environments
- Updated parameter names for some auth-related parameters
- Simplified account kind picker logic
  - `MongoDB` is only used for API for MongoDB RU accounts
- Updated key-based/local auth disabling logic
  - If NoSQL or Table, default to disabled (Use Entra auth)
  - If MongoDB RU or Apache Gremlin, default to enabled (Entra auth not supported)
- Removed legacy secrets export configuration types and logic
- Updated parameter names for new secrets export feature
- Updated key-vault export e2e test
  - Use a key-vault resource and correctly illustrated how to set secrets
  - Get rid of unused dependencies
- Update unit test default locations
  - Updated default locations based on recommendations from the product group
- Disable zone-redundant backup for unit tests that don't need it
  - This should greatly speed up tests and allow more regions
- Updated multi-region test to only test multi-region support
  - Use geo-redundant backup
  - Don't worry about zone redundancy
  - Use two regions
- Updated WAF test to test minimal recommended redundancy
  - Use zone-redundant backup
  - Use two regions
  - Enable zone redundancy
- Resolve various PSRule linter warnings
  - Resolved warnings around key-based/local auth being enabled
  - Resolved warnings around WAF-compliant module not having tags

The goal here is to prepare this module for the UDT rework that's being requested in multiple GitHub issues.

## Pipeline Reference

| Pipeline |
| --- |
| [avm.res.document-db.database-account](https://github.com/seesharprun/bicep-registry-modules/actions/workflows/avm.res.document-db.database-account.yml) |

## Type of Change

  - [ ] Update to documentation

## Checklist


seesharprun:
I'm done with my intended changes, but I'm experiencing transient build issues. I'll keep trying to debug/fix this in the open.

avm/res/document-db/database-account/tests/e2e/kvSecrets/main.test.bicep
AlexanderSehr, line 1: Would argue to remove the test. Yes, it shows how to pass the outputs of the module into a key vault - but technically, this is rather an illustration of the key vault module than document DB 😄.
Moreover, the deployment is now **not** testing the DB Account anymore in any meaningful way (which is expected, given that the previous logic was replaced with the newer, simple, outputs). 💪 
 Thoughts?

AlexanderSehr:
Great work @seesharprun, didn't find anything to comment, but the one bit regarding the secrets-export test to discuss. 💪

avm/res/document-db/database-account/tests/e2e/kvSecrets/main.test.bicep
seesharprun: Is there a meaningful way to document this as an example other than to use a unit test?

The motivation for the unit test is to validate that the deployment is correctly outputting the secrets for each value since they are using hard-coded index after the key lookup:

```bicep
@secure()
@description('The primary read-write connection string.')
output primaryReadWriteConnectionString string = databaseAccount.listConnectionStrings().connectionStrings[0].connectionString

@secure()
@description('The primary read-only connection string.')
output primaryReadOnlyConnectionString string = databaseAccount.listConnectionStrings().connectionStrings[2].connectionString
```

My hope is that the e2e test will "catch" the scenario if someone changes those indexes to either an invalid value or remove the output altogether.

The secondary motivation is to give a simple example of how to use the outputs since we're changing the behavior significantly. If there's a better way to document this, I'm all for it.

AlexanderSehr: There is a way in a sense, yes.

You will find that some `main.test.bicep` like

https://github.com/Azure/bicep-registry-modules/blob/99dbc81a66c1be5e121f770c6e31d84a9764b696/avm/res/cdn/profile/tests/e2e/afd/main.test.bicep#L123-L124

or

https://github.com/Azure/bicep-registry-modules/blob/99dbc81a66c1be5e121f770c6e31d84a9764b696/avm/res/document-db/database-account/tests/e2e/kvSecrets/main.test.bicep#L64-L70

pass the outputs thru, which is something you could also do, say, in the `max` test (but really any would work). The value would **not** be exposed if declared as `secure()` - but - you'd be able to see the names of the outputs in the workflow log with an obscured value.

Aside from this option you can also always add a `## Notes` section to the readme, which is the one section that is actually retained by the automation upon regenerating the readme and appended at the end.

Lastly, again e.g., in the max test you can modify the tests description (it can be multiline afterall) and add guidelines to it as set description is pulled 1:1 into the readme.

I hope the above provided some ideas :)

seesharprun: I'm going to opt to just drop the `max` test altogether and just expose them as outputs on another test like you suggested. Once I get that fix in, I'll run the tests one last time.

AlexanderSehr: I guess you mean to drop the `kvSecrets` and not the `max` test? 
Come to think of it, the module doesn't even have a `max` test. The contributor back then opted to create many seperate tests instead of a few that would test all features together.

seesharprun: You're right, I dropped the `kvSecrets` test and added the secrets to other tests.

Specifically, I exposed them in the **MongoDB RU** and **Apache Gremlin** tests as those two APIs don't support Entra auth at all. Both APIs already have key-based access enabled.

This issue is resolved and I'm waiting for the tests to finalize.

seesharprun:
All tests have passed, and all feedback resolved! @AlexanderSehr

AlexanderSehr (approved):
Thanks for putting in the massive effort @seesharprun 💪 Looks like a big improvement 🚀

PR #5394, opened 2025-06-13
feat: Added enforcement for UDTs/RDTs for modules with version `1.0` and above

AlexanderSehr:
## Description

- Added enforcement for UDTs/RDTs for modules with version `1.0` and above
- Tested with APIM backend module which has several parameters without a type
  - Tested without version file: Created warning
  - Tested with version file with version 0.1: Created warning
  - Tested with version file with version 1.1: Failed the test (as expected)

## Pipeline Reference

| Pipeline |
| -------- |
|   [avm.res.key-vault.vault](https://github.com/Azure/bicep-registry-modules/actions/workflows/avm.res.key-vault.vault.yml)       |

## Type of Change

  - [ ] Update to documentation


utilities/pipelines/staticValidation/compliance/module.tests.ps1
AlexanderSehr, line 1: Recommend disabling actions before merging

ReneHezser (dismissed):
LGTM

utilities/pipelines/staticValidation/compliance/module.tests.ps1
```diff
@@ -799,15 +800,25 @@ Describe 'Module tests' -Tag 'Module' {
...
 
-                    Write-Output @{
-                        Warning = ("$warningMessage<br>- <code>{0}</code><br>" -f ($incorrectParameters -join '</code><br>- <code>'))
+                    $versionFilePath = Join-Path (Split-Path $templateFilePath) 'version.json'
+                    $hasVersionFile = Test-Path $versionFilePath
+                    if ($hasVersionFile) {
+                        $moduleVersion = [version](Get-Content $versionFilePath -Raw | ConvertFrom-Json).version
+                    }
+                    if ((Test-Path $versionFilePath) -and $moduleVersion -ge [version]'1.0') {
+                        # Enforcing test for 1.0 modules
```
eriqua:
```suggestion
                        # Enforcing test for modules with a version greater than 1.0
```

utilities/pipelines/staticValidation/compliance/module.tests.ps1
```diff
@@ -799,15 +800,25 @@ Describe 'Module tests' -Tag 'Module' {
...
+                        $incorrectParameters | Should -BeNullOrEmpty -Because ('all parameters which are of type [object] or [array-of-objects] should implement a user-defined, or resource-derived type. Found incorrect items: [{0}].' -f ($incorrectParameters -join ', '))
+                    } else {
+
+                        $warningMessage = 'All parameters which are of type [object] or [array-of-objects] should implement a user-defined, or resource-derived type. Found incorrect items '
```
eriqua:
```suggestion
                        $warningMessage = 'All parameters which are of type [object] or [array-of-objects] should implement a user-defined, or resource-derived type. Found incorrect items: '
```

utilities/pipelines/staticValidation/compliance/module.tests.ps1
```diff
@@ -799,15 +800,25 @@ Describe 'Module tests' -Tag 'Module' {
                 }
 
                 if ($incorrectParameters.Count -gt 0) {
-                    $warningMessage = 'All parameters which are of type [object] or [array-of-objects] should implement a user-defined, or resource-derived type. Found incorrect items '
-                    Write-Warning ("$warningMessage`n- {0}`n" -f ($incorrectParameters -join "`n- "))
 
-                    Write-Output @{
-                        Warning = ("$warningMessage<br>- <code>{0}</code><br>" -f ($incorrectParameters -join '</code><br>- <code>'))
+                    $versionFilePath = Join-Path (Split-Path $templateFilePath) 'version.json'
+                    $hasVersionFile = Test-Path $versionFilePath
```
eriqua: Wondering if we can leverage the versionFileExists at row [540](https://github.com/Azure/bicep-registry-modules/pull/5394/files#diff-9919fd54ac0072d19fc1983df20b7363f589a83758d0ae2db89ec5b6294c8ff5R540)

utilities/pipelines/staticValidation/compliance/module.tests.ps1
```diff
@@ -799,15 +800,25 @@ Describe 'Module tests' -Tag 'Module' {
...
+                    if ($hasVersionFile) {
+                        $moduleVersion = [version](Get-Content $versionFilePath -Raw | ConvertFrom-Json).version
+                    }
+                    if ((Test-Path $versionFilePath) -and $moduleVersion -ge [version]'1.0') {
```
eriqua: Why retesting if we got a var
```suggestion
                    if ($hasVersionFile -and $moduleVersion -ge [version]'1.0') {
```

utilities/pipelines/staticValidation/compliance/module.tests.ps1
AlexanderSehr: We can and I was thinking about it. The only reason I ended up not doing it was because I anyways need to pass the path in. So instead of having 2 parameters sit there and 1 check, I opted for 1 parameter and 2 checks 😄 
Happy to change it though if you prefer the parameter option :)

utilities/pipelines/staticValidation/compliance/module.tests.ps1
AlexanderSehr: See above

utilities/pipelines/staticValidation/compliance/module.tests.ps1
eriqua: Fair, just double checking if conscious choice 🙂 Please revolve

utilities/pipelines/staticValidation/compliance/module.tests.ps1
eriqua: maybe I'm missing something. In this care the var is created in this test (2 lines above)

utilities/pipelines/staticValidation/compliance/module.tests.ps1
AlexanderSehr: Updated

utilities/pipelines/staticValidation/compliance/module.tests.ps1
AlexanderSehr: Ah pardon I think I misunderstood. The reason is that I want to write a warning for all cases where there's **not** a version file **or** if there's one when its version is below 1.0.

If I replace the `$moduleVersion -ge [version]'1.0'` with `[version](Get-Content $versionFilePath -Raw | ConvertFrom-Json).version` directly, it will throw an exception if the parth does not exist. 
```pwsh
if ($VersionFileExists -and [version](Get-Content $versionFilePath -Raw | ConvertFrom-Json).version -ge [version]'1.0') {
    # Enforcing test for modules with a version greater than 1.0
    $incorrectParameters | Should -BeNullOrEmpty -Because ('all parameters which are of type [object] or [array-of-objects] should implement a user-defined, or resource-derived type. Found incorrect items: [{0}].' -f ($incorrectParameters -join ', '))
} else {
  # Write as warning
}
```

So the only alternative is that I pull the `[version](Get-Content $versionFilePath -Raw | ConvertFrom-Json).version` into the if and then have an extra condition nested in here. However, if I do, I cannot write a warning anymore in case the version is less than `1.0`:

```pwsh
if ($VersionFileExists) {
    $moduleVersion = [version](Get-Content $versionFilePath -Raw | ConvertFrom-Json).version
    if ($moduleVersion -ge [version]'1.0') {
        # Enforcing test for modules with a version greater than 1.0
        $incorrectParameters | Should -BeNullOrEmpty -Because ('all parameters which are of type [object] or [array-of-objects] should implement a user-defined, or resource-derived type. Found incorrect items: [{0}].' -f ($incorrectParameters -join ', '))
    } else {
        # MALEDETTO!
        # I'd need to duplicate the code from the warning below
    }
} else {
  # Write as warning
}
```

eriqua (approved):
🆗

PR #5431, opened 2025-06-18
Added extended zone support to virtual network module

mjrich19:
I don't know how to properly create a pull request but I have added support for extended zones to the virtual network module and verified that the updated module works.


ReneHezser:
@mjrich19 we can support you on this. No worries.

ReneHezser:
There are a couple of things you must do:

- Bump the version, as you introduce a new parameter
- Create a new testcase for that one, or update existing tests, at least the max test, to include the new parameter in testing.
- Update the Readme
- Add a Changelog entry

You can read about the contribution guidance here: https://azure.github.io/Azure-Verified-Modules/contributing/bicep/bicep-contribution-flow/generate-bicep-module-files/

eriqua:
@mjrich19 is this still being worked on? Please let us know if you need help.

mjrich19:
@eriqua Yes. I am still working on this. I switched laptops and this caused delays because I had to set everything up again.

However, I still have an issue. I cannot create a test case because extended zones are only available in 2 Azure regions and they require a special subscription registration step. Not simply a resource provider registration. I have manually validated that the updates work extensively in my own environment.

I also have another parameter in my local branch I have added support for and a bunch of other changes so I am thinking that I will recall this pull request, combine all of my changes into a larger pull request, and then request help on the proper way to get all of the code into this repository.

eriqua:
> @eriqua Yes. I am still working on this. I switched laptops and this caused delays because I had to set everything up again.
> 
> However, I still have an issue. I cannot create a test case because extended zones are only available in 2 Azure regions and they require a special subscription registration step. Not simply a resource provider registration. I have manually validated that the updates work extensively in my own environment.
> 
> I also have another parameter in my local branch I have added support for and a bunch of other changes so I am thinking that I will recall this pull request, combine all of my changes into a larger pull request, and then request help on the proper way to get all of the code into this repository.

Hi @mjrich19 sounds good. Thanks for the quick reply. Once the new PR is raised, please close this one mentioning the other PR will superseed. 

However, in case a change is required to the target subscription, please share its details so that we check/address the target validation environment properly

PR #5491, opened 2025-06-26
fix: CI for new modules and Changelogs

### LINKED ISSUE #5489: [AVM CI Environment Issue]: Changelog tests fail for new modules

### Issue Type?

Bug

### Description

New modules don't have a published tag in the mcr. `Get-PublishedModuleVersionsList` will fail in that case.

ReneHezser:
## Description

New modules don't have published versions. Instead of getting the latest version, assume 0.1.0 is the current version.

Closes #5489 

## Pipeline Reference

| Pipeline |
| -------- |
|  [avm.res.signal-r-service.signal-r](https://github.com/ReneHezser/bicep-registry-modules/actions/workflows/avm.res.signal-r-service.signal-r.yml)        |

Tested with a new module, a changed module (that will be published) and an unchanged module.

## Type of Change

  - [ ] Update to documentation

## Checklist

### DIFF
--- a/utilities/pipelines/publish/helper/Get-ModulesToPublish.ps1
+++ b/utilities/pipelines/publish/helper/Get-ModulesToPublish.ps1
@@ -204,6 +204,7 @@ Get any template (main.json) files in the given folder path that would qualify f
 .DESCRIPTION
 Get any template (main.json) files in the given folder path that would qualify for publishing.
 Uses Head^-1 to check for changed files and filters them by the module path & path filter of the version.json
+Note: This logic does not return new files that were added, but only files that were modified. Therefore, it is not suitable for new modules that have not been published before.
 
 .PARAMETER ModuleFolderPath
 Mandatory. The path to the module to check for changed files in.
--- a/utilities/pipelines/publish/helper/Get-ModulesToPublish.ps1
+++ b/utilities/pipelines/publish/helper/Get-ModulesToPublish.ps1
@@ -204,7 +204,6 @@ Get any template (main.json) files in the given folder path that would qualify f
 .DESCRIPTION
 Get any template (main.json) files in the given folder path that would qualify for publishing.
 Uses Head^-1 to check for changed files and filters them by the module path & path filter of the version.json
-Note: This logic does not return new files that were added, but only files that were modified. Therefore, it is not suitable for new modules that have not been published before.
 
 .PARAMETER ModuleFolderPath
 Mandatory. The path to the module to check for changed files in.
--- a/utilities/pipelines/sharedScripts/Get-PublishedModuleVersionsList.ps1
+++ b/utilities/pipelines/sharedScripts/Get-PublishedModuleVersionsList.ps1
@@ -14,7 +14,7 @@ Mandatory. The module name of the module. (e.g., `vault/key-vault`, `network/vir
 .EXAMPLE
 Get-PublishedModuleVersionsList -ModuleType 'res' -ModuleName 'vault/key-vault'
 
-Returns the module versions for module `avm/res/vault/key-vault`. E.g., `@(0.1.0, 0.1.1, (...))`
+Returns the module versions for module `avm/res/vault/key-vault`. E.g., `@(0.1.0, 0.1.1, (...))`, `@()`
 
 #>
 
@@ -31,21 +31,26 @@ function Get-PublishedModuleVersionsList {
     )
 
     $tagListUrl = 'https://mcr.microsoft.com/v2/bicep/avm/{0}/{1}/tags/list' -f $ModuleType, $ModuleName
-    Write-Verbose "Getting available tags at '$tagListUrl'..." -Verbose
-    try {
-        $tagListResponse = Invoke-RestMethod -Uri $tagListUrl
-    } catch {
-        if (Test-McrConnection) {
-            Write-Verbose "MCR connection test passed. New modules don't have a published version."
-            return $null
+    for ($i = 0; $i -lt 5; $i++) {
+        try {
+            Write-Verbose "Getting available tags at '$tagListUrl' (attempt $($i + 1))..." -Verbose
+            $tagListResponse = Invoke-RestMethod -Uri $tagListUrl
+            $publishedTags = $tagListResponse.tags | Sort-Object { [Version]$_ } -Culture 'en-US'
+            Write-Verbose "  Found tags: $($publishedTags -join ', ')" -Verbose
+            return $publishedTags
+        } catch {
+            if ($i -eq 4) {
+                if (Test-McrConnection) {
+                    Write-Verbose "MCR connection test passed. New modules don't have a published version."
+                    return @()
+                }
+                Write-Error 'Failed to get tags from MCR ($tagListUrl) after 5 attempts. Please check your network connection and try again.'
+                Write-Error "Error message: $($_.Exception.Message)"
+                throw $_.Exception
+            }
+            Start-Sleep -Seconds 5
         }
-        Write-Error "Error occurred while accessing URL: $tagListUrl"
-        Write-Error "Error message: $($_.Exception.Message)"
-        throw $_.Exception
     }
-    $publishedTags = $tagListResponse.tags | Sort-Object { [Version]$_ } -Culture 'en-US'
-    Write-Verbose "  Found tags: $($publishedTags -join ', ')" -Verbose
-    return $publishedTags
 }
 
 <#
--- a/utilities/pipelines/sharedScripts/Get-PublishedModuleVersionsList.ps1
+++ b/utilities/pipelines/sharedScripts/Get-PublishedModuleVersionsList.ps1
@@ -35,6 +35,10 @@ function Get-PublishedModuleVersionsList {
     try {
         $tagListResponse = Invoke-RestMethod -Uri $tagListUrl
     } catch {
+        if (Test-McrConnection) {
+            Write-Verbose "MCR connection test passed. New modules don't have a published version."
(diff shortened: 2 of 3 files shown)


utilities/pipelines/sharedScripts/Get-PublishedModuleVersionsList.ps1
```diff
@@ -35,6 +35,10 @@ function Get-PublishedModuleVersionsList {
     try {
         $tagListResponse = Invoke-RestMethod -Uri $tagListUrl
     } catch {
+        if (Test-McrConnection) {
```
ReneHezser: Recommend disabling actions for merging.

utilities/pipelines/sharedScripts/Get-PublishedModuleVersionsList.ps1
```diff
@@ -35,6 +35,10 @@ function Get-PublishedModuleVersionsList {
...
+            Write-Verbose "MCR connection test passed. New modules don't have a published version."
```
eriqua: Sometimes connection errors happen. Not reaching mcr does not  guarantee that version does not exist

utilities/pipelines/sharedScripts/Get-PublishedModuleVersionsList.ps1
```diff
@@ -35,6 +35,10 @@ function Get-PublishedModuleVersionsList {
...
+            return '0.1.0'
```
eriqua: Why returning a fake 0.1.0? I suggest relying on the nextExpectedVersion in this case

ReneHezser: It fits the use-case. Returning $null would also be an option.

utilities/pipelines/sharedScripts/Get-PublishedModuleVersionsList.ps1
jtracey93: True. Should we add some retry and also maybe check the last published json file to see if it exists there. 

The combination of these should guarantee it doesn't exist

eriqua: I noticed after that the tested link is mcr and not the tag link. Sorry @jtracey93 I didn't see your comment so resolved. Reopening and adding a suggestion below

eriqua: Suggestion:

This function is used in 2 Pester tests. Both are also retrieving the expected next module version by calling Get-ModuleTargetVersion.
If Get-ModuleTargetVersion returns 0.1.0, we know the module is a new one and we don't need to call Get-PublishedModuleVersionsList.

utilities/pipelines/publish/helper/Get-ModulesToPublish.ps1
```diff
@@ -204,6 +204,7 @@ Get any template (main.json) files in the given folder path that would qualify f
 .DESCRIPTION
 Get any template (main.json) files in the given folder path that would qualify for publishing.
 Uses Head^-1 to check for changed files and filters them by the module path & path filter of the version.json
+Note: This logic does not return new files that were added, but only files that were modified. Therefore, it is not suitable for new modules that have not been published before.
```
eriqua: I don't think this is accurate, otherwise we wouldn't be able to publish new modules 😜

ReneHezser: This module will be published with a new minor version:
image

This module did not change:
image

And for a new module, it is also empty:
image

utilities/pipelines/sharedScripts/Get-PublishedModuleVersionsList.ps1
ReneHezser: We are now retrying 5 times to get the published versions. If it fails, try once to see if we can connect to GitHub at all.
In all cases, the local version.json is taken into consideration.

utilities/pipelines/sharedScripts/Get-PublishedModuleVersionsList.ps1
ReneHezser: Rewritten to be more flexible.

utilities/pipelines/publish/helper/Get-ModulesToPublish.ps1
eriqua: Not sure I follow, I'll reach out offline

utilities/pipelines/sharedScripts/Get-PublishedModuleVersionsList.ps1
eriqua: Is there a reason why you don't want to rely on the suggestion above? That would decrease the number of steps and avoid calling mcr in the first place if not needed

utilities/pipelines/staticValidation/compliance/module.tests.ps1
```diff
@@ -1550,16 +1550,27 @@ Describe 'Module tests' -Tag 'Module' {
 
             $changelogContent | Should -Not -BeNullOrEmpty -Because 'CHANGELOG.md file not found or uncomplete.'
 
-            if ((Get-ModulesToPublish -ModuleFolderPath $moduleFolderPath) -ge 1) {
-                # the module will be published. Use the new version
-                $expectedModuleVersion = Get-ModuleTargetVersion -ModuleFolderPath $moduleFolderPath
+            $moduleTargetVersion = Get-ModuleTargetVersion -ModuleFolderPath $moduleFolderPath
+            if ((Get-ModulesToPublish -ModuleFolderPath $moduleFolderPath) -ge 1 -or $moduleTargetVersion -eq '0.1') {
```
eriqua: Not sure I understand this condition `$moduleTargetVersion -eq '0.1'`

Get-ModuleTargetVersion will return a semantic version x.y.z. it will never return 0.1.

utilities/pipelines/sharedScripts/Get-PublishedModuleVersionsList.ps1
ReneHezser: Which one? Changing from mcr to GitHub tags is on the backlog.

utilities/pipelines/staticValidation/compliance/module.tests.ps1
ReneHezser: Indeed. That was copy&paste and rearchitecture mistake. Fixed.

utilities/pipelines/staticValidation/compliance/module.tests.ps1
```diff
@@ -1550,16 +1550,24 @@ Describe 'Module tests' -Tag 'Module' {
...
             $changelogContent | Should -Not -BeNullOrEmpty -Because 'CHANGELOG.md file not found or uncomplete.'
 
-            if ((Get-ModulesToPublish -ModuleFolderPath $moduleFolderPath) -ge 1) {
-                # the module will be published. Use the new version
-                $expectedModuleVersion = Get-ModuleTargetVersion -ModuleFolderPath $moduleFolderPath
+            $moduleTargetVersion = Get-ModuleTargetVersion -ModuleFolderPath $moduleFolderPath
+            if ((Get-ModulesToPublish -ModuleFolderPath $moduleFolderPath) -ge 1 -or $moduleTargetVersion -eq '0.1.0') {
+                # the module will be published
+                $expectedModuleVersion = $moduleTargetVersion
             } else {
```
eriqua, line 1558: In case the module will not get published, why checking anything? I think the if statement is enough for this test.  if the module does not qualify for publishing the changelog tests should be skipped

utilities/pipelines/staticValidation/compliance/module.tests.ps1
```diff
@@ -1550,16 +1550,24 @@ Describe 'Module tests' -Tag 'Module' {
...
-                # the module will not be published. Use the current version
-                Write-Verbose 'A new version will not be published. Use the current version.' -Verbose
-                $publishedVersions = Get-PublishedModuleVersionsList -ModuleType $moduleType -ModuleName ($moduleFolderName -replace '\\', '/')
+                Write-Verbose 'The module will not be published (in that case use the current version), or is a new module without a version.' -Verbose
```
eriqua: the `is a new module without a version` part is not applicable anymore after adding this condition to the previous if block

utilities/pipelines/staticValidation/compliance/module.tests.ps1
eriqua: Thanks, I think this fixes. Don't forget to implement the same logic at row [1759](https://github.com/Azure/bicep-registry-modules/pull/5491/files#diff-9919fd54ac0072d19fc1983df20b7363f589a83758d0ae2db89ec5b6294c8ff5R1759), the test checking that all versions are published also uses the Get-PublishedModuleVersionsList and in case of new modules we can avoid calling that.

utilities/pipelines/staticValidation/compliance/module.tests.ps1
ReneHezser: We took this offline

eriqua (approved):
:ok:

utilities/pipelines/sharedScripts/Get-PublishedModuleVersionsList.ps1
eriqua: Meant relying on Get-ModuleTargetVersion returning 0.1.0 to skip the call to mcr if not required. Looks good now so resolving

PR #5531, opened 2025-07-02
fix: CI - Added single API version case handling to API test 

AlexanderSehr:
## Description

Before the change, the logic gives the user an incorrectly formatted warning for an outdates API version even though that's not true if only a single API version is available in the first place:

image
([ref](https://github.com/Azure/bicep-registry-modules/actions/runs/16023121941))

Now after the fix, onstead of the incorrectly formatted 10 warnings, only 5 show up: [ref](https://github.com/Azure/bicep-registry-modules/actions/runs/16033652322)

image

## Pipeline Reference

| Pipeline |
| -------- |
|    [avm.res.data-factory.factory](https://github.com/Azure/bicep-registry-modules/actions/workflows/avm.res.data-factory.factory.yml)      |

## Type of Change

  - [ ] Update to documentation


utilities/pipelines/staticValidation/compliance/module.tests.ps1
AlexanderSehr, line 1: Recommend disabling actions before merging

eriqua (approved):
:ok:

utilities/pipelines/staticValidation/compliance/module.tests.ps1
eriqua: I wonder if these tests still make sense now that we have native support in vsc.

AlexanderSehr: Hey @eriqua, I thought the same thing. But there are 2 catches
1. The linter check only produces warnings, and can't see those in the pipeline unless you go into the deployment log
2. The linter check produces a massive amount of warnings (e.g., regarding diagnosticSettings) that we can't do anything about and that I don't expect to be addressed any time soon. Our own tests on the other hand do not have the same issue. Might be worth raising an issue in the Azure/Bicep repository to have their logic be changed and take inspiration from our checks

PR #5547, opened 2025-07-04
fix: CI - Changed Token-Replacement check to only check the current module, as opposed to all including cross-referenced modules

AlexanderSehr:
## Description

- Changed Token-Replacement check to only check the current module, as opposed to all including cross-referenced modules
- This resolves an issue with a top-level module having a child that itself implements telemetry. During the current token replacement, this child is ignored as the token replacement happens in the main bicep file - not the JSON, nor linked files. The original solution was used in the early days to catch issues with incorrectly published cross-referenced modules that has since been ironed out. As such, we can savely pivot and only check the current template for a correct replacement.
- Tested locally using the failing storage account module. 

## Type of Change

  - [ ] Update to documentation

### DIFF
--- a/utilities/pipelines/publish/Publish-ModuleFromPathToPBR.ps1
+++ b/utilities/pipelines/publish/Publish-ModuleFromPathToPBR.ps1
@@ -89,7 +89,7 @@ function Publish-ModuleFromPathToPBR {
         $null = Convert-TokensInFileList @tokenConfiguration
 
         # Double-check that tokens are correctly replaced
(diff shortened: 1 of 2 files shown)


utilities/pipelines/publish/Publish-ModuleFromPathToPBR.ps1
AlexanderSehr, line 1: Recommend disabling tests

eriqua (approved):
:ok:

PR #5633, opened 2025-07-20
feat: Added `weekDaysScheme` and `fail()` conditions

johnlokerse:
## Description

In addition to the `allWeekScheme`, I have added the `weekDaysScheme`, which resolves to Monday through Friday when selected. I have also added checks that trigger `fail(<message>)` if multiple values are provided in the `daysData` property. If both the `weekDaysScheme` and `allWeekScheme` properties are present, `fail()` is triggered. The same applies when either `weekDaysScheme` or `allWeekScheme` is used in combination with an individual day.

@AlexanderSehr What is the official way to test if the `fail()` conditions work in the CI? A failed test would be good in this case. I ran the test locally for now (see additional evidence).

## Pipeline Reference

| Pipeline |
| -------- |
|     [avm.res.dev-ops-infrastructure.pool](https://github.com/johnlokerse/bicep-registry-modules/actions/workflows/avm.res.dev-ops-infrastructure.pool.yml)     |

## Type of Change

  - [] Breaking changes and I have bumped the MAJOR version in `version.json`.
  - [x] Update to documentation

## Checklist

## Additional evidence

The correct weekdays are configured when weekDaysScheme is chosen:

weekdays

Deployment error when `weekDaysScheme` and `allWeekScheme` are used (as designed):

weekdays plus allweek

Deployment error when `weekDaysScheme` and `Monday` is used (as designed):

weekdays plus monday

Deployment error when `allWeekScheme` and `Monday` is used (as designed):

allweek plus monday


avm/res/dev-ops-infrastructure/pool/main.bicep
```diff
@@ -475,4 +502,7 @@ type daysDataType = {
     @description('Required. The agent count to provision throughout the week.')
     provisioningCount: int
   }?
+
+  @description('Optional. A schema to apply to weekdays (Monday to Friday).')
```
AlexanderSehr:
```suggestion
  @description('Optional. A schema to apply to weekdays (Monday to Friday). Overrules daily configurations')
```

avm/res/dev-ops-infrastructure/pool/main.bicep
```diff
@@ -243,6 +269,7 @@ resource managedDevOpsPool_roleAssignments 'Microsoft.Authorization/roleAssignme
   }
 ]
 
+#disable-next-line use-recent-api-versions
```
AlexanderSehr, line 278: Hey @eriqua,
what's you take on this change? The diagnostic settings have not been updated in a long time and the Linter is not robust enough to pick up on the fact that there are no newer versions available (something that may justify a bug in the `Azure/bicep` repo by the way. So right now, using this RP produces a bunch of warnings and noise which I guess is why @johnlokerse introduced the suppression. Just a bit undecide if we should do this as it defeats the Linter rule's purpose at the same time and would not be picked up on if a new version were to be introduced.

avm/res/dev-ops-infrastructure/pool/main.bicep
```diff
@@ -121,23 +121,49 @@ var identity = !empty(managedIdentities)
...
+
+#disable-next-line no-unused-vars
+var allWeekSchemeConflict = hasAllWeekScheme && length(daysData) > 1
+  ? fail('Configuration error: allWeekScheme cannot be combined with other day configurations. Please use either allWeekScheme for all 7 days or other individual day/scheme configurations.')
+  : null
+
+#disable-next-line no-unused-vars
+var weekDaysSchemeConflict = hasWeekDaysScheme && length(daysData) > 1
+  ? fail('Configuration error: weekDaysScheme cannot be combined with other day configurations. Please use either weekDaysScheme for weekdays or individual day configurations.')
+  : null
```
AlexanderSehr: I guess there would also be the 3rd error case, one defines both the `weekDays` & `allWeek` scheme?

AlexanderSehr: We should maybe also surround those checks with a comment block explaining that they're only there for validation. Like
```bicep
// Input validation //
// ================ //

var daysData = agentProfile.?resourcePredictions.?daysData
var hasAllWeekScheme = !empty(daysData) && contains(daysData, 'allWeekScheme')
var hasWeekDaysScheme = !empty(daysData) && contains(daysData, 'weekDaysScheme')

#disable-next-line no-unused-vars
var allWeekSchemeConflict = hasAllWeekScheme && length(daysData) > 1
  ? fail('Configuration error: allWeekScheme cannot be combined with other day configurations. Please use either allWeekScheme for all 7 days or other individual day/scheme configurations.')
  : null

#disable-next-line no-unused-vars
var weekDaysSchemeConflict = hasWeekDaysScheme && length(daysData) > 1
  ? fail('Configuration error: weekDaysScheme cannot be combined with other day configurations. Please use either weekDaysScheme for weekdays or individual day configurations.')
  : null

// ================
```

AlexanderSehr:
Cool idea @johnlokerse 🚀 

Regarding your ask: No, we don't have an approach today to testing that a module failed on purpose. 
I would be a nice feature to add to the CI, though I'm not sure how complicated it would be to achieve. @eriqua, your thoughts would be appreciated too. 💪

avm/res/dev-ops-infrastructure/pool/main.bicep
johnlokerse: Yep, that's why I suppressed it. As far as I know, there is no fix for this yet.

There are multiple issues reported already: https://github.com/Azure/bicep/issues/14840, https://github.com/Azure/bicep/issues/10740, https://github.com/Azure/bicep/issues/10345

avm/res/dev-ops-infrastructure/pool/main.bicep
johnlokerse: There is no need to check if both exist, because it will break when on the first check (`allWeekSchemeConflict`).  See: 
image

avm/res/dev-ops-infrastructure/pool/main.bicep
johnlokerse: Implemented!

avm/res/dev-ops-infrastructure/pool/main.bicep
johnlokerse: @AlexanderSehr Added the validation block.

johnlokerse:
> Cool idea @johnlokerse 🚀
> 
> Regarding your ask: No, we don't have an approach today to testing that a module failed on purpose. I would be a nice feature to add to the CI, though I'm not sure how complicated it would be to achieve. @eriqua, your thoughts would be appreciated too. 💪

I think it's feasible with `continue-on-error` https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions, then catch the output/error to match it against the thrown fail message. Purely theoretical, but I think it can work :)

avm/res/dev-ops-infrastructure/pool/main.bicep
AlexanderSehr: > There is no need to check if both exist, because it will break when on the first check (`allWeekSchemeConflict`). See: image

Touche. You're correct. Thanks for elaborating 🚀

AlexanderSehr:
> > Cool idea @johnlokerse 🚀
> > Regarding your ask: No, we don't have an approach today to testing that a module failed on purpose. I would be a nice feature to add to the CI, though I'm not sure how complicated it would be to achieve. @eriqua, your thoughts would be appreciated too. 💪
> 
> I think it's feasible with `continue-on-error` https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions, then catch the output/error to match it against the thrown fail message. Purely theoretical, but I think it can work :)

That would be one way to not fail on the error, that's correct. 👍 
It may however be a bit better to soft-fail the command itself (i.e., in PowerShell) what we already do today for retries. 

That being said, the bigger challenge is to tell the script somehow, that it should not only soft fail, but also to tell it that it is an intended error (and e.g., not perform a retry) and also implement a mechanic that allows us to validate if the correct error for an individual test case was thrown. And then we'd need to be able to pass the error thru to set validation. I'm sure it's all doable, but it has its pitfalls.

johnlokerse:
Any updates on this @eriqua and @AlexanderSehr?

AlexanderSehr:
Hey @surajguptha & @elizatargithub7, please take a look at your earliest convenience. 🚀 

@johnlokerse - not yet. Pinged the folks again internally. Please excuse the delay - we want to give the owners at least the chance to review their PRs.

AlexanderSehr (dismissed):
I guess the owners won't come around (also did not respond to a direct ping). Will hence go ahead an approve the PR on their behalf.

fyi @matebarabas

avm/res/dev-ops-infrastructure/pool/CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
...
 The latest version of the changelog can be found [here](https://github.com/Azure/bicep-registry-modules/blob/main/avm/res/dev-ops-infrastructure/pool/CHANGELOG.md).
 
+## 0.7.0
+
+### Changes
+
+- Added the `weekDaysScheme` to the `daysData` configuration
+- Added checks to ensure that `daysData` does not contain both `allWeekScheme` and `weekDaysScheme` at the same time or a combination of `weekDaysScheme` and `allWeekScheme` with individual days.
+- Added a test for `weekDaysScheme`
+
```
AlexanderSehr, line 16:
```suggestion

### Breaking Changes

- None

```

Must have a specific pattern

johnlokerse: Fixed!

AlexanderSehr (approved):

johnlokerse:
Thanks @AlexanderSehr 🚀

AlexanderSehr:
> Thanks @AlexanderSehr 🚀

A pleasure as always. Thank you for the great contribution @johnlokerse

PR #5716, opened 2025-07-29
fix: Update Test-ModuleLocally.ps1 to support recursive test cases

mjrich19:
## Description

Enables deployment and validation steps to support test cases which exist in subdirectories. Also skips dependency bicep files

## Pipeline Reference

| Pipeline |
| -------- |

## Type of Change

  - [ ] Update to documentation

## Checklist


AlexanderSehr:
Hey @mjrich19,
thank you for the PR. Looks good per se, but I'd be curious: How does your test folder structure look like? Is it like
- tests
  - e2e
    - defaults
      - main.test.bicep
    - someTestGroup
      - subTestA
        - main.test.bicep 
      - subTestB
        - main.test.bicep

mjrich19:
Not quite. My test folder structure looks like this:

- tests
  - e2e
    - scenario1.defaults
      - main.test.bicep
    - scenario1.max
      - dependencies.bicep
      - main.test.bicep
    - scenario1.waf
      - dependencies.bicep
      - main.test.bicep
    - scenario2.defaults
      - main.test.bicep
    - scenario2.max
      - dependencies.bicep
      - main.test.bicep
    - scenario2.waf
      - dependencies.bicep
      - main.test.bicep
    - scenario3
      - main.test.bicep

I did not test my proposed code with a folder structure like you proposed although I believe it will still work. The only part that might be concerning is if a test case does not end with "main.test.bicep"

AlexanderSehr:
> test my proposed code with a folder structure like you proposed although I believe it will still work. The only part that might be concerning is if a test case does not end with "main.test.bicep"

Gotcha. So your structure is 1:1 the current structure - and instead of nesting folders what you want to do is invoke the script e.g. on the module folder for it to pick up on all test templates in it? For example
```pwsh
$TestModuleLocallyInput = @{
    TemplateFilePath           = 'C:\network\route-table\main.bicep'
    ModuleTestFilePath         = 'C:\network\route-table' # <====
    (...)
}
Test-ModuleLocally @TestModuleLocallyInput -Verbose
```

Also makes sense. Will just add one comment, then we're good to go. Thank you @mjrich19 💪

utilities/tools/Test-ModuleLocally.ps1
```diff
@@ -240,7 +240,7 @@ function Test-ModuleLocally {
             # Find Test Parameter Files
             # -------------------------
             if ((Get-Item -Path $ModuleTestFilePath) -is [System.IO.DirectoryInfo]) {
-                $moduleTestFiles = (Get-ChildItem -Path $ModuleTestFilePath -File).FullName
+                $moduleTestFiles = (Get-ChildItem -Path $ModuleTestFilePath -Filter '*main.test.bicep' -Recurse -Force).FullName
```
AlexanderSehr:
```suggestion
                $moduleTestFiles = (Get-ChildItem -Path $ModuleTestFilePath -Filter '*main.test.bicep' -Recurse).FullName
```
No need to search for hidden/system files 😉

mjrich19:
> > test my proposed code with a folder structure like you proposed although I believe it will still work. The only part that might be concerning is if a test case does not end with "main.test.bicep"
> 
> Gotcha. So your structure is 1:1 the current structure - and instead of nesting folders what you want to do is invoke the script e.g. on the module folder for it to pick up on all test templates in it? For example
> 
> ```powershell
> $TestModuleLocallyInput = @{
>     TemplateFilePath           = 'C:\network\route-table\main.bicep'
>     ModuleTestFilePath         = 'C:\network\route-table' # <====
>     (...)
> }
> Test-ModuleLocally @TestModuleLocallyInput -Verbose
> ```
> 
> Also makes sense. Will just add one comment, then we're good to go. Thank you @mjrich19 💪

Yes. Exactly. I want to be able to invoke the script once and execute all the tests.

mjrich19:
Removed -Force flag from Get-ChildItem call per @AlexanderSehr suggestion. I believe this PR is ready to be accepted and merged

AlexanderSehr (approved):

PR #5812, opened 2025-08-18
fix: CI - Added retry to git fetch

### LINKED ISSUE #5788: [AVM CI Environment Issue]: Publishing sometimes does not detect changes

### Issue Type?

Bug

### Description

As seem for example here: https://github.com/Azure/bicep-registry-modules/actions/runs/16938466039/job/48003564984#step:4:158 no modified files are found even though the commit https://github.com/Azure/bicep-registry-modules/commit/440eee2911af97c5825e581f1f501ed0d1b6fc5c definitely shows them.

Based on the output, it would appears that the command `$diff = git diff --name-only --diff-filter=AM 'upstream/main^'` sometimes does not return the expected output, that is, no output. This code runs here:  https://github.com/Azure/bicep-registry-modules/blob/cc10e43d1854fd8e70a56c87e08a075f52347764/utilities/pipelines/publish/helper/Get-ModulesToPublish.ps1#L28

The interesting thing is that it's not only running the correct code and that I can't replicate the issue, but that it also sometimes works as shown here: https://github.com/Azure/bicep-registry-modules/actions/runs/16938740124/job/48002619979#step:4:159 with the commit https://github.com/Azure/bicep-registry-modules/commit/cc10e43d1854fd8e70a56c87e08a075f52347764

In both cases, the `main.bicep` & `main.json` files are changed. The only difference in the later case is that also the `changelog.md` file changes. However that change is not relevant for publishing and filtered out by the logic.

AlexanderSehr:
## Description

Adding a retry to git fetch as it may not always propagate in the expected time frame.

Example:
Failed case: [ref](https://github.com/Azure/bicep-registry-modules/actions/runs/17022323966/job/48257648824#step:4:157)
image

Should have been:
image
of the same PR for a differnt module: [ref](https://github.com/Azure/bicep-registry-modules/actions/runs/17022323977/job/48253445536#step:4:157)

## Pipeline Reference

| Pipeline |
| -------- |
[avm.res.analysis-services.server](https://github.com/Azure/bicep-registry-modules/actions/workflows/avm.res.analysis-services.server.yml)

## Type of Change

  - [ ] Update to documentation

### DIFF
--- a/utilities/pipelines/publish/helper/Get-ModulesToPublish.ps1
+++ b/utilities/pipelines/publish/helper/Get-ModulesToPublish.ps1
@@ -26,11 +26,17 @@ function Get-ModifiedFileList {
         $currentCommit = git rev-parse 'main' # Get the current main's commit
         $previousCommit = git rev-parse 'upstream/main^' # Get the previous main's commit in upstream
 
+        $retryCount = 0
         while ($currentCommit -eq $previousCommit) {
             Write-Warning 'Current and previous commits are the same. Trying again'
             git fetch 'upstream' 'main' -q # Fetch the latest changes from upstream main
             Start-Sleep 5 # Wait for git to finish fetching
             $previousCommit = git rev-parse 'upstream/main^' # Get the previous main's commit in upstream
+
+            if ($retryCount -ge 5) {
+                throw 'Failed to get a different previous commit after 5 retries. Exiting.'
+            }
+            $retryCount++
         }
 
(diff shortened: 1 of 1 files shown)


utilities/pipelines/publish/helper/Get-ModulesToPublish.ps1
AlexanderSehr, line 1: Actions should be disabled before mergin

jtracey93 (approved):

PR #5819, opened 2025-08-18
feat: Common-Types - Added type to `tags` properties

AlexanderSehr:
## Description

Changed the `object` type of each private endpoint interface's `tags` property to use a user-defined type

## Pipeline Reference

| Pipeline |
| -------- |
[avm.utl.types.avm-common-types](https://github.com/Azure/bicep-registry-modules/actions/workflows/avm.utl.types.avm-common-types.yml)

## Type of Change

  - [ ] Update to documentation

### DIFF
--- a/avm/utl/types/avm-common-types/main.bicep
+++ b/avm/utl/types/avm-common-types/main.bicep
@@ -298,6 +298,7 @@ type privateEndpointSingleServiceType = {
 
   @description('Optional. Tags to be applied on all resources/Resource Groups in this deployment.')
   tags: {
+    @description('Optional. A tag key-value pair.')
     *: string
   }?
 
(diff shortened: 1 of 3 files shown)


avm/utl/types/avm-common-types/main.bicep
```diff
@@ -353,7 +355,9 @@ type privateEndpointMultiServiceType = {
   roleAssignments: roleAssignmentType[]?
 
   @description('Optional. Tags to be applied on all resources/resource groups in this deployment.')
-  tags: object?
+  tags: {
+    *: string
```
jtracey93: Can't tags technically be anything like an object or an array? Not just a string?

jtracey93 (changes requested):

avm/utl/types/avm-common-types/main.bicep
AlexanderSehr: That would be the first time I'd hear about that. Just tested it with a few resources and I always get a warning as soon as I try to use anything but a string. Quote "The property "iamatag" expected a value of type "string" but the provided value is of type "object".bicepBCP036".
Anyways, I changed the type now to the private endpoints-derived one.

jtracey93 (approved):

PR #5887, opened 2025-08-25
docs: Export peeringType and subnetType to enable intellisense formatting

CorneHogerheijde:
We want to use the vnet module from AVM and have subnetType intellisense. However this is not exported, this PR makes sure the type is exported.

## Pipeline Reference

| Pipeline |
| -------- |

## Type of Change

  - [X] Update to documentation

## Checklist

- [X] I have updated the module's CHANGELOG.md file with an entry for the next version


CorneHogerheijde:
@microsoft-github-policy-service agree

mjrich19 (approved):
Looks fine. I don't see any other peering types or subnet types in other modules so this should not cause issues. Thanks for your contribution.

AlexanderSehr:
Hey @mjrich19,
please note that @CorneHogerheijde did not run the static tests and they failed in main: https://github.com/Azure/bicep-registry-modules/actions/runs/17212935387 - presumably because the CI was not set up. There can be good reasons.

However, this means the module update was **not** published. Please take a look into the failed tests (which should be concerned with an outdated `main.json` file) and open a new PR.

CorneHogerheijde:
Looks like it has to do with the changelog changes which were probably updated incorrectly then. Sorry! How to fix this?

CorneHogerheijde:
Ah, I am missing the ### Changes and ### Breaking Changes section. Will provide a PR!

AlexanderSehr:
> Ah, I am missing the ### Changes and ### Breaking Changes section. Will provide a PR!

Good catch @CorneHogerheijde, that's correct. But I think also a `bicep build` for the Bicep template is in order as I believe the `export` flag is also added to the ARM template. Ideally you'd just be running the `Set-AVMModule` script on the module folder as per the contribution guide and run at least the static tests either locally (as also described in the contribution guide) or, even better, in a fork's workflow. In either case the idea is just to make sure everything is compliant.

Granted, given the scope of your implemented change, it may be a bit of an overkill to set up the CI.

PR #5903, opened 2025-08-27
fix: adding cmk to data disks

Dylan-Prins:
## Description

CMK was not added to the Data disks, while the parameter was there.

## Pipeline Reference

| Pipeline |
| -------- |

## Type of Change

  - [ ] Update to documentation

## Checklist


josunefon:
Hi @Dylan-Prins thanks a lot for your contribution, before we approve it, please take a look to the static validation tests that are [failing ](https://github.com/Dylan-Prins/bicep-registry-modules/actions/runs/17280294264)

Dylan-Prins:
> Hi @Dylan-Prins thanks a lot for your contribution, before we approve it, please take a look to the static validation tests that are [failing ](https://github.com/Dylan-Prins/bicep-registry-modules/actions/runs/17280294264)

Yes, i was a little bit hastu with the PR. I will update the PR when the tests are successfull.

Dylan-Prins:
@josunefon

ok, the tests are passed! I only have trouble setting up my key vault for passing the additional parameters. I have created a key vault and gave the app reg the permissions. following by adding the GitHub secret with the key vault name.

Is there someone who can help me troubleshoot?

josunefon:
Hi @Dylan-Prins ,

I cheked the last GH action and I saw that you're having the problem with the backup management service, did you check this requirement for the [custom secrets](https://azure.github.io/Azure-Verified-Modules/contributing/bicep/bicep-contribution-flow/custom-ci-secrets/#example-use-case)?

On the other hand, does the SPN you're using has permissions at key vault secret level? because on the actions error what I see is that it can't be retrieved

Dylan-Prins:
@josunefon  not going to merged, because of the tests? maybe someone can run them for me?

AlexanderSehr:
> @josunefon not going to merged, because of the tests? maybe someone can run them for me?

Hey @Dylan-Prins,
to chime in here: Yes, pretty much. We have to ensure the tests work before the PR goes in as any bug/failed static test etc. would block any other PR to the module until it is fixed in upstream. 
If you don't have an option to validate the change on your end, we can implement a workaround by merging your PR not directly into main, but an intermediate branch in upstream from where we can test. If any test happens to fail, we'd then need to fix it though.

Anyways, it's a great contribution and should definitely go in. 🚀 We just have to make sure it works in one way or another before it hits main :)

avm/res/compute/virtual-machine/main.bicep
```diff
@@ -536,6 +536,10 @@ resource managedDataDisks 'Microsoft.Compute/disks@2024-03-02' = [
       diskMBpsReadWrite: dataDisk.?diskMBpsReadWrite
       publicNetworkAccess: publicNetworkAccess
       networkAccessPolicy: networkAccessPolicy
+      encryption: {
+        diskEncryptionSetId: dataDisk.managedDisk.?diskEncryptionSetResourceId
+        type: 'EncryptionAtRestWithCustomerKey'
```
AlexanderSehr, line 541: I guess this should rather be something akin to `dataDisk.managedDisk.?type ?? 'EncryptionAtRestWithPlatformKey`'?

If defaulting to `EncryptionAtRestWithCustomerKey` I guess you'd always need to provide a key.

avm/res/compute/virtual-machine/main.bicep
```diff
@@ -536,6 +536,10 @@ resource managedDataDisks 'Microsoft.Compute/disks@2024-03-02' = [
       diskMBpsReadWrite: dataDisk.?diskMBpsReadWrite
       publicNetworkAccess: publicNetworkAccess
       networkAccessPolicy: networkAccessPolicy
+      encryption: {
```
AlexanderSehr, line 539: I wonder if the entire property should be conditional / if the deployment works if no encryption is specified, yet the parameters are passed in as
```bicep
encryption: {
  diskEncryptionSetId: null
  type: null
}
```
No need to change it now, it should just be tested. However in case it may fail we'd need to change it to e.g. 
```bicep
encryption: !empty(dataDisk.managedDisk.?diskEncryptionSetResourceId) ? {
  diskEncryptionSetId: dataDisk.managedDisk.?diskEncryptionSetResourceId
  type: dataDisk.managedDisk.?type
} : null
```
or
```bicep
...(!empty(dataDisk.managedDisk.?diskEncryptionSetResourceId)  ? { encryption: {
  diskEncryptionSetId: dataDisk.managedDisk.?diskEncryptionSetResourceId
  type: dataDisk.managedDisk.?type
} : {})
```
UNLESS, the resource type anyways defaults to 
```bicep
encryption: {
  diskEncryptionSetId: null
  type: 'EncryptionAtRestWithPlatformKey'
}
```
i.e., MS-managed encryption. In that case, the default for `type` should just be `EncryptionAtRestWithPlatformKey` and you're good to go

brianwest:
@Dylan-Prins and @AlexanderSehr, why not just modify the logic for the data disk to mirror the OS disk?

[OS disk diskEncryptionSet](https://github.com/Azure/bicep-registry-modules/blob/a106f9ea4a630f69d2b9acbf45bc01555921f8b6/avm/res/compute/virtual-machine/main.bicep#L584):

```Bicep
diskEncryptionSet: {
  id: osDisk.managedDisk.?diskEncryptionSetResourceId
}
```

[data disk diskEncryptionSet](https://github.com/Azure/bicep-registry-modules/blob/a106f9ea4a630f69d2b9acbf45bc01555921f8b6/avm/res/compute/virtual-machine/main.bicep#L602):

```Bicep
diskEncryptionSet: contains(dataDisk.managedDisk, 'diskEncryptionSet')
  ? {
      id: dataDisk.managedDisk.diskEncryptionSet.id
    }
  : null
```

diskEncryptionSet.id is not valid. Like the OS disk, the property is meant to be diskEncryptionSetResourceId, as defined in the [dataDiskType type definition](https://github.com/Azure/bicep-registry-modules/blob/a106f9ea4a630f69d2b9acbf45bc01555921f8b6/avm/res/compute/virtual-machine/main.bicep#L1224).

I've tested the change on a local copy of the module and mirroring the logic for the OS disk on the data disk encrypts the data disk with a customer-managed key.

AlexanderSehr:
Consistency is usually king - especially consistency from the angle of the consumer. Just to add to the mix, this is the implementation in the latest 'managed disk' module:

https://github.com/Azure/bicep-registry-modules/blob/11709a2f1814cf83998eb50168c08b55a95ff38a/avm/res/compute/disk/main.bicep#L246-L250

which we btw could reference here as a module instead of using a custom resource implementation. 

In any case @brianwest is absolutely correct in that the reference `diskEncryptionSet.id` is not valid as the type for the disk doesn't have it. Instead, it only knows `dataDisk.managedDisk.diskEncryptionSetResourceId`. This bug would be in line `604`. which is currently not touched by this PR

AlexanderSehr:
Closing in favor of #7024 as stale.

PR #5912, opened 2025-09-01
feat: Intoduce 'Oras' as alternative to MCR tags for publishing validation

AlexanderSehr:
## Description

Introducing 'Oras' as alternative validation method which is not only faster (as it directly pulls from the registry), but also more robust.

This PR may further unlock future usages by e.g. comparing to-be versions with already existing latest to validate whether a module should, or should not be published.

## Pipeline
- [Reference run with existing version](https://github.com/Azure/bicep-registry-modules/actions/runs/17385895567/job/49352179205)
- [Reference run with non-existend version](https://github.com/Azure/bicep-registry-modules/actions/runs/17386196568)

## Type of Change

  - [ ] Update to documentation


.github/actions/templates/avm-publishModule/action.yml
AlexanderSehr, line 1: Recommend disabling actions before merging

utilities/pipelines/publish/Confirm-ModuleIsPublished.ps1
```diff
@@ -14,10 +14,22 @@ Mandatory. The path of the module to check for. For example: 'avm/res/key-vault/
 .PARAMETER GitTagName
 Mandatory. The tag name of the module's git tag to check for. For example: 'avm/res/event-hub/namespace/0.2.0'
 
+.PARAMETER ValidationMethod
+Optional. Define how a successful publish should be validated.
+- Oras: A tool to restore repositories directly from a registry endpoint
```
eriqua:
```suggestion
- Oras: A tool to restore repositories directly from a registry endpoint (default)
```

utilities/pipelines/publish/Confirm-ModuleIsPublished.ps1
```diff
@@ -14,10 +14,22 @@ Mandatory. The path of the module to check for. For example: 'avm/res/key-vault/
...
+- Catalogue: A method to ping the MCR data, that is, the Tags-Lists
+
+Note: The 'Catalogue' validation method may take longer to complete than the 'Oras' method.
+
 .EXAMPLE
 Confirm-ModuleIsPublished -Version '0.2.0' -PublishedModuleName 'avm/res/key-vault/vault' -Verbose
 
-Check if module 'key-vault/vault' has been published with version '0.2.0
+Check if module 'key-vault/vault' has been published with version '0.2.0 using the default validation method
```
eriqua:
```suggestion
Check if module 'key-vault/vault' has been published with version '0.2.0 using Oras, the default validation method
```

utilities/pipelines/publish/Confirm-ModuleIsPublished.ps1
```diff
@@ -52,61 +59,108 @@ function Confirm-ModuleIsPublished {
...
+            $index = 0
+
+            $registryReference = 'mcr.microsoft.com/bicep/{0}:{1}' -f $PublishedModuleName, $version
+            while ($true) {
+                $index++
+
+                oras pull $registryReference --format 'json' 2>$null
+
+                if (Test-Path (Join-Path -Path $PWD 'Compiled ARM template')) {
+                    # A very simple check
```
eriqua: could we briefly elaborate on what the check is?

AlexanderSehr: Sure thing :)

When you run the command, you get a file called `Compiled ARM template` (no extension, nothing). However, it **is** in fact the ARM template. To confirm this, I added this check that renames the file to `main.json`, tries to import it (to confirm it's actually valid JSON), find finally confirm that's it's an actual proper ARM template, by checking if it has the property `$schema`:
https://github.com/Azure/bicep-registry-modules/blob/5357029516231233bc1a2847024d7dcb68629eac/avm/res/cache/redis/main.json#L2

eriqua: Thanks for elaborating, what about
```suggestion
                    # Checking if the file "Compiled ARM template" returned by the oras cmdlet is valid
                    # Rename the file to "main.json", import it and check it has the property $schema
```

eriqua (approved):
:ok:

PR #5943, opened 2025-09-08
fix: VirtualMachine - CSE Extension update

### LINKED ISSUE #2413: [Bug Report]: Configuration of CustomScriptExtension (Virtual Machines) is faulty
### Describe the bug

According to [the CustomScriptExtension docs](https://learn.microsoft.com/en-us/azure/virtual-machines/extensions/custom-script-windows), the property "fileUri" should be set in "protectedSettings" under certain conditions, but it can be specified in public settings too. It has to be set in either of those, otherwise there will be an error.

Now this is the bug: the current "CustomScriptExtension" code hardcodes the "fileUri" property to the public settings, meaning you can't set it in the protected settings.

Apart from this issue, the current code is IMHO much too restricting in how you can use this extension. The code should not be too restricting in how you can use the extension, because this keeps users from deploying useful and valid configurations. Please see the "Code snippet" box for the code I am referring to (file is: .../Microsoft.Compute/virtualMachines/deploy.bicep). In my opinion, the code should be just like the one from the preceding "DesiredStateConfiguration" Extension - here, the settings are handled this way:
```bicep
    // ...
    settings: contains(extensionDSCConfig, 'settings') ? extensionDSCConfig.settings : {}
    protectedSettings: contains(extensionDSCConfig, 'protectedSettings') ? extensionDSCConfig.protectedSettings : {}
    // ...
```
So I believe the best implementation for the "CustomScriptExtension" would just be:
```bicep
    // ...
    settings: contains(extensionCustomScriptConfig, 'settings') ? extensionCustomScriptConfig.settings : {}
    protectedSettings: contains(extensionCustomScriptConfig, 'protectedSettings') ? extensionCustomScriptConfig.protectedSettings : {}
    // ...
```

BTW: why does the extension depend on the "DesiredStateConfiguration" extension? there is a "depends_on". you may want to check this too, while you are at it! ;)

### To reproduce

Just see current code: v0.10.0

### Code snippet

```Bicep
module vm_customScriptExtension 'extensions/deploy.bicep' = if (extensionCustomScriptConfig.enabled) {
  name: '${uniqueString(deployment().name, location)}-VM-CustomScriptExtension'
  params: {
    virtualMachineName: vm.name
    name: 'CustomScriptExtension'
    // ...
    settings: {
      fileUris: [for fileData in extensionCustomScriptConfig.fileData: contains(fileData, 'storageAccountId') ? '${fileData.uri}?${listAccountSas(fileData.storageAccountId, '2019-04-01', accountSasProperties).accountSasToken}' : fileData.uri]
    }
    protectedSettings: extensionCustomScriptProtectedSetting
    // ...
  }
  dependsOn: [
    vm_desiredStateConfigurationExtension
  ]
}
```

### Relevant log output

_No response_

### LINKED ISSUE #5929: [AVM Module Issue]: AVM virtual machine module breaks when fileUris is defined in extensionCustomScriptProtectedSetting

### Issue Type?

Bug

### Module Name

avm/res/compute/virtual-machine

### (Optional) Module Version

v0.18.0

### Description

Problem :

I already use extensionCustomScriptProtectedSetting.fileUris to securely access scripts stored in a Storage Account with private endpoint/VNET access (no anonymous access).

In AVM v0.15, when setting:
```bicep
extensionCustomScriptConfig: {
  enabled: true
  fileData: []
}
```
the module correctly compiled to: "settings": { }

This allowed us to rely only on protectedSettings.fileUris as documented by Microsoft.

From v0.16 - v0.18, even when fileData: [ ], the module now generates:

```bicep
"settings": {
  "fileUris": []
}
```

This breaks deployments because the Custom Script Extension rejects having fileUris present in both settings and protectedSettings.

Reference to code:- https://github.com/Azure/bicep-registry-modules/blob/7964b63fd4cdcc29503ffaac21b460a84713e267/avm/res/compute/virtual-machine/extension/main.bicep?utm_source=chatgpt.com#L60

I have already deployed avm version 15 with CSE and fileData has empty array now i was updating a version from 15-18 but need to configure the Custom Script Extension so that **fileUris** are provided only in **protectedSettings**, which is required when scripts are stored in a Storage Account with private endpoint/VNET access (no anonymous access).

Impact : 

Cannot upgrade from v0.15 → v0.16/0.17/0.18 while keeping secure configuration.
Causes deployment validation/runtime errors (Error code: 22).
Even leaving fileData: [] results in invalid settings: { fileUris: [] }.

Image

### (Optional) Correlation Id

_No response_

AlexanderSehr:
### Changes

- Introduced user-defined type for `extensionCustomScriptConfig` parameter that is aligned with its [official documentation](https://learn.microsoft.com/en-us/azure/virtual-machines/extensions/custom-script-windows).
- Added parameters `provisionAfterExtensions` & `provisionAfterExtensions`  to 'extension' child module.
- Updated `Set-BlobContent` used by tests to work for both Linux & Windows (failed for Linux until now)

### Breaking Changes

- Merged `extensionCustomScriptProtectedSetting` parameter into `extensionCustomScriptConfig`
- Removed support for the CustomScriptExtension extension to automatically append SAS-Keys to file specified via the `extensionCustomScriptConfig.fileData` property. Instead, the SAS token must either be pre-provided with the URL, or either the settings `extensionCustomScriptConfig.protectedSettings.storageAccountKey` & `extensionCustomScriptConfig.protectedSettings.storageAccountName` or (recommended) `extensionCustomScriptConfig.protectedSettings.managedIdentityResourceId`. For the latter, you can provide either the full resource ID - or set it to `''` if you want it to use the VM's system-assigned identity (if enabled) instead. Note, in either case, the Identity must be granted access to correct Storage Account scope.

Closes #5929

## Pipeline Reference

| Pipeline |
| -------- |
[avm.res.compute.virtual-machine](https://github.com/Azure/bicep-registry-modules/actions/workflows/avm.res.compute.virtual-machine.yml)

## Type of Change

  - [ ] Update to documentation


josunefon (dismissed):

josunefon (dismissed):

avm/res/compute/virtual-machine/CHANGELOG.md
AlexanderSehr, line 1: Disable actions before merging

ReneHezser (approved):
LGTM. Thank you.

PR #5944, opened 2025-09-08
fix: DevOps CI/CD agents pattern module DNS properties were not used and added `dnsIsDeployedByPolicy` property

johnlokerse:
## Description

Ran into some DNS issues when bringing your own DNS setup and when using DINE policies. This pull request removes `privateDnsZoneGroupName` and `privateDnsZoneResourceIds` in module `acr`. These properties were not used. Updated them to the correct properties as defined in the UDT. Additionally, added the parameter `dnsIsDeployedByPolicy` to be able to skip the configuration of privateDnsZoneGroups. 

Updated modules to the latest version.

## Pipeline Reference

Do you have an environment set up @sebassem so this change can be tested?

## Type of Change

  - [x] Update to documentation

## Checklist


sebassem (dismissed):
LGTM

jtracey93 (dismissed):

jtracey93:
@johnlokerse can you resolve the merge conflicts

johnlokerse:
@jtracey93 Done, but I have some issues with generating the README via `Set-AVMModule` command. This pattern module does not have a `waf-aligned` e2e test but the command creates it, resulting in errors because the generated waf-aligned tests have module errors. See console output:

CleanShot 2025-10-31 at 09 15 10@2x

avm/ptn/dev-ops/cicd-agents-and-runners/main.bicep
```diff
@@ -282,7 +282,7 @@ module acr 'br/public:avm/res/container-registry/registry:0.9.3' = {
           {
             subnetResourceId: networkingConfiguration.networkType == 'createNew'
               ? filter(
-                  newVnet!.outputs.subnetResourceIds,
+                  newVnet.?outputs.subnetResourceIds ?? [],
```
AlexanderSehr, line 285: Hey @johnlokerse,
I'd be curious why change this to a tryGet() if the case is only execute if the `networkType == 'createNew'`? Wouldn't the `!` (i.e., it won't be `null`) be more accurate (yet alone requiring the fallback)? Or is Bicep executing the function regardless, hence leading to errors.

AlexanderSehr: The same question applies to the NAT GW etc.

johnlokerse: Hey @AlexanderSehr, suppressing would be fine here due to the conditions, and we know `newVnet` has a value anyway so there is no need to use any safe-dereferencing here. Same thing for NAT GW. Btw, I didn't do any validation, my answer is based on theory.. 😄

johnlokerse: Also, this is not my change so I don't know the reasoning behind it

AlexanderSehr: How odd. It showed it up as your change before, but now not anymore. Let me resolve the comment then (even though my point still stands @sebassem). Both works eventually though, so it's nothint to prioritize.

AlexanderSehr:
> @jtracey93 Done, but I have some issues with generating the README via `Set-AVMModule` command. This pattern module does not have a `waf-aligned` e2e test but the command creates it, resulting in errors because the generated waf-aligned tests have module errors. See console output:
> 
> CleanShot 2025-10-31 at 09 15 10@2x

Regarding this @johnlokerse, please run the script with the corresponding skips. For example:
`Set-AvmModule -ModuleFolderPath '<path>' -SkipFileAndFolderSetup -ThrottleLimit 10`.

Maybe it's time to flip this skip-switch parameter on its head.

johnlokerse:
> Set-AvmModule -ModuleFolderPath '<path>' -SkipFileAndFolderSetup -ThrottleLimit 10

That works! Pushed changes.

AlexanderSehr:
Closing in favor of #7030 

@johnlokerse you'll remain the author of the changes as I just rebased the PR to resolve the conflicts.

PR #6013, opened 2025-09-26
feat: add `serviceGroupTargetResourceIds` to utl common types module

jtracey93:
## Description

Add `serviceGroupTargetResourceIds` to utl common types module as per discussion https://github.com/Azure/Azure-Verified-Modules/issues/2324

## Pipeline Reference

| Pipeline |
| -------- |
|  https://github.com/jtracey93/bicep-registry-modules/actions/runs/18043889300 | 

## Type of Change

  - [ ] Update to documentation

## Checklist

> Developed in Lanzarote 🚢


jtracey93:
@AlexanderSehr would appreciate your 👀 on this one

avm/utl/types/avm-common-types/main.bicep
```diff
@@ -431,3 +431,10 @@ type secretsOutputType = {
   @description('An exported secret\'s references.')
   *: secretSetOutputType
 }
+
+// ================== //
+//   Service Groups   //
+// ================== //
+@export()
+@description('An AVM-aligned type for the target resource IDs to associate resource to service groups.')
+type serviceGroupTargetResourceIdsType = resourceInput<'Microsoft.Relationships/serviceGroupMember@2023-09-01-preview'>.properties.targetId
```
AlexanderSehr, line 440: Hey @jtracey93, thanks fir the PR(s) 🚀
Regarding this one: I wonder - what 's the point of the type if it's effectively a `string`?  In other words, what's the type's added value compared to a module's individual implementation via a `string` parameter?

jtracey93: I was thinking similar, however I thought I’d raise the PR for us to discuss.

I think centralising it here in the UTL module with an RDT ensures that valid resource IDs are provided for they’d import as its not the typical resource ID, e.g. `/subscriptions/xxxxx/resourceGroups/xxxx/RP/type/xxx` as its a tenant level input `/providers/Microsoft.Management/serviceGroups/xxx`

Thoguhts?

AlexanderSehr: I think this would only be possible in Terraform as Bicep does not have a feature like regex validation and the like. In other words, the `resourceInput` in this case would only check that it's a string - that 'should' be it.
I'd suggest an alternative which is to enforce `string` as a param type via our Pester tests if anybody defines a `serviceGroupTargetResourceId` parameter.
I'll bring this topic up in our next team call.

jtracey93:
closing and will add as pester check in static validation for `string[]` for `serviceGroupTargetResourceIds`

PR #6086, opened 2025-10-15
feat: `Microsoft.Communication/emailServices` Add output for the domain verification records and bump API versions

thecmdradama:
## Description

* Adds output for `domainVerificationRecords`
* Updates `Microsoft.Communication/emailServices` API version from `2023-04-01` to `2025-05-01`
* Updates `Microsoft.Communication/emailServices/domains` API version from `2023-04-01` to `2025-05-01`
* Updates `Microsoft.Communication/emailServices/domains/senderUsernaems` API version from `2023-04-01` to `2025-05-01`

## Pipeline Reference

| Pipeline |
| -------- |
[avm.res.communication.communication-service](https://github.com/thecmdradama/bicep-registry-modules/actions/workflows/avm.res.communication.communication-service.yml)

## Type of Change

  - [x] Update to documentation

## Checklist


thecmdradama:
I was thinking of bundling in the functionality to create the `Domain`, `SOA`, `DKIM`, and `DKIM2` DNS records on a public DNS Zone for which you pass in a DNS Zone ResourceId, however that then pivots to this becoming a more complex pattern rather than a resource module... Keen to get thoughts on that approach..

avm/res/communication/email-service/domain/main.bicep
```diff
@@ -138,6 +138,9 @@ output resourceId string = domain.id
 @description('The name of the resource group the domain was created in.')
 output resourceGroupName string = resourceGroup().name
 
+@description('The verification records for the domain.')
+output verificationRecords verificationRecordsOutputType = domain.properties.verificationRecords
```
AlexanderSehr:
```suggestion
output verificationRecords resourceOutput<'Microsoft.Communication/emailServices/domains@2025-05-01'>.properties.verificationRecords = domain.properties.verificationRecords
```
This **should** also work and avoid adding & maintaining a custom type. Could you give it a shot?

AlexanderSehr:
That's a great PR @thecmdradama 💪 

Thank you very much. @donk-msft please also give it a look when you get the chance :)

Off-topic @thecmdradama: Please run the [platform.toggle-avm-workflows.yml](https://github.com/thecmdradama/bicep-registry-modules/actions/workflows/platform.toggle-avm-workflows.yml) workflow in your fork once. It disables all individual workflows so that they're not automatically triggered when anything in your main branch changes. This does not apply for newly added workflows (which would require a re-run), but it goes a long way. It's also documented in the Contribution Guide [here](https://azure.github.io/Azure-Verified-Modules/contributing/bicep/bicep-contribution-flow/enable-or-disable-workflows/).

avm/res/communication/email-service/domain/main.bicep
thecmdradama: Ah awesome, thanks for that suggestion and agreed that is a much better approach.

thecmdradama:
> That's a great PR @thecmdradama 💪
> 
> Thank you very much. @donk-msft please also give it a look when you get the chance :)
> 
> Off-topic @thecmdradama: Please run the [platform.toggle-avm-workflows.yml](https://github.com/thecmdradama/bicep-registry-modules/actions/workflows/platform.toggle-avm-workflows.yml) workflow in your fork once. It disables all individual workflows so that they're not automatically triggered when anything in your main branch changes. This does not apply for newly added workflows (which would require a re-run), but it goes a long way. It's also documented in the Contribution Guide [here](https://azure.github.io/Azure-Verified-Modules/contributing/bicep/bicep-contribution-flow/enable-or-disable-workflows/).

Also thank you for that suggestion. Will run now. Ta

donk-msft (approved):

AlexanderSehr:
Hey @thecmdradama,
sadly the static tests failed in upstream as the module files were not regenerated with some later changes: [ref](https://github.com/Azure/bicep-registry-modules/actions/runs/18587240533). 
I'll update a PR on your behalf to address this though. Should ideally go in today then.

Regarding your ask 
> I was thinking of bundling in the functionality to create the Domain, SOA, DKIM, and DKIM2 DNS records on a public DNS Zone for which you pass in a DNS Zone ResourceId, however that then pivots to this becoming a more complex pattern rather than a resource module... Keen to get thoughts on that approach..

I can't really say it's right or wrong. In general it is said that resource modules should deploy only itself or **strongly** correlated resources (e.g., private endpoints, child modules, sometimes a public Ip). The reason it's somewhat restrictive is that every feature that's added to a module will be part of it for all users. While that is great in general, it also contributes to a template's size and the 4mb limit is sadly still very real. Because of that (and to keep the `res` modules somewhat focused), every decision to add anything extra must be on a case-by-case basis. For this module specifically, the question to ask is whether the creation of the records is something that people would usually always do (i.e., it's a standard case), or, if it's only relevant for a small subnet of users. If it's the former, there's a case to be made, if the latter, it's probably something better implemented in a `ptn` module. I hope that made sense :) @donk-msft, happy to year your thoughts as well.

thecmdradama:
> Hey @thecmdradama, sadly the static tests failed in upstream as the module files were not regenerated with some later changes: [ref](https://github.com/Azure/bicep-registry-modules/actions/runs/18587240533). I'll update a PR on your behalf to address this though. Should ideally go in today then.

Ah bugger, thanks for fixing that up @AlexanderSehr 
> 
> Regarding your ask
> 
> > I was thinking of bundling in the functionality to create the Domain, SOA, DKIM, and DKIM2 DNS records on a public DNS Zone for which you pass in a DNS Zone ResourceId, however that then pivots to this becoming a more complex pattern rather than a resource module... Keen to get thoughts on that approach..
> 
> I can't really say it's right or wrong. In general it is said that resource modules should deploy only itself or **strongly** correlated resources (e.g., private endpoints, child modules, sometimes a public Ip). The reason it's somewhat restrictive is that every feature that's added to a module will be part of it for all users. While that is great in general, it also contributes to a template's size and the 4mb limit is sadly still very real. Because of that (and to keep the `res` modules somewhat focused), every decision to add anything extra must be on a case-by-case basis. For this module specifically, the question to ask is whether the creation of the records is something that people would usually always do (i.e., it's a standard case), or, if it's only relevant for a small subnet of users. If it's the former, there's a case to be made, if the latter, it's probably something better implemented in a `ptn` module. I hope that made sense :) @donk-msft, happy to year your thoughts as well.

Awesome explanation Alex, thanks. That's one of the thoughts that was making me err on the side of not adding in the extra functionality... For a portion of people consuming this module, I'd expect they will also be managing their public DNS records also within Azure. The other thing that made me reconsider, is the separate zone configuration scenarios referenced in the docs... [ref.](https://learn.microsoft.com/en-us/azure/communication-services/quickstarts/email/add-custom-verified-domains?pivots=platform-azp#configure-sender-authentication-for-custom-domain) This would then require some transformation of the default values to create the records for higher level domains. 

With all that in mind, yeah I do believe that logic should go in a `ptn` module instead.

AlexanderSehr:
Interesting. I checked now already 2 subscriptions and neither support the updated API version, yes Bicep itself does suggest it and it seemed to have work for you  @thecmdradama. In any case, it upstream it does not, so we may need to change it to the preview version

AlexanderSehr:
So, I'm not sure what's going on, but the used API version does not exist in any environment I have access to, and the latest that's document in the Bicep types and works for us is the original. I hence needed to roll the api version update back but kept everything else (most notably the added outputs): #6100 

Why this issue exists is something to further investigate. Can't say I've seen anything like this before. For reference, that's what the Bicep types know as versions:
image

and this is what our environments know:
- (...), 
- 2023-04-01, 
- 2023-06-01-preview, 
- 2024-11-01-preview, 
- 2025-01-25-preview

Not much of an overlap there

AlexanderSehr:
Quick update: @thecmdradama, `0.4.1` is now published

thecmdradama:
Hah, I'm going to stick my hand up and admit I'm an idiot here... I was running https://github.com/thecmdradama/bicep-registry-modules/actions/workflows/avm.res.communication.communication-service.yml not https://github.com/thecmdradama/bicep-registry-modules/actions/workflows/avm.res.communication.email-service.yml... I was testing against the wrong workflow... According to the portal. `Microsoft.Communication/communicationServices@2025-05-01` does exist but I hadn't touched that resource module at all.. 
image

That is why my deployment tests succeeded -_-. I also see the same API versions available for `emailServices` `emailServices/domains` in my tenant that you were seeing. 
image

Still incredibly strange that the api version `2025-05-01` is supposedly valid according to the resource reference docs.

https://learn.microsoft.com/en-us/azure/templates/microsoft.communication/2025-05-01/emailservices?pivots=deployment-language-bicep
https://learn.microsoft.com/en-us/azure/templates/microsoft.communication/2025-05-01/emailservices/domains?pivots=deployment-language-bicep
https://learn.microsoft.com/en-us/azure/templates/microsoft.communication/2025-05-01/emailservices/domains/senderusernames?pivots=deployment-language-bicep

thecmdradama:
> Quick update: @thecmdradama, `0.4.1` is now published

Absolute legend, thank you. Monday problem to incorporate the new outputs in the pattern I'm building.

AlexanderSehr:
> > Quick update: @thecmdradama, `0.4.1` is now published
> 
> Absolute legend, thank you. Monday problem to incorporate the new outputs in the pattern I'm building.

Glad to hear. :) I hope it's going well 💪

PR #6134, opened 2025-10-26
fix: Databricks - Fixed issue with having seperate keys for encrpytion & disk-encryption in the same vault

### LINKED ISSUE #6002: [AVM Module Issue]: Databricks customerManagedKeyManagedDisk version uses the same version as customerManagedKey

### Issue Type?

Bug

### Module Name

avm/res/databricks/workspace

### (Optional) Module Version

0.11

### Description

When using the same keyvault but separate unique keys for customerManagedKey and customerManagedKeyManagedDisk parameters the templates logic to retrieve the key version causes the customerManagedKeyManagedDisk encryption to use the key version of the customerManagedKey instead of its own key version.

I believe the issue is at line #349 

`: cMKKeyVault::cMKKey!.properties.keyUriWithVersion,`

this fallback logic for the managedDisk encryption configuration is using cMKKey which is defined earlier on line 204 as the key for customerManagedKey 

I spotted this when we deployed databricks and interestingly it did not throw an error, despite the databricks encryption GUI and the databricks JSON object showing the key version as the same for both CMKs using their unique key names.

### (Optional) Correlation Id

_No response_

AlexanderSehr:
## Description

- Fixed an issue where having seperate keys for encrpytion & disk-encryption in the same vault would not work as expected.
- Added test case to validate this setup specifically 

Closes #6002

- [x] Test with 2 keys in 1 vault
  image
- [x] Test 1 key in 1 vault
  image
- [x] Test with 2 keys in 2 vaults
  image

## Pipeline Reference

| Pipeline |
| -------- |
[avm.res.databricks.workspace](https://github.com/Azure/bicep-registry-modules/actions/workflows/avm.res.databricks.workspace.yml)

## Type of Change

  - [ ] Update to documentation

### DIFF
--- a/.github/workflows/avm.res.databricks.workspace.yml
+++ b/.github/workflows/avm.res.databricks.workspace.yml
@@ -84,7 +84,8 @@ jobs:
     with:
       workflowInput: "${{ needs.job_initialize_pipeline.outputs.workflowInput }}"
       # moduleTestFilePaths: "${{ needs.job_initialize_pipeline.outputs.moduleTestFilePaths }}"
-      moduleTestFilePaths: '[{"path":"tests/e2e/encr/main.test.bicep","name":"encr","e2eIgnore":false}]'
+      # moduleTestFilePaths: '[{"path":"tests/e2e/encr/main.test.bicep","name":"encr","e2eIgnore":false}]'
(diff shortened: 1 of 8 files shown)


ReneHezser (approved):
Thanks Alex. Looks great :-)

PR #6274, opened 2025-11-26
feat: `avm/res/container-registry/registry` - Add `autoGeneratedDomainNameLabelScope` and `roleAssignmentMode` params

### LINKED ISSUE #6273: [AVM Module Issue]: Add `roleAssignmentMode` and `autoGeneratedDomainNameLabelScope` to ACR module

### Issue Type?

Feature Request

### Module Name

avm/res/container-registry/registry

### (Optional) Module Version

_No response_

### Description

The ACR module lacks the parameters `roleAssignmentMode` and `autoGeneratedDomainNameLabelScope`, which are readily available when creating a registry via the portal.

### (Optional) Correlation Id

_No response_

cecheta:
## Description

* Add `autoGeneratedDomainNameLabelScope` and `roleAssignmentMode` params
* Add new ABAC roles

Closes #6273

## Pipeline Reference

| Pipeline |
| -------- |
| [avm.res.container-registry.registry](https://github.com/cecheta/bicep-registry-modules/actions/workflows/avm.res.container-registry.registry.yml?query=branch%3Acontainer-registry) |

## Type of Change

  - [ ] Update to documentation

## Checklist


avm/res/container-registry/registry/tests/e2e/defaults/main.test.bicep
```diff
@@ -43,7 +57,13 @@ module testDeployment '../../../main.bicep' = [
...
       name: '${namePrefix}${serviceShort}001'
       location: resourceLocation
-      acrSku: 'Standard'
+      publicNetworkAccess: 'Disabled'
+      replications: [
+        {
+          name: nestedDependencies.outputs.pairedRegionName
+          location: nestedDependencies.outputs.pairedRegionName
+        }
+      ]
```
AlexanderSehr: The default test should only test the bare minimum, i.e., required parameters. As this PR only adds to optional parameters, please remove the added values

cecheta: The default test was failing PsRule so I was attempting to get it to pass.
I guess the better solution would be to update the module itself so that the default test passes, but not sure how to set a default geo-replication.
I will leave this PR in draft until the existing module is fixed.

AlexanderSehr: Hey @cecheta,
thanks for elaborating. I see the last run in `main` passed ([ref](https://github.com/Azure/bicep-registry-modules/actions/runs/17158809127)) which is curious. Let me take a closer look at the PR later again to try and understand why this may have changed.
In any case, regarding PSRule you're right, for the most part. There are actually 3 paths:
1. Either, if we're talking about parameters that can have a WAF-aligned defaults (e.g., a string of an allowed set, a boolean value etc.), this value should be set to that compliant default. Absolutely correct
2. Then there are the others, that need actual user input (e.g., tags would be recommended, but can't be assumed on behalf of the user). For those cases we don't define a default and actual disable that rule specifically for default tests via this file: [min-suppress.Rule.yaml](https://github.com/Azure/bicep-registry-modules/blob/main/utilities/pipelines/staticValidation/psrule/.ps-rule/min-suppress.Rule.yaml).
3. Some rules may not make any sense. E.g., I've seen rules that you should use a managed identity for authentication. While this makes sense on the surface (as aposed to credential-based authentication), it's even better to not even define a managed-identity if don't need any authentication. Hence we disable that rule usually as well.

I hope this helped to set the context :)

AlexanderSehr: Looks like it was just unlucky timing in that the rule was introduced in September and as you're the first updating the module since, it failed for you.

As described above, @cecheta, we should add the failed `Azure.ACR.GeoReplica` test to the suppression rules for the defaults tests. To do so, just navigate to that file in your branch, add a new line for `Azure.ACR.GeoReplica` and add a comment as to why it should not be enforced (i.e., `# Excluded as it requires user input`).

After, just re-run the static tests again and they should pass 💪

cecheta: Thanks @AlexanderSehr , it's passing now

AlexanderSehr:
Nothing more to add from my side. Let's give @JPEasier a chance to review the PR as he's the module owner and I'll then approve the changes to the PSRule config file. Thanks @cecheta, a pleasure as always 💪

JPEasier (approved):
Looks good for me 👍

AlexanderSehr (approved):

PR #6322, opened 2025-12-08
fix: #6295 - Deployment fails when deployment().location contains spaces (e.g., 'West Europe', 'East US 2')

### LINKED ISSUE #6295: [AVM Module Issue]: ptn sub-vending fails when deployment location is set to a region containing a space

### Issue Type?

Bug

### Module Name

avm/ptn/lz/sub-vending

### (Optional) Module Version

0.4.1

### Description

Running below pwsh command results in an exception.

```PowerShell
New-AzDeployment -Name test -Location 'West Europe' -TemplateParameterFile C:\Git\subscription-vending\src\bicep\sub-vending.bicepparam -DeploymentDebugLogLevel All -WhatIf -SkipTemplateParameterPrompt
```

### Exception:
```PowerShell
New-AzDeployment:
DeploymentWhatIfResourceError - Long running operation failed with status 'Failed'. Additional Info:'The request to predict template deployment changes to scope '/subscriptions/00000000-0000-0000-0000-000000000000' has failed due to a resource error. See details for more information.'
InvalidResourceGroup - The provided resource group name 'rsg-West Europe-identities' has these invalid characters: ' '. The name can only be a letter, digit, '-', '.', '(', ')' or '_'. For more details, visit https://aka.ms/ResourceGroupNamingRestrictions .
```

### Possible fix (line 120):
```Bicep
param userAssignedIdentityResourceGroupName string = 'rsg-${replace(deployment().location, ' ', '')}-identities'
```

### (Optional) Correlation Id

_No response_

danieletten:
## Description

Fixes #6295 - Deployment fails when `deployment().location` contains spaces (e.g., 'West Europe', 'East US 2')

### Problem
Azure location names with spaces (like 'West Europe') are valid for the `location` property, but when interpolated into resource group names, they violate naming constraints which don't allow spaces.

### Solution
Implemented location normalization pattern (following the approach used in `avm/ptn/network/private-link-private-dns-zones`):
- Added 73-region mapping dictionary to convert display names with spaces to lowercase short names
- Created normalized variables for 8 parameters that use `deployment().location` in their default values
- Parameters remain unchanged (still use `deployment().location` as default)
- Variables conditionally normalize: only if parameter equals default value, otherwise respect custom values
- Module calls always use normalized variables

### Changes
- Added `azureRegionShortNameDisplayNameAsKey` mapping dictionary (73 Azure regions)
- Added `locationLoweredAndSpacesRemoved` normalization logic
- Normalized 8 resource name parameters:
  - `userAssignedIdentityResourceGroupName`
  - `virtualNetworkLocation`
  - `deploymentScriptResourceGroupName`
  - `deploymentScriptName`
  - `deploymentScriptManagedIdentityName`
  - `deploymentScriptVirtualNetworkName`
  - `deploymentScriptNetworkSecurityGroupName`
  - `deploymentScriptLocation`

### Testing
- ✅ All 194 Pester tests passed
- ✅ Successfully deployed with `--location "West Europe"` (spaces and capitals)
- ✅ Backward compatible - no breaking changes

## Pipeline Reference

Not possible, due to complex landingzone deployment and no rights to create subscriptions

## Type of Change

  - [x] Update to documentation

## Checklist

- [] My corresponding pipelines / checks run clean and green without any errors or warnings

### DIFF
--- a/avm/ptn/lz/sub-vending/main.bicep
+++ b/avm/ptn/lz/sub-vending/main.bicep
@@ -117,7 +117,7 @@ param virtualNetworkResourceGroupName string = ''
 
 @maxLength(90)
 @description('Optional. The name of the resource group to create the user-assigned managed identities in.')
-param userAssignedIdentityResourceGroupName string = 'rsg-${toLower(replace(deployment().location, ' ', ''))}-identities'
+param userAssignedIdentityResourceGroupName string = 'rsg-${deployment().location}-identities'
 
 @description('''Optional. An object of Tag key & value pairs to be appended to the Resource Group that the Virtual Network is created in.
 
@@ -131,7 +131,7 @@ param virtualNetworkResourceGroupLockEnabled bool = true
 
 @description('''Optional. The location of the virtual network. Use region shortnames e.g. `uksouth`, `eastus`, etc. Defaults to the region where the ARM/Bicep deployment is targeted to unless overridden.
 ''')
-param virtualNetworkLocation string = toLower(replace(deployment().location, ' ', ''))
+param virtualNetworkLocation string = deployment().location
 
 @minLength(2)
 @maxLength(64)
@@ -289,20 +289,20 @@ param pimRoleAssignments pimRoleAssignmentTypeType[] = []
 param enableTelemetry bool = true
 
 @description('Optional. The name of the resource group to create the deployment script for resource providers registration.')
-param deploymentScriptResourceGroupName string = 'rsg-${toLower(replace(deployment().location, ' ', ''))}-ds'
+param deploymentScriptResourceGroupName string = 'rsg-${deployment().location}-ds'
 
 @description('Optional. The name of the deployment script to register resource providers.')
-param deploymentScriptName string = 'ds-${toLower(replace(deployment().location, ' ', ''))}'
+param deploymentScriptName string = 'ds-${deployment().location}'
 
 @description('Optional. The name of the user managed identity for the resource providers registration deployment script.')
-param deploymentScriptManagedIdentityName string = 'id-${toLower(replace(deployment().location, ' ', ''))}'
+param deploymentScriptManagedIdentityName string = 'id-${deployment().location}'
 
 @maxLength(64)
 @description('Optional. The name of the private virtual network for the deployment script. The string must consist of a-z, A-Z, 0-9, -, _, and . (period) and be between 2 and 64 characters in length.')
-param deploymentScriptVirtualNetworkName string = 'vnet-ds-${toLower(replace(deployment().location, ' ', ''))}'
+param deploymentScriptVirtualNetworkName string = 'vnet-ds-${deployment().location}'
 
 @description('Optional. The name of the network security group for the deployment script private subnet.')
-param deploymentScriptNetworkSecurityGroupName string = 'nsg-ds-${toLower(replace(deployment().location, ' ', ''))}'
+param deploymentScriptNetworkSecurityGroupName string = 'nsg-ds-${deployment().location}'
 
 @description('Optional. The address prefix of the private virtual network for the deployment script.')
 param virtualNetworkDeploymentScriptAddressPrefix string = '192.168.0.0/24'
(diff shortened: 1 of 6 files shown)


danieletten:
Deployment: 
image

Pester:
image

avm/ptn/lz/sub-vending/main.bicep
```diff
@@ -427,6 +427,110 @@ var deploymentNames = {
...
+  'indonesia central': 'indonesiacentral'
+}
+
+var locationLowered = toLower(deployment().location)
+var locationLoweredAndSpacesRemoved = contains(locationLowered, ' ')
+  ? azureRegionShortNameDisplayNameAsKey[locationLowered]
+  : locationLowered
+
+// Normalized resource names - replaces deployment().location in default values to remove spaces and capitals
+var userAssignedIdentityResourceGroupNameNormalized = userAssignedIdentityResourceGroupName == 'rsg-${deployment().location}-identities'
```
jtracey93:
```suggestion
var userAssignedIdentityResourceGroupNameNormalized = userAssignedIdentityResourceGroupName == userAssignedIdentityResourceGroupName
```

Could this be used instead to reduce the number of places that the default value of the parameter needs to be maintained? As with this approach you'd need to update the parameter default value and then also the value here - and thats just more to maintain and something that is easily missed 😢 

Might be worth trying and testing - if it works repeat for all the variables youve added here 👍

avm/ptn/lz/sub-vending/main.bicep
```diff
@@ -427,6 +427,110 @@ var deploymentNames = {
...
+  ? 'rsg-${locationLoweredAndSpacesRemoved}-identities'
+  : userAssignedIdentityResourceGroupName
+
+var virtualNetworkLocationNormalized = virtualNetworkLocation == deployment().location
+  ? locationLoweredAndSpacesRemoved
+  : virtualNetworkLocation
```
jtracey93: is this correct, as this will always be true in a default scenario as it will you are checking `virtualNetworkLocation` == itself?

You probably need to check that `virtualNetworkLocation` doesnt contain spaces, thats really when you then need to invoke the replacement with a lowercase name etc.

avm/ptn/lz/sub-vending/main.bicep
```diff
@@ -427,6 +427,110 @@ var deploymentNames = {
   )
 }
 
+var azureRegionShortNameDisplayNameAsKey = {
```
jtracey93, line 428: nit pick (feel free to ignore) but could we order this alphabetically?

jtracey93 (changes requested):
Thanks @danieletten a few comments

avm/ptn/lz/sub-vending/tests/e2e/defaults/main.test.bicep
```diff
@@ -34,6 +34,16 @@ module testDeployment '../../../main.bicep' = {
     subscriptionWorkload: 'Production'
     subscriptionManagementGroupAssociationEnabled: true
     subscriptionManagementGroupId: 'bicep-lz-vending-automation-child'
+    // Test location normalization with location containing spaces and capitals
+    virtualNetworkEnabled: true
+    virtualNetworkLocation: 'UK South' // Will be normalized to 'uksouth'
+    virtualNetworkResourceGroupName: 'rsg-UK South-vnets' // Will have spaces removed
```
jtracey93: @AlexanderSehr we want to test this to prove our logic for replacing ` ` (spaces) in inputs are replaced to get successfull deployment, but in the resulting readme - it just looks odd - even though it will work.

Is there any way to get comments pulled across into the README in the example sections?

AlexanderSehr: Hola @jtracey93,
I hear you, but the short answer is _no, - or at least not anytime soon_.

The long answer is: Maybe in the future, but it would require is to either rewrite how the 'usage examples' in the readme are created - or we find a smart way to match comments to their corresponding lines in the generated content after the fact.
The reason this is so complicated is because we convert the deployent example into multiple different formats (e.g., a module deployment, using a bicep param file and a JSON param file). The foundation for each is a compiled version of the params block (i.e., JSON), which does not play with comments at all. That's the stage where they are removed.
So, if we want to change this approach we'd need to really do this on the fundamental level and rethink our approach. If I'm not mistaken, JSON was used (since the CARML days) as the basis as it allows us to process the data in a native way (e.g., iterate through the parameters, order them, differentiate required vs optional, ...).
It can certainly be done - but it's far from a quick win.
If we can avoid it by somehow remarring the comments with their corresponding processed lines after the fact we may have an easier time. Still not a quick win, but certainly quicker 😄

jtracey93:
Thank you for your contribution @danieletten!

Due to CI/CD limitations with fork PRs and secrets access, I've migrated your changes to #6335. All your commits and authorship have been preserved. Please follow the new PR for any updates or requested changes.

The new PR will run all required checks and can be merged once approved.

PR #6464, opened 2025-12-30
fix: CI - Corrected test wording

AlexanderSehr:
## Description

Corrected the test's 'Conditional parameters' & UDT's description should contain 'Required if' followed by the condition making the parameter required.' _because_ block as it currently asks to **not** specify the reason (which is not only incorrect, but also conflicts with the test's title).

## Type of Change

  - [ ] Update to documentation

### DIFF
--- a/utilities/pipelines/staticValidation/compliance/module.tests.ps1
+++ b/utilities/pipelines/staticValidation/compliance/module.tests.ps1
@@ -969,7 +969,7 @@ Describe 'Module tests' -Tag 'Module' {
                         }
                     }
                 }
-                $incorrectParameters | Should -BeNullOrEmpty -Because ('the description of conditional parameters in the template file must contain a sentence that  starts with "Required." to explain ''when'' the parameter is mandatory. Found incorrect items: [{0}].' -f ($incorrectParameters -join ', '))
+                $incorrectParameters | Should -BeNullOrEmpty -Because ('the description of conditional parameters in the template file must contain a sentence that starts with "Required." to explain ''when'' the parameter is mandatory. Found incorrect items: [{0}].' -f ($incorrectParameters -join ', '))
             }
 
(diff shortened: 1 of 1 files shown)


utilities/pipelines/staticValidation/compliance/module.tests.ps1
AlexanderSehr, line 1: Actions should be disabled before merging

utilities/pipelines/staticValidation/compliance/module.tests.ps1
```diff
@@ -969,7 +969,7 @@ Describe 'Module tests' -Tag 'Module' {
                         }
                     }
                 }
-                $incorrectParameters | Should -BeNullOrEmpty -Because ('conditional parameters in the template file should lack a description that starts with "Required.". Found incorrect items: [{0}].' -f ($incorrectParameters -join ', '))
+                $incorrectParameters | Should -BeNullOrEmpty -Because ('the description of conditional parameters in the template file must contain a sentence that starts with "Required." to explain ''when'' the parameter is mandatory. Found incorrect items: [{0}].' -f ($incorrectParameters -join ', '))
```
eriqua:
```suggestion
                $incorrectParameters | Should -BeNullOrEmpty -Because ('the description of conditional parameters in the template file must contain a sentence that starts with "Required if", explaining when the parameter is mandatory. Found incorrect items: [{0}].' -f ($incorrectParameters -join ', '))
```

AlexanderSehr: I'd commit if you add a `,` after the `"Required if"` since the 'to' was removed.

eriqua: Good point and updated

eriqua (approved):
🆗

PR #6492, opened 2026-01-11
feat: CI - Adjusted workflow conditions to avoid auto-triggering workflows in `main`

AlexanderSehr:
## Description

- Added condition that avoids auto-triggered workflows in forks for pushes into the main branch
- Aligned disabled platform workflows for forks (some had it some didn't). Excluded PR title workflows as one may one to use those in forks

### Tests

In fork
- [x] (in main) push to main should not run the workflow ([ref](https://github.com/AlexanderSehr/bicep-registry-modules/actions/runs/20896155784))
- [x] (in main) workflow dispatch should run the workflow ([ref](https://github.com/AlexanderSehr/bicep-registry-modules/actions/runs/20896060812))
- [x] (in branch after merged to main) push to branch should not run the workflow
- [x] (in branch after merged to main) workflow dispatch should run the workflow ([ref](https://github.com/AlexanderSehr/bicep-registry-modules/actions/runs/20896060812))

Upstream
- [ ] (in main) push to main should run the workflow
- [ ] (in main) workflow dispatch should run the workflow

## Pipeline Reference

| Pipeline |
| -------- |
See above

## Type of Change

  - [ ] Update to documentation

### DIFF
--- a/.github/workflows/avm.res.analysis-services.server.yml
+++ b/.github/workflows/avm.res.analysis-services.server.yml
@@ -49,6 +49,7 @@ jobs:
   job_initialize_pipeline:
     runs-on: ubuntu-latest
     name: "Initialize pipeline"
+    # Only run if not cancelled, in upstream, or in forks on workflow_dispatch
     if: ${{ !cancelled() && !(github.repository != 'Azure/bicep-registry-modules' && github.event_name != 'workflow_dispatch') }}
     steps:
       - name: "Checkout"
--- a/.github/workflows/avm.res.analysis-services.server.yml
+++ b/.github/workflows/avm.res.analysis-services.server.yml
@@ -49,6 +49,12 @@ jobs:
   job_initialize_pipeline:
     runs-on: ubuntu-latest
     name: "Initialize pipeline"
+    if: |
+      !cancelled() &&
+      github.repository == 'Azure/bicep-registry-modules' ||
+      (github.repository != 'Azure/bicep-registry-modules' &&
+      github.event_name == 'push' &&
+      github.ref != 'refs/heads/main')
     steps:
       - name: "Checkout"
         uses: actions/checkout@v5
--- a/.github/workflows/avm.res.analysis-services.server.yml
+++ b/.github/workflows/avm.res.analysis-services.server.yml
@@ -42,14 +42,6 @@ env:
 concurrency:
   group: ${{ github.workflow }}
 
-# Test
-# x In fork (in main) push to main should not trigger the workflow
-# x        (in main) workflow dispatch should trigger the workflow
-#         (in branch after merged to main) push to branch should not trigger the workflow
-#         (in branch after merged to main) workflow dispatch should trigger the workflow
-# Upstream (in main) push to main should trigger the workflow
-#          (in main) workflow dispatch should trigger the workflow
-# Test change 1
 jobs:
   ###########################
   #   Initialize pipeline   #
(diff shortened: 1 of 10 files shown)


.github/workflows/avm.res.analysis-services.server.yml
AlexanderSehr, line 1: Disable actions before merging

.github/workflows/platform.check.psrule.yml
```diff
@@ -40,6 +40,7 @@ jobs:
   job_init_psrule_pipeline:
     runs-on: ubuntu-latest
     name: "Initialize pipeline"
+    if: github.repository == 'Azure/bicep-registry-modules'
```
eriqua: Is workflow dispatch working in forks? Asking in case we wanted to test platform pipelines in our forks

AlexanderSehr: They do, yes. Good point. Let me update it to only apply this condition if run on schedule (which is why I added this condition here). 👍

.github/workflows/avm.res.analysis-services.server.yml
```diff
@@ -49,6 +49,7 @@ jobs:
   job_initialize_pipeline:
     runs-on: ubuntu-latest
     name: "Initialize pipeline"
+    if: ${{ !cancelled() && !(github.repository != 'Azure/bicep-registry-modules' && github.event_name != 'workflow_dispatch') }}
```
eriqua, line 52: Minor, but what about a new line for each condition? If I'm not mistaken that's generally how other conditions are written

.github/workflows/platform.set-avm-github-pr-labels.yml
eriqua, line 1: Is the removal intended?

.github/workflows/avm.res.analysis-services.server.yml
AlexanderSehr: Had that before. We can have that - but - it will be rendered like
image

instead of

image

That's why I moved it into one line. Neither has an impact on the functionality, but just readability (with syntax highlighting -> incorrect coloring of almost every element). If you would still prefer multi-line here I can change it again.

.github/workflows/platform.set-avm-github-pr-labels.yml
AlexanderSehr: Yes. Didn't see the point of not running it for forks the same way we run e.g., the title check. No hard feelings though, if you prefer I can change it back.

.github/workflows/avm.res.analysis-services.server.yml
eriqua: Umh I guess I miss that extension 😄 
image

How do you render this instead?
```suggestion
    if: |
      !cancelled() && 
      !(github.repository != 'Azure/bicep-registry-modules' && 
      github.event_name != 'workflow_dispatch')
```

AlexanderSehr: It's the GitHub Actions extension :) 
If I recall correctly the `!=` is misinterpreted without the `${{ }}` as it is interpreted like some YAML-internal syntax instead of GitHub. But regardless, it would look like 
image

eriqua: Fair, I don't want to block on this one. Resolving

eriqua (approved):
🆗

PR #6535, opened 2026-02-03
feat: CI - Enable the injection of custom tags on test resources

### LINKED ISSUE #6505: [AVM CI Environment Issue]: Update diagnostic.dependencies.bicep to support Azure tags

### Issue Type?

Feature Request

### Description

I have my AVM CI/CD environment configured to use an internal Azure subscription; there are Azure policies that will prevent deployment of Storage Accounts with public access, and Event Hub Namespaces without Local Authentication disabled:

- Storage Account - Requires either publicNetworkAccess disabled or a skip tag
- Event Hub Namespace - Requires disableLocalAuth set to true or a skip tag

To work around this the policy managers have enabled the use of skip tags; Azure tags with specific values can be added to resources to allow them to be deployed normally. The problem is that  diagnostic.dependencies.bicep does not have the tags parameter exposed, so the necessary skip tags can't be added to allow my tests to complete without being impeded by policy. 

Propose the following:

- Add the 'tags' parameter to the resources defined in diagnostic.dependencies.bicep, and expose the parameter as an input
- Consider disabling public access for these resources altogether if not needed for any other patterns or resources
- API updates as appropriate

AlexanderSehr:
## Description

This PR shows how on could inject tags to their resources from the CI, that is, a key vault secret as per the already established environment-specific value injection logic.

Of this PR, the only change that will remain is the update of the shared dependency templates.

Closes #6505

Result in Azure Portal:

Diagnostic dependency LAW resource
image

Main module resource
image

Key Vault secret
image
image

## Pipeline Reference

| Pipeline |
| -------- |
[avm.res.key-vault.vault](https://github.com/Azure/bicep-registry-modules/actions/workflows/avm.res.key-vault.vault.yml)

## Type of Change

  - [ ] Update to documentation

### DIFF
--- a/.github/workflows/avm.res.key-vault.vault.yml
+++ b/.github/workflows/avm.res.key-vault.vault.yml
@@ -84,8 +84,7 @@ jobs:
     uses: ./.github/workflows/avm.template.module.yml
     with:
       workflowInput: "${{ needs.job_initialize_pipeline.outputs.workflowInput }}"
-      # moduleTestFilePaths: "${{ needs.job_initialize_pipeline.outputs.moduleTestFilePaths }}"
-      moduleTestFilePaths: '[{"path":"tests/e2e/waf-aligned/main.test.bicep","name":"waf-aligned","e2eIgnore":false}]'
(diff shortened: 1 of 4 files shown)


AlexanderSehr:
Closing as stale

PR #6584, opened 2026-02-21
fix(app-service-lza): correct web app kind parameter to respect OS configuration

Acestus:
## Description

The `kind` parameter value passed to the web app module is being replaced with a hardcoded `'app,linux'` value when any kind is specified. This breaks Windows App Service deployments because the staging slot gets created as Linux while the production site is Windows.

Found during deployment testing - the conditional on line 190 in `app-service.module.bicep`:

kind: !empty(kind) ? 'app,linux' : 'app'

This always evaluates to `'app,linux'` when a kind parameter is passed (e.g., `'app'` for Windows), regardless of what the actual value is. Changed to pass through the parameter value directly:

kind: kind

Error message encountered:

The slot 'kind' property 'app,linux' must match the Production site 'kind' property 'app'.

## Pipeline Reference

| Pipeline |
| -------- |

## Type of Change

  - [ ] Update to documentation

## Checklist


AlexanderSehr:
Hola @MikeTB-Microsoft & @ibersanoMS,
please triage this issue when you get the chance 👍

PR #6615, opened 2026-02-27
Fix/issue 6517 parameters system and user assigned

### LINKED ISSUE #6517: [AVM Module Issue]: Deploying Automation Account with some parameters failes

### Issue Type?

Bug

### Module Name

avm/res/automation/automation-account

### (Optional) Module Version

0.17.1

### Description

When using the parameter managedIdentities and with the parameter systemAssigned:false, the deployment fails. I suppose that the AA provider does not accept {type: None}, managedIdentities (or the var identity specifically) becomes {type: None} when SystemAssigned is set to false and userAssignedResourceIds are not configured. When the var identity is {} when systemAssigned:false, then the deployment is successfull.

```bicep
    managedIdentities: {
      systemAssigned: false
    }
  ```

### (Optional) Correlation Id

_No response_

### PR DESCRIPTION
**Pull request by gpacetti**
## Description

Fixes #6517

## Pipeline Reference

| Pipeline |
| -------- |
[avm.res.automation.automation-account](https://github.com/gpacetti/bicep-registry-modules/actions/workflows/avm.res.automation.automation-account.yml)

## Type of Change

  - [ x] Bugfix containing backwards-compatible bug fixes, and I have NOT bumped the MAJOR or MINOR version in `version.json`:
  - [ ] Update to documentation

## Checklist

- [ x] I'm sure there are no other open Pull Requests for the same update/change
- [x ] I have run `Set-AVMModule` locally to generate the supporting module files.
- [x ] My corresponding pipelines / checks run clean and green without any errors or warnings
- [ x] I have updated the module's CHANGELOG.md file with an entry for the next version

### DIFF
--- a/avm/res/compute/virtual-machine/CHANGELOG.md
+++ b/avm/res/compute/virtual-machine/CHANGELOG.md
@@ -9,7 +9,8 @@ The latest version of the changelog can be found [here](https://github.com/Azure
 - Skip pre-creating managedDataDisks when createOption is 'FromImage'
 - Preserve 'FromImage' createOption instead of overriding to 'Attach'
 - Don't set managedDisk.id for 'FromImage' disks (only storageAccountType)
-- Fixed typo: createoption → createOption
+- Fixed typo: `createoption` → `createOption`
+- Updated avm-common-types references to `0.7.0`
 
 ### Breaking Changes
 - The settings property support to the extensionMonitoringAgentConfig parameter for the Azure Monitor Agent extension.
--- a/avm/res/compute/virtual-machine/README.md
+++ b/avm/res/compute/virtual-machine/README.md
@@ -8111,13 +8111,11 @@ This section gives you an overview of all local-referenced module files (i.e., o
 
 | Reference | Type |
 | :-- | :-- |
-| `br/public:avm/res/network/network-interface:0.5.1` | Remote reference |
 | `br/public:avm/res/network/network-interface:0.5.3` | Remote reference |
 | `br/public:avm/res/network/public-ip-address:0.12.0` | Remote reference |
 | `br/public:avm/res/network/public-ip-address:0.8.0` | Remote reference |
-| `br/public:avm/utl/types/avm-common-types:0.5.1` | Remote reference |
-| `br/public:avm/utl/types/avm-common-types:0.6.0` | Remote reference |
 | `br/public:avm/utl/types/avm-common-types:0.6.1` | Remote reference |
+| `br/public:avm/utl/types/avm-common-types:0.7.0` | Remote reference |
 
 ## Notes
 
--- a/avm/res/compute/virtual-machine/main.bicep
+++ b/avm/res/compute/virtual-machine/main.bicep
@@ -25,7 +25,6 @@ param vTpmEnabled bool = false
 @description('Conditional. OS image reference. In case of marketplace images, it\'s the combination of the publisher, offer, sku, version attributes. In case of custom images it\'s the resource ID of the custom image. Required if not creating the VM from an existing os-disk via the `osDisk.managedDisk.resourceId` parameter.')
 param imageReference resourceInput<'Microsoft.Compute/virtualMachines@2025-04-01'>.properties.storageProfile.imageReference?
 
-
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

File: avm/res/automation/automation-account/main.bicep [hunk 1]
```diff
@@ -96,21 +96,25 @@ param sourceControlConfigurations sourceControlConfigurationType[]?
...
-      type: (managedIdentities.?systemAssigned ?? false)
-        ? (!empty(managedIdentities.?userAssignedResourceIds ?? {}) ? 'SystemAssigned, UserAssigned' : 'SystemAssigned')
-        : (!empty(managedIdentities.?userAssignedResourceIds ?? {}) ? 'UserAssigned' : 'None')
-      userAssignedIdentities: !empty(formattedUserAssignedIdentities) ? formattedUserAssignedIdentities : null
+      type: systemAssignedEnabled
+        ? (hasUserAssigned ? 'SystemAssigned, UserAssigned' : 'SystemAssigned')
+        : 'UserAssigned'
+      userAssignedIdentities: hasUserAssigned ? formattedUserAssignedIdentities : null
     }
   : null
```
AlexanderSehr: Hola @gpacetti,
thanks for the PR to address the fix. However, I'd need to ask you to please change the implementation back to the spec implementation that you can find [here](https://azure.github.io/Azure-Verified-Modules/specs/bcp/res/interfaces/#managed-identities) which is essentially what you had before. You are however welcome to change details about it.
If the resource provider does not accept `'None'` as a value for the `type`, it may be accepted to use `null` as the fallback value. Resource providers are unfortunately not exactly consistent when it comes to that. 

If `null` is not accepted either, I'd agree a change would be required, but also then I'd suggest to handle it slightly differently via
```bicep
var identity = !empty(managedIdentities) && (managedIdentities.?systemAssigned ?? false || !empty(formattedUserAssignedIdentities))
  ? {
      type: (managedIdentities.?systemAssigned ?? false)
        ? (!empty(formattedUserAssignedIdentities) ? 'SystemAssigned,UserAssigned' : 'SystemAssigned')
        : 'UserAssigned'
      userAssignedIdentities: !empty(formattedUserAssignedIdentities) ? formattedUserAssignedIdentities : null
    }
  : null
```

Any chance you could validate `null` and only if that doesn't work go with the adjusted implementation? Sorry for dwelling on this, we just try to keep the implementation as consistently as possible across modules to keep a high level of maintainability.

gpacetti: @AlexanderSehr  thanks for it. I'll work on it.

gpacetti: Ciao @AlexanderSehr I've tested the code provided but it is not compliant with BCP036. The AVM Bicep interface code provide null when UserAssigned is not specify and it conflict if SystenAssignedIdenity is false. 
```bicep
  var identity = !empty(managedIdentities) ? {
    type: (managedIdentities.?systemAssigned ?? false) ? (!empty(managedIdentities.?userAssignedResourceIds ?? {}) ? 'SystemAssigned,UserAssigned' : 'SystemAssigned') : (!empty(managedIdentities.?userAssignedResourceIds ?? {}) ? 'UserAssigned' : null)
    userAssignedIdentities: !empty(formattedUserAssignedIdentities) ? formattedUserAssignedIdentities : null
  } : null
  ```
  I can produce different code, but it will not compliant with the AVM guideline. 
  When managedIdentities is supplied, but systemAssigned is false and no userAssignedResourceIds are provided, the template still constructs an identity object even though no managed identity should be configured. In this case, Azure Automation accepts the deployment only when the identity property is effectively omitted. Therefore, the module must return null for identity when neither system-assigned nor user-assigned identities are requested.
 Input parameter:
```bicep
     managedIdentities: {
      systemAssigned: false
    }
```
 it is not compliant with our Bicep guideline, the suggestion here is to omit completely the managedIdentities parameter. 
 Based on this, I suggest to close this PR and the Issue #6517 too with no fix.

AlexanderSehr: Hola @gpacetti,
that's unfortunate to hear. Thanks for trying 🙏.
If you don't mind, please keep the issue open and I'll try to take a stab at it as my capacity allows. I'll re-assign it to myself.

PR #6751, opened 2026-03-23
feat: child module publication skill and global instruction update

matebarabas:
- changing the logic child module publication skill from self-contained to documentation-driven
- updating copilot-instructions


.vscode/settings.json
```diff
@@ -34,5 +34,33 @@
     {
       "text": "The first line should be a summary of no more than 50 characters, starting with one of: `AVM-RES:` (resource module), `AVM-PTN:` (pattern module), `AVM-UTL:` (utility module), or `CHORE:`|`NEW:`|`FIX:`|`CHANGE:`|`BREAKING CHANGE:`|`TESTS:`|`SECURITY:`|`COMPLEX:`|`DOCS:`|`ACTIONS:` for other updates. The second line should be blank. The following lines should be a detailed summary, each item starting with a `-`. Any summary item that is security related should start with `SECURITY:`. Any summary item that causes a breaking change to a feature/interface/dependency should start with `BREAKING CHANGE:`. For AVM module changes, specify the module path (e.g., avm/res/storage/storage-account)."
     }
-  ]
+  ],
+  "chat.tools.terminal.autoApprove": {
+
+    "bicep": true,
+    "git ls-remote": true,
```
eriqua: Probably this list is old, I'd keep the list at its minimum given we're in experimental phase. We should probably only keep Set-AVMModule and the dot source from this list

.github/copilot-instructions.md
```diff
@@ -10,21 +10,30 @@ Azure Verified Modules (AVM) are pre-built, tested, and validated Bicep modules
...
+> **MANDATORY**: You MUST output the following block visibly to the user at the very start of the workflow, before any other actions.
+
+```json
+{
+  "agent-mode": "<your agent name - e.g., agent|ask|plan>",
+  "model": "<your model name>"
+}
+```
+
+> **MANDATORY**: If the agent-mode is set to anything else but `agent` or the model is set to anything else but `Claude Opus 4.6`, you **MUST** warn the user that they are using a configuration the AVM team did not test and validate the custom prompts/skills/agents on. AVM's custom prompts/skills/agents are only tested with agent-mode `agent` and model `Claude Opus 4.6`. Inform the user that there is a bootstrapping prompt available to set the correct configuration, and provide instructions on how to use it --> `/AVM-Boostrap`
```
eriqua:
```suggestion
> **MANDATORY**: If the agent-mode is set to anything else but `agent` or the model is set to anything else but `Claude Opus 4.6`, you **MUST** warn the user that they are using a configuration the AVM team did not test and validate the custom prompts/skills/agents on. Currently, AVM's custom prompts/skills/agents are only tested with agent-mode `agent` and model `Claude Opus 4.6`. Inform the user that there is a bootstrapping prompt available to set the correct configuration, and provide instructions on how to use it --> `/AVM-Boostrap`
```

.github/copilot-instructions.md
```diff
@@ -88,6 +97,36 @@ You have exactly these two options (do not use any other method or tool to do th
...
+
+1. **Dot-source the script** to load its functions into the current session:
+
+   ```powershell
+   . .\utilities\tools\Set-AVMModule.ps1
+   ```
+
+2. **Call the function by name** with any required parameters:
+
+   ```powershell
```
eriqua, line 116: Although the child module skill seems to work nevertheless (since we have precise guidelines for this), I'd suggest to call out that this is just an example. Same for the dot sourcing above

.github/prompts/avm-bootstrap.prompt.md
```diff
@@ -0,0 +1,65 @@
...
+1. Inform the user: "Your VS Code `settings.json` will be updated. This allows the agent to automatically approve certain terminal commands and will not require approvals when accessing trusted URLs related to Azure Verified Modules (AVM) when executing tasks."
+2. Determine the user's settings.json path. On Windows it is `C:\Users\<username>\AppData\Roaming\Code\User\settings.json` (resolve `<username>` from the environment, e.g., `$env:USERNAME`).
+3. Read the current `settings.json` file.
+4. For each key below, check if it already exists with the correct value. If it is missing or differs, merge it into the file (preserve all existing settings).
+5. Save the updated file.
+
+**Required configuration entries**:
+
+```json
+"chat.tools.terminal.autoApprove": {
```
eriqua: Updating the user's local settings.json sounds very wrong. Also this is duplicated from the repo settings and hence difficult to maintain. It will diverge in no time. 
I suggest to hold on this prompt for the time being. This is anyway not strictly in the scope of the skill update

.github/skills/avm-child-module-publishing/SKILL.md
```diff
@@ -1,24 +1,13 @@
...
-
-## Triggers
-
-Activate this skill when user wants to:
-
-- Publish an existing Bicep child module to the AVM public registry
-- Enable a child module for independent publishing
-- Add telemetry to a child module
-- Register a child module in the allowed list
+> This skill implements the official AVM Bicep Child Module Publishing workflow from Azure Verified Modules. Follow these steps exactly. Do not skip or reorder steps.
```
eriqua, line 10: If it was for me I'd remove this as well.

.github/prompts/avm-bootstrap.prompt.md
matebarabas: The divergence is a good callout - however, I'd still prefer keeping this feature, as the repo level setting stops working the moment you have more than one repo hooked up in your workspace - which many people do. Here's a suggestion: I'll update the logic so that it draws the configuration from the repo level settings.json. I'll also review the content of which one more time. We will likely have to add more to it too, to make the experience smooth(er). I think the driving logic should be that those commands that list/get/search/export/convertto- and from/create new/etc. should be listed by default. Those that can potentially delete/remove/destroy/etc. should not.

On the overall hold-up, I disagree. This is a bootstrapping file that sets up the user's environment. It's only applied when the user explicitly runs it.

.vscode/settings.json
matebarabas: I'll remove those commands that can also potentially cause harm (without allowed params defined). For a relatively smooth experience, we'll need more than just that one script though. Please see my justification [below](https://github.com/Azure/bicep-registry-modules/pull/6751#discussion_r3034567877).

.github/skills/avm-child-module-publishing/SKILL.md
matebarabas: We didn't test the workflow without this, but we certainly could. Not sure if skipping the "Follow these steps exactly. Do not skip or reorder steps." part makes any difference in certain edge cases.

.github/copilot-instructions.md
matebarabas: Good call, let's try to avoid confusing Copilot on how to run this script in other scenarios. :)

I added this above: 

> Note: This is just an example of how to run any PowerShell script in this repository, and does not mean that this script always needs exactly these parameters.

.github/prompts/avm-bootstrap.prompt.md
eriqua: Sorry, I'm totally against this so I cannot approve it. As it's a separate topic than the skill update, I suggest to at least discuss it separately from this PR.

.vscode/settings.json
```diff
@@ -34,5 +34,29 @@
...
+        "approveRequest": false,
+        "approveResponse": true
+    },
+    "https://raw.githubusercontent.com/Azure/Azure-Verified-Modules/refs/heads/main": {
+        "approveRequest": false,
+        "approveResponse": true
+    },
+    "https://learn.microsoft.com": true,
+    "https://*.microsoft.com": true
+  }
```
jtracey93: Should we document this somewhere on the AVM site just to cover ourselves in terms of we auto-approve these and the reasons why, etc.?

.github/prompts/avm-bootstrap.prompt.md
matebarabas: As agreed, this has now been moved to a [separate PR](https://github.com/Azure/bicep-registry-modules/pull/6813)

.vscode/settings.json
matebarabas: Do we have any of the other configuration that is in this file documented somewhere? Not against capturing a blurb, just asking if this already exists. If it doesn't yet, I'd cover it in a later PR. The short answer is that these are commonly used scripts - either our own ones (that should be trusted withing the context of the repo) or system cmdlets that don't make any changes, therefore can't make harm.

.vscode/settings.json
```diff
@@ -34,5 +34,29 @@
...
-  ]
+  ],
+  "chat.tools.terminal.autoApprove": {
+    "ConvertFrom-Json": true,
+    "ConvertTo-Json": true,
+    "Set-AVMModule": true,
+    ". .\\utilities\\": true,
+    "Invoke-Pester": true,
+    "New-PesterContainer": true,
+    "Resolve-Path": true
```
eriqua: Please keep the repo list restricted to what specifically used by the skill. I believe everything else comes from unrelated tasks (e.g. using copilot to retrieve the list of child modules to publish) which shouldn't land here

```suggestion
    "Set-AVMModule": true,
    ". .\\utilities\\": true
```

.github/skills/avm-child-module-publishing/SKILL.md
eriqua: As agreed for this version it is fair to keep and if useful we'll further rework this later

.vscode/settings.json
eriqua: Moved comment to new PR

.vscode/settings.json
eriqua: Resolving as the file has been moved to a new PR #6813

eriqua (approved):
🆗

PR #6753, opened 2026-03-24
feat: module maintenance skill

matebarabas:
odule maintenance skill


.github/skills/avm-module-maintenance/SKILL.md
```diff
@@ -0,0 +1,225 @@
+---
```
eriqua, line 1: My main concern with this is that this would potentially result in a huge PR, which I personally don't support.

I believe it would be wiser to cover smaller scopes, for the time being, but optimizing it. Let's consider the API version check as an example. Updating the api version is not always as easy as changing the version. It involves new features, additional parameters, potentially, although not too often, even breaking changes. Not to mention new child resources. Depending on this, we should mention what version update we expect, major (not now), minor or patch. This is where we should leverage AI support imho. It's very much a skill on its own.

Once we have a few of those well scoped skills, we can think about implementing a "super skill".

I'd probably deprioritize content already fully covered by static validation. For example I'd avoid checking if we have waf-aligned and defaults tests. If we didn't, the module wouldn't be there. Note: the max test is not mandatory, so it's not correct to include it. Plus, "Deploys with all (or most) parameters set" is not providing clear guidance. Instead, what about: for a max test I would like to verify if all possible parameters are included and out of all possible parameters have a clear understanding of which parameters cannot be used together, as they would otherwise result in a conflict. Just food for thoughts

ericscheffler: See also #6817 for my API update skill; the main thing I wanted to add to the existing prompt was the recommendation & handling of version.json and changelog updates (since I kept forgetting it). I also made sure it notifies the user & handles new parameters missing from the current module. Open to suggestions on it though!

PR #6792, opened 2026-04-02
feat: Utility script to retrieve specific modules (e.g., orphaned, versioned, children)

eriqua:
## Description

Utility to list modules with specific characteristic (e.g., orphaned, versioned, children)

## Pipeline Reference

| Pipeline |
| -------- |

## Type of Change

  - [ ] Update to documentation

## Checklist


utilities/pipelines/sharedScripts/Get-ModuleList.ps1
```diff
@@ -0,0 +1,132 @@
+﻿<#
```
eriqua, line 1: disable actions before merge

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
```diff
@@ -0,0 +1,132 @@
...
+
+.PARAMETER Orphaned
+Optional. If set, filter by orphaned status:
+- -Orphaned         : Only return modules that contain an ORPHANED.md file.
+- -Orphaned:$false  : Only return modules that do NOT contain an ORPHANED.md file.
+
+.PARAMETER Versioned
+Optional. If set, filter by versioning status:
+- -Versioned         : Only return modules that contain a version.json file.
+- -Versioned:$false  : Only return modules that do NOT contain a version.json file.
```
AlexanderSehr: `-versioned:$false`? 😄 Something tells me AI defined this. Please remove both examples. A switch statement should imply the true and false statement by providing it. You only need the `:$false` if splatting it in an object.

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
```diff
@@ -0,0 +1,132 @@
...
+
+        [Parameter(Mandatory = $false)]
+        [ValidateSet('All', 'TopLevel', 'Child')]
+        [string] $Scope = 'All',
+
+        [Parameter(Mandatory = $false)]
+        [switch] $Orphaned,
+
+        [Parameter(Mandatory = $false)]
+        [switch] $Versioned,
```
AlexanderSehr:
```suggestion
        [Parameter(Mandatory = $false)]
        [switch] $IsOrphaned,

        [Parameter(Mandatory = $false)]
        [switch] $IsVersioned,
```
What do you think? Maybe be a tat more descriptive

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
```diff
@@ -0,0 +1,132 @@
...
+        if ($Orphaned) {
+            $modules = $modules | Where-Object {
+                Test-Path (Join-Path $RepoRoot $_ 'ORPHANED.md')
+            }
+        } else {
+            $modules = $modules | Where-Object {
+                -not (Test-Path (Join-Path $RepoRoot $_ 'ORPHANED.md'))
+            }
+        }
+    }
```
AlexanderSehr:
```suggestion
    if ($Orphaned) {
        $modules = $modules | Where-Object {
            Test-Path (Join-Path $RepoRoot $_ 'ORPHANED.md')
        }
    } else {
        $modules = $modules | Where-Object {
            -not (Test-Path (Join-Path $RepoRoot $_ 'ORPHANED.md'))
        }
    }
```
Why the extra check? Switch statements are `$false` if not provided

AlexanderSehr: Same comment for `-Versioned`

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
eriqua: nope, this is intentional. the difference is without the switch you'll get all. -versioned:false will instead provide the not versioned ones

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
```diff
@@ -0,0 +1,132 @@
+﻿<#
+.SYNOPSIS
+Retrieve a filtered list of AVM module folders.
+
+.DESCRIPTION
+Scans a given root path for Bicep modules (folders containing a main.bicep file) and returns them filtered by the requested characteristics.
+Filters can be combined: -Scope controls hierarchy (All/TopLevel/Child), -Orphaned limits to orphaned modules, -Versioned limits to versioned or not-versioned modules.
+
+.PARAMETER Path
```
AlexanderSehr, line 12: I'd recommend using the absolute path, not relative. Relative is a bit of a pain to use and less intuitive. If changed, the repo root parameter could still be used - or - split (like we have in out current utilities) like
`$repoRoot = ($Path -split '[\/|\\]avm[\/|\\](res|ptn|utl)[\/|\\]')[0]`

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
```diff
@@ -0,0 +1,132 @@
...
+
+        [Parameter(Mandatory = $false)]
+        [switch] $Versioned,
+
+        [Parameter(Mandatory = $false)]
+        [string] $RepoRoot = (Get-Item -Path $PSScriptRoot).parent.parent.parent.FullName
+    )
+
+    $absolutePath = Resolve-Path -Path (Join-Path $RepoRoot $Path) -ErrorAction Stop
+    $repoRootNormalized = $RepoRoot -replace '[\\/]$', ''
```
AlexanderSehr:
```suggestion
    $repoRootNormalized = $RepoRoot.TrimEnd('/\')
```
May be a bit simpler :)

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
```diff
@@ -0,0 +1,132 @@
...
+
+    # Discover all module folders (i.e., containing a main.bicep) and convert to module names (e.g., avm/res/storage/storage-account)
+    $modules = Get-ChildItem -Path $absolutePath -Filter 'main.bicep' -Recurse -File | ForEach-Object {
+        $_.Directory.FullName -replace [regex]::Escape($repoRootNormalized), '' -replace '^[\\/]', '' -replace '\\', '/'
+    }
```
AlexanderSehr: Could be simplified to
```pwsh
$repoRoot = ($Path -split '[\/|\\]avm[\/|\\](res|ptn|utl)[\/|\\]')[0]
$modules = Get-ChildItem -Path $Path -Filter 'main.bicep' -Recurse -File | ForEach-Object {
    $_.Directory.FullName -replace ('{0}[\\|\/]?' -f [regex]::Escape($RepoRoot))
}
```

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
```diff
@@ -0,0 +1,132 @@
...
+
+    # Apply scope filter
+    switch ($Scope) {
+        'TopLevel' {
+            $modules = $modules | Where-Object { ($_ -split '/').Count -eq 4 }
+        }
+        'Child' {
+            $modules = $modules | Where-Object { ($_ -split '/').Count -gt 4 }
+        }
+    }
```
AlexanderSehr, line 95: I'd recommend to speed this function up by not interating through the list of `$modules` for each condition, but just loop through them once and check the conditions while doing so. E.g.
```pwsh
$modules | Where-Object {
    $Orphaned ? (Test-Path (Join-Path $repoRoot $_ 'ORPHANED.md')) : -not (Test-Path (Join-Path $repoRoot $_ 'ORPHANED.md')) -and
    $Versioned ? (Test-Path (Join-Path $repoRoot $_ 'version.json')) : -not (Test-Path (Join-Path $repoRoot $_ 'version.json')) -and
    $Scope -eq 'TopLevel' ? (($_ -split '/').Count -eq 4) : $Scope -eq 'Child' ? (($_ -split '/').Count -gt 4) : $true
}
```

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
eriqua: like

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
eriqua: the reason is what replied above. getting 3 out of a switch: all, versioned/orphaned and not versioned/not orphaned

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
eriqua: The splitting on avm caused some issues in the past for those including avm in the path where the repo is locally cloned. I understand that's still the case with the test-modulelocally, although not sure if related to the above split as I didn't get to troubleshoot it.
In general I support consistency and it's true we are using the absolute path across our utilities so let me update that. For now I'd leverage the reporoot for the split, we can still change it later

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
eriqua: Also in this case I believe we'd loose the chance to specifically retrieve not orphaned and not versioned. I'll try to simplify in the suggested sense if that is not the case

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
AlexanderSehr: But that's the whole point of using a switch parameter - to be $false by default and interpret it as such😄 In the current case I'd recommend switching the type to `[bool]` which would then, by default, be `null` 👍 

Using a switch is like the `-Recurse` / `-SkipReadme` / etc. flags for the `Set-AVMModule` script. If you specify them that means you want to set the setting / behavior for the function. If you don't, you don't specify them with `$false` and the function should then not do what the parameter says.

The way the switch statement is used here feels like a misuse of the intend behind the parameter type which explains why you had to refer to a somewhat hacky-feeling `$PSBoundParameters.ContainsKey(...)`. With boolean you can instead just check for `($null -eq $isVersioned)`  making the function's logic also a bit easier to follow.

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
AlexanderSehr: That's why we used the split regex `'[\/|\\]avm[\/|\\](res|ptn|utl)[\/|\\]'`. You're right that we used to have the issue because we split only on e.g. `/avm/` and folks named the repo's folder for example `avm`, causing issues. Since then we changed it to the aforementioned regex, going under the assumption that there will never be a path like `/avm/res/` twice in the same absolute path (e..g, `C:/dev/avm/res/avm/res/compute/virtual-machine`). Personally I much prefer not to have a parameter that can safely be deferred, but I won't block you over it either. It's just my recommendation that you're fee to dismiss and resolve this comment 😉

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
AlexanderSehr: If using a boolean parameter as suggested in my follow-up comment, the above would, I guess, change to
```pwsh
$modules | Where-Object {
    $null -ne $IsOrphaned ? ($IsOrphaned ? (Test-Path (Join-Path $repoRoot $_ 'ORPHANED.md')) : -not (Test-Path (Join-Path $repoRoot $_ 'ORPHANED.md'))) : $true -and
    $null -ne $IsVersioned ? ($IsVersioned ? (Test-Path (Join-Path $repoRoot $_ 'version.json')) : -not (Test-Path (Join-Path $repoRoot $_ 'version.json'))) : $true -and
    $Scope -eq 'TopLevel' ? (($_ -split '/').Count -eq 4) : $Scope -eq 'Child' ? (($_ -split '/').Count -gt 4) : $true
}
```

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
eriqua: I somehow liked the hybrid approach. With bool one would need to specify $true as well. However, if you feel the current implementation leans towards a hack/antipattern, I'm happy to change that. I'd avoid the native switch and use a bool. 

What's your take on the -Scope at that point? Would it be more consistent to have a IsChild bool, null by default, following the same implementation?

AlexanderSehr: Yes, I'd strongly lean towards a bool here if we want to have a tripple meaning of a 2-value parameter 😄

Regarding the scope I'd not change anything as the scope 'could' take any number of different values and is not as black and white as 'isVersioned'. That being said, if you just want to go after 'IsChild' or not (and not support other scopes in the future) its a different story. Your call.

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
eriqua: Updated input to nullable bool

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
eriqua: Updated input to nullable bool

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
eriqua: I aligned the implementation when retrieving the modules from the resolved module paths. I cannot use the same for the repo root since path is not necessarily a module path nor a moduleClass path. The above would fail with Path='path/to/avm/' for example. For the time being I prefer to retrieve the RepoRoot relatively to the script by default, but allowing a different value via input. This is also common to other existing utilities

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
eriqua: Check updated implementation

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
eriqua: Check updated implementation

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
eriqua: done

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
AlexanderSehr: I have my preference, but won't pressure you into it. Will resolve :)

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
```diff
@@ -0,0 +1,105 @@
...
+
+Get all orphaned modules (those with an ORPHANED.md file).
+
+.EXAMPLE
+Get-ModuleList -Path 'C:/repo/avm/res/storage/storage-account' -Scope 'Child'
+
+Get all child (nested) modules under the storage-account module.
+
+.EXAMPLE
+Get-ModuleList -Path 'C:/repo/avm/res' -Scope 'Child' -IsVersioned:$false
```
AlexanderSehr:
```suggestion
Get-ModuleList -Path 'C:/repo/avm/res' -Scope 'Child' -IsVersioned $false
```

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
```diff
@@ -0,0 +1,105 @@
...
+
+Get all child resource modules that do NOT have a version.json file, i.e., modules not published to the public bicep registry.
+
+.EXAMPLE
+Get-ModuleList -Path 'C:/repo/avm/res' -Scope 'TopLevel' -IsOrphaned:$true -IsVersioned:$true
```
AlexanderSehr:
```suggestion
Get-ModuleList -Path 'C:/repo/avm/res' -Scope 'TopLevel' -IsOrphaned:$true -IsVersioned $true
```

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
```diff
@@ -0,0 +1,105 @@
...
+
+.PARAMETER IsVersioned
+Optional. If set, filter by versioning status:
+- -IsVersioned:$true   : Only return modules that contain a version.json file.
+- -IsVersioned:$false  : Only return modules that do NOT contain a version.json file.
+
+.PARAMETER HasChildren
+Optional. If set, filter by whether a module has nested child modules (subfolders containing a main.bicep file):
+- -HasChildren:$true   : Only return modules that contain at least one nested module.
+- -HasChildren:$false  : Only return modules that do NOT contain any nested modules.
```
AlexanderSehr:
```suggestion
- `-HasChildren $true`  : Only return modules that contain at least one nested module.
- `-HasChildren $false` : Only return modules that do NOT contain any nested modules.
```

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
```diff
@@ -0,0 +1,105 @@
...
+
+.PARAMETER IsOrphaned
+Optional. If set, filter by orphaned status:
+- -IsOrphaned:$true   : Only return modules that contain an ORPHANED.md file.
+- -IsOrphaned:$false  : Only return modules that do NOT contain an ORPHANED.md file.
+
+.PARAMETER IsVersioned
+Optional. If set, filter by versioning status:
+- -IsVersioned:$true   : Only return modules that contain a version.json file.
+- -IsVersioned:$false  : Only return modules that do NOT contain a version.json file.
```
AlexanderSehr:
```suggestion
- `-IsVersioned $true`   : Only return modules that contain a version.json file.
- `-IsVersioned $false`  : Only return modules that do NOT contain a version.json file.
```

utilities/pipelines/sharedScripts/Get-ModuleList.ps1
```diff
@@ -0,0 +1,105 @@
...
+.PARAMETER Scope
+Optional. Controls the hierarchy filter:
+- 'All'      : (Default) Return every folder that contains a main.bicep.
+- 'TopLevel' : Return only the first-level modules (depth 4: avm/<type>/<provider>/<resource>).
+- 'Child'    : Return only child modules nested under the given -Path (excludes the root module itself).
+
+.PARAMETER IsOrphaned
+Optional. If set, filter by orphaned status:
+- -IsOrphaned:$true   : Only return modules that contain an ORPHANED.md file.
+- -IsOrphaned:$false  : Only return modules that do NOT contain an ORPHANED.md file.
```
AlexanderSehr:
```suggestion
- `-IsOrphaned $true`   : Only return modules that contain an ORPHANED.md file.
- `-IsOrphaned $false`  : Only return modules that do NOT contain an ORPHANED.md file.
```

AlexanderSehr (approved):

PR #6804, opened 2026-04-06
fix(alz/ama): use unique lock name for PerfOnly VM Insights DCR

### LINKED ISSUE #6794: [AVM Module Issue]: Duplicate lock resource names cause InvalidTemplate error in Deployment Stacks

### Issue Type?

Bug

### Module Name

avm/ptn/alz/ama

### (Optional) Module Version

0.1.1

### Summary
 When deploying `avm/ptn/alz/ama` version `0.1.1` via Azure Deployment Stacks, the deployment fails with:

```
InvalidTemplate: Deployment template validation failed: 'The resource
'Microsoft.Insights/dataCollectionRules/dcr-vmi-alz-westeurope/providers/Microsoft.Authorization/locks/dcr-vmi-alz-westeurope-lock'
is not defined in the template.
```
 
### Root Cause
 
The module defines two conditionally-exclusive lock resources for the VM Insights DCR:
 
1. `dataCollectionRuleVMInsightsPerfandMapLock` — condition: `PerfAndMap` experience
2. `dataCollectionRuleVMInsightsPerfOnlyLock` — condition: `PerfOnly` experience
 
Both resources generate the **same lock name** using the same expression:

```json
"name": "[coalesce(tryGet(parameters('lockConfig'), 'name'), format('{0}-lock', parameters('dataCollectionRuleVMInsightsName')))]"
```

Although only one lock is deployed at runtime (they are mutually exclusive via condition), Azure Deployment Stacks performs template validation before evaluating conditions. This causes a conflict because two resources reference the same scope+name combination.

The error occurs even when lockConfig is null or kind is set to 'None'.

**Steps to Reproduce**

- Deploy avm/ptn/alz/ama:0.1.1 via New-AzSubscriptionDeploymentStack
- Pass lockConfig: null or any lockConfig value
- Deployment fails immediately at template validation

**Expected Behavior**
Deployment should succeed. The two lock resources are mutually exclusive by condition so there should be no conflict at runtime.

**Actual Behavior**
Template validation fails with InvalidTemplate because Deployment Stacks resolves resource IDs for all conditional resources regardless of their condition.

**Suggested Fix**
Give dataCollectionRuleVMInsightsPerfOnlyLock a unique name to avoid the naming collision:
 
```
"name": "[coalesce(format('{0}-perfonly', tryGet(parameters('lockConfig'), 'name')), format('{0}-perfonly-lock', parameters('dataCollectionRuleVMInsightsName')))]"
```

This is the same pattern used successfully within the module for other resources.

### (Optional) Correlation Id

0e84d9ef-e514-47b2-9d02-a920690f058b

oZakari:
## Description

The module defines two conditionally-exclusive lock resources for the VM Insights DCR (`PerfAndMap` and `PerfOnly`). Both generated the same lock name because both DCR resources use the same `dataCollectionRuleVMInsightsName` parameter. While only one deploys at runtime, Azure Deployment Stacks validates all resource IDs before evaluating conditions, causing an `InvalidTemplate` error.

The fix gives both locks distinct names (`-perfandmap-lock` and `-perfonly-lock` suffixes) so the two lock resource IDs never collide during template validation.

Also updated `avm/res/managed-identity/user-assigned-identity` from `0.4.3` to `0.5.0` and `Microsoft.Resources/deployments` API from `2024-03-01` to `2024-07-01`.

Fixes #6794

## Pipeline Reference

| Pipeline |
| -------- |
[avm.ptn.alz.ama](https://github.com/oZakari/bicep-registry-modules/actions/workflows/avm.ptn.alz.ama.yml)

## Type of Change

  - [ ] Update to documentation

## Checklist

### DIFF
--- a/avm/ptn/alz/ama/main.bicep
+++ b/avm/ptn/alz/ama/main.bicep
@@ -37,7 +37,7 @@ param enableTelemetry bool = true
 // Resources      //
 // ============== //
 
-module userAssignedManagedIdentity 'br/public:avm/res/managed-identity/user-assigned-identity:0.4.3' = {
+module userAssignedManagedIdentity 'br/public:avm/res/managed-identity/user-assigned-identity:0.5.0' = {
   name: '${uniqueString(subscription().id, resourceGroup().id, location)}-UserAssignedIdentity'
   params: {
     name: userAssignedIdentityName
@@ -151,7 +151,7 @@ resource dataCollectionRuleVMInsightsPerfOnly 'Microsoft.Insights/dataCollection
 
 resource dataCollectionRuleVMInsightsPerfandMapLock 'Microsoft.Authorization/locks@2020-05-01' = if (lockConfig.?kind != 'None' && !(empty(lockConfig.?name)) && dataCollectionRuleVMInsightsExperience == 'PerfAndMap') {
   scope: dataCollectionRuleVMInsightsPerfAndMap
-  name: lockConfig.?name ?? '${dataCollectionRuleVMInsightsPerfAndMap.name}-lock'
+  name: '${lockConfig.?name ?? dataCollectionRuleVMInsightsPerfAndMap.name}-perfandmap-lock'
   properties: {
     level: lockConfig.?kind ?? 'ReadOnly'
   }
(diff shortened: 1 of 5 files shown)


avm/ptn/alz/ama/main.bicep
```diff
@@ -151,15 +151,15 @@ resource dataCollectionRuleVMInsightsPerfOnly 'Microsoft.Insights/dataCollection
 
 resource dataCollectionRuleVMInsightsPerfandMapLock 'Microsoft.Authorization/locks@2020-05-01' = if (lockConfig.?kind != 'None' && !(empty(lockConfig.?name)) && dataCollectionRuleVMInsightsExperience == 'PerfAndMap') {
   scope: dataCollectionRuleVMInsightsPerfAndMap
-  name: lockConfig.?name ?? '${dataCollectionRuleVMInsightsPerfAndMap.name}-lock'
+  name: '${lockConfig.?name ?? dataCollectionRuleVMInsightsPerfAndMap.name}-perfandmap-lock'
```
AlexanderSehr, line 154: Jojojo @oZakari,
I hope you're doing well. Just wondering, regarding this change: As this changes the (fallback) name, I guess this could be considered a breaking change, right?
If you agree, I'd recommend upgrading the version in the `version.json`.

oZakari: Agreed, made the adjustments.

AlexanderSehr (approved):

PR #6813, opened 2026-04-09
feat: add bootstrap skill

matebarabas:
Adding a bootstrapping skill that sets the model, agent, tools and list of allowed tools.

### DIFF
--- a/.github/prompts/avm-bootstrap.prompt.md
+++ b/.github/prompts/avm-bootstrap.prompt.md
@@ -1,0 +1,38 @@
+---
+name: "AVM-Bootstrap"
+description: "Bootstrap Azure Verified Modules work in this repository"
+agent: agent
+model: "Claude Opus 4.6 (copilot)"
+tools:
+  [
+    vscode,
+    execute,
+    read,
+    agent,
+    edit,
+    search,
+    web,
+    "bicep/*",
+    "github/*",
+    browser,
+    todo,
+  ]
+---
+
+## Configuration
+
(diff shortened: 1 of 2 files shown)


.vscode/settings.json
```diff
@@ -34,5 +34,29 @@
...
-  ]
+  ],
+  "chat.tools.terminal.autoApprove": {
+    "ConvertFrom-Json": true,
+    "ConvertTo-Json": true,
+    "Set-AVMModule": true,
+    ". .\\utilities\\": true,
+    "Invoke-Pester": true,
+    "New-PesterContainer": true,
+    "Resolve-Path": true
```
eriqua, line 45: Porting comments from related PR https://github.com/Azure/bicep-registry-modules/pull/6751#discussion_r3062995130

Please keep the repo list restricted to what specifically used by the skill. I believe everything else comes from unrelated tasks (e.g. using copilot to retrieve the list of child modules to publish) which shouldn't land here
```suggestion
    "Set-AVMModule": true,
    ". .\\utilities\\": true,
```
We should probably agree on how to maintain this list. For example make sure only functions (or a subset of functions) internal to the repo are allowed for auto-approve. I wouldn't recommend broadening the scope to include any harmless Azure cmdlet, as this would likely result in an ever‑growing list that is difficult to maintain and scale over time.

.vscode/settings.json
```diff
@@ -34,5 +34,29 @@
...
+  },
+    "chat.tools.urls.autoApprove": {
```
eriqua, line 47: Porting comment from related PR https://github.com/Azure/bicep-registry-modules/pull/6751#discussion_r3060038968

We should agree on where to document reasoning behind auto-approvals

.github/prompts/avm-bootstrap.prompt.md
eriqua, line 1: Porting comment from related PR: let's further discuss the fundamental approach of updating the settings.json file local to the developer machine

PR #7071, opened 2026-05-18
feat: avm/res/api-center/service initial implementation

maciejporebski:
## Description

Initial implementation of `avm/res/api-center/service` module proposed in https://github.com/Azure/Azure-Verified-Modules/issues/2500.

## Design Decisions
- The template abstracts away the "workspace" sub-resource, as only a single 'default' one is currently supported with no published plans to support additional ones at the moment. The inclusion of additional non-default workspaces can be added in the future by introducing an additional parameter and continuing to use the current parameters for deployment of the default workspace.
- API Sources (`Microsoft.ApiCenter/services/workspaces/apiSources`) has been included using a preview API version. No stable version currently exists for the type, but the type is key for most API Center deployments. While the type requires a preview API version the underlying feature itself is not in preview.

## Known Issues
- Purging API Center workspaces does not currently work. To prevent soft-deleted workspaces from failing tests the resource names are suffixed with a string derived from the deployment name. This can be replaced by a purge operation once the RP fixes/implements the purge operation. (https://github.com/Azure/azure-rest-api-specs/issues/43339)

## Pipeline Reference

| Pipeline |
| -------- |
[avm.res.api-center.service](https://github.com/maciejporebski/bicep-registry-modules/actions/workflows/avm.res.api-center.service.yml)

## Type of Change

  - [ ] Update to documentation

## Checklist


.github/workflows/avm.res.api-center.service.yml
```diff
@@ -0,0 +1,90 @@
...
+            - main
+        paths:
+            - ".github/actions/templates/avm-**"
+            - ".github/workflows/avm.template.module.yml"
+            - ".github/workflows/avm.res.api-center.service.yml"
+            - "avm/res/api-center/service/**"
+            - "utilities/pipelines/**"
+            - "!utilities/pipelines/platform/**"
+            - "!*/**/child-module-publish-allowed-list.json"
+            - "!*/**/README.md"
```
jtracey93, line 31: We have recently changed the paths triggers for AVM modules, can you update them to match another module as an example here: https://github.com/Azure/bicep-registry-modules/blob/main/.github/workflows/avm.res.cache.redis.yml#L28-L31

just needs to be like this:
```yml
paths:
      - ".github/workflows/avm.res.api-center.service.yml"
      - "avm/res/api-center/service/**"
      - "!*/**/README.md"
```

jtracey93 (changes requested):

jtracey93 (approved):

PR #7099, opened 2026-05-23
fix: `avm/res/sql/server` handle versionless CMK autorotation URIs

### LINKED ISSUE #7073: [AVM Module Issue]: `sql/server` autoRotationEnabled software backed CMK broken after #6193

### Issue Type?

Bug

### Module Name

avm/res/sql/server

### (Optional) Module Version

0.21.0 ~ 0.21.2

### Description

last known good working version 0.17.0 & 0.20.3
anything higher and initial create & updates fails
```
module sql 'br/public:avm/res/sql/server:0.20.3' = {
  scope: resourceGroup(rg.name)
  name: sqlName
  params: {
    name: sqlName
    customerManagedKey: {
      keyName: sqlCmkKeyName
      keyVaultResourceId: sqlCmkKeyVaultResourceId
      autoRotationEnabled: true
      userAssignedIdentityResourceId: sqlIdentity.resourceId
    }
// omitted
```

specifically, this bit seems to fail
https://github.com/Azure/bicep-registry-modules/blob/1fa9ba802f539b4b4cdbc44e57e99a03c2952e17/avm/res/sql/server/main.bicep#L513-L531

xxxxxxx-Sql-Key fails with this message
```
{
    "status": "Failed",
    "error": {
        "code": "InvalidTemplate",
        "message": "Deployment template validation failed: 'The template variable 'serverKeyName' is not valid: The language expression property array index '5' is out of bounds.. Please see https://aka.ms/arm-functions for usage details.'.",
        "additionalInfo": [
            {
                "type": "TemplateViolation",
                "info": {
                    "lineNumber": 1,
                    "linePosition": 1787,
                    "path": "properties.template.variables.serverKeyName"
                }
            }
        ]
    }
}
```

seems like its trying to split my absolute URI to my key vault key, but it doesn't have a /version at the end, because auto rotate is enabled
https://github.com/Azure/bicep-registry-modules/blob/1fa9ba802f539b4b4cdbc44e57e99a03c2952e17/avm/res/sql/server/key/main.bicep#L17-L26

the input param for the -Sql-Key template
```
      "uri": {
            "value": "https://mysuperspecialname.vault.azure.net/keys/the-name-of-my-key-ekm/5f3a29cb740e465ba1b584d65383ed9f"
        }
```

### (Optional) Correlation Id

_No response_

peterbud:
## Description

This PR fixes a regression where server-level CMK deployments with `autoRotationEnabled: true` and no explicit `keyVersion` could fail for software-backed Key Vault keys. The child key module was still deriving the server key name from a `versionless` URI instead of honoring the explicit name passed from the parent module.

This change updates the key module to support that `autorotation` path correctly and adds regression coverage for the scenario.

Fixes #7073

## Pipeline Reference

| Pipeline |
| -------- |

Due to MCAPS policy enforcement, my pipelines are failing when the test creates an SQL audit setting backed by a storage - which is the default, so basically with a few exceptions, all other tests are failing in my MCAPS subscription - any idea how to avoid that would be appreciated.

## Type of Change

  - [ ] Update to documentation

## Checklist

### DIFF
--- a/avm/res/sql/server/CHANGELOG.md
+++ b/avm/res/sql/server/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 The latest version of the changelog can be found [here](https://github.com/Azure/bicep-registry-modules/blob/main/avm/res/sql/server/CHANGELOG.md).
 
+## 0.21.3
+
+### Changes
+
+- Fixed server-level customer-managed key autorotation when using a versionless Key Vault key URI
+- Added regression coverage for CMK autorotation without an explicit server key version
+
+### Breaking Changes
+
+- None
+
 ## 0.21.2
 
 ### Changes
--- a/avm/res/sql/server/README.md
+++ b/avm/res/sql/server/README.md
@@ -480,7 +480,6 @@ module server 'br/public:avm/res/sql/server:<version>' = {
       autoRotationEnabled: true
       keyName: '<keyName>'
       keyVaultResourceId: '<keyVaultResourceId>'
-      keyVersion: '<keyVersion>'
     }
     databases: [
       {
@@ -546,8 +545,7 @@ module server 'br/public:avm/res/sql/server:<version>' = {
       "value": {
         "autoRotationEnabled": true,
         "keyName": "<keyName>",
-        "keyVaultResourceId": "<keyVaultResourceId>",
-        "keyVersion": "<keyVersion>"
+        "keyVaultResourceId": "<keyVaultResourceId>"
       }
     },
     "databases": {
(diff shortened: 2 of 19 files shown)


AlexanderSehr:
Hey @peterbud,
regarding the tests I think you could only get around the policy challenges by using our visualstudio subscription and not the MCAPS one. 
Alternatively, we could merge this PR into a branch in upstream (from where we'd re-open this PR) and test in the upstream subscription. Not the prettiest approach as quite manual, but it would allow us your test your code while you cannot. Let me know what you think.

peterbud:
Hi @AlexanderSehr , yes, let's merge to an upstream branch and test it.

AlexanderSehr:
Sounds good. On it @peterbud

PR #7109, opened 2026-05-27
fix: avm/res/sql/server handle versionless CMK autorotation URIs

### LINKED ISSUE #7073: [AVM Module Issue]: `sql/server` autoRotationEnabled software backed CMK broken after #6193

AlexanderSehr:
## Description

> Follow-up to #7099

This PR fixes a regression where server-level CMK deployments with `autoRotationEnabled: true` and no explicit `keyVersion` could fail for software-backed Key Vault keys. The child key module was still deriving the server key name from a `versionless` URI instead of honoring the explicit name passed from the parent module.

This change updates the key module to support that `autorotation` path correctly and adds regression coverage for the scenario.

Fixes #7073

## Pipeline Reference

| Pipeline |
| -------- |
[avm.res.sql.server](https://github.com/Azure/bicep-registry-modules/actions/workflows/avm.res.sql.server.yml)

Due to MCAPS policy enforcement, my pipelines are failing when the test creates an SQL audit setting backed by a storage - which is the default, so basically with a few exceptions, all other tests are failing in my MCAPS subscription - any idea how to avoid that would be appreciated.

## Type of Change

  - [ ] Update to documentation

## Checklist


AlexanderSehr:
Hola @peterbud,
the new CMK implementation seems not to work. E.g., the [cmk-uami
](https://github.com/Azure/bicep-registry-modules/actions/runs/26540977344/job/78183095133) resulted in the error `The operation could not be completed because 'dep-***-kv-sscmk-mda_keyServerEncryptionKey_8ffa802125334ed4b419ef8589e52cd7' is an invalid Server Key name. Please provide a key name in the format of 'vault_key_version'. For example, if the keyId is https://yourvaultname.vault.azure.net/keys/YourKeyName/YourKeyVersion, then the Server Key Name should be formatted as: YourVaultName_YourKeyName_YourKeyVersion.` There are more failed tests, but that seems to be the only one inflicted by the changes.

Not sure why, as the logic appears to be correct and the error message seemingly contradicts itself. If you have an idea / are able to replicate the issue and suggest a change, please let us know. I'll be out of office until mid June starting this weekend, so if we could sort it before then it would be a dream come true 😉

peterbud:
Clearly, it's not working. I'm looking into it

AlexanderSehr:
Somehow not, no.

However, the key URI of the latest test run looks like this:

image

and the error like

>The operation could not be completed because 'dep-avmx-kv-sscmk-onf_keyServerEncryptionKey_f4e0118c7f6a42a3abcdaeb767d1e5df' is an invalid Server Key name. Please provide a key name in the format of 'vault_key_version'. For example, if the keyId is https://yourvaultname.vault.azure.net/keys/YourKeyName/YourKeyVersion, then the Server Key Name should be formatted as: YourVaultName_YourKeyName_YourKeyVersion. (Code: SecurityAzureKeyVaultInvalidKeyName)

So the name is definitely correct. Not sure what's happening here of if its a provider bug. I also tested it with a newer API version, but to no avail.

PS: This is what's passed as an input into the `/keys` deployment
image

peterbud:
I have spotted that the parent module (the server/main.bicep) is still overriding the `name` prop of the key module. I'll check it maybe that's the issue.

AlexanderSehr:
> I have spotted that the parent module (the server/main.bicep) is still overriding the `name` prop of the key module. I'll check it maybe that's the issue.

It does, true. It appears the `autorotationEnabled` is currently only implemeted in line 523 for the key uri, but not 517 for the key name. Can't wrap my head around it yet. Don't even understand why the RP would want to have the key version in the key name if no version in the uri. What even is the point then 😄. Why not `<keyVaultName>_<keyName>`. Ah well...

peterbud:
@eriqua - can you initiate a temp deployment while Alex is OOF? He has asked for quota support last week for zone redundant deployment, so if that has granted, we should see a proper deployment run.

eriqua:
> @eriqua - can you initiate a temp deployment while Alex is OOF? He has asked for quota support last week for zone redundant deployment, so if that has granted, we should see a proper deployment run.

@peterbud I've fixed the zone redundant failures for max and waf tests by changing the enforced deployment location. This was done via PR #7139. UK South is lately experiencing capacity pressures, hence I suspect the related quota restrictions.

Deployments are currently running, once finished I'll mark the PR ready for review

peterbud:
Thanks @eriqua - it seems the temp deployment was successful!

peterbud (approved):

PR #7145, opened 2026-06-10
fix: `avm/res/app/container-app` - add `.NET` runtime support to fix env failed state after Portal-enabled .NET stack

### LINKED ISSUE #7129: [AVM Module Issue]: Container Apps Environment enters failed state after enabling .NET Development Stack and running subsequent deployment

### Issue Type?

Bug

### Module Name

avm/res/app/container-app

### (Optional) Module Version

_No response_

### Description

After enabling the .NET Development Stack on a Container App via the Azure Portal, subsequent deployments using the AVM module fail, and the associated Container Apps Managed Environment transitions to a failed provisioning state.

```
"details": [
    {
        "code": "OperationFailed",
        "message": "Update environment AKS handler failed: 'R' is an invalid start of a value. Path: $ | LineNumber: 0 | BytePositionInLine: 0.."
    }
]
```

### (Optional) Correlation Id

_No response_

oZakari:
## Description

Closes #7129.

When the **.NET Development Stack** is enabled on a Container App via the Azure Portal, subsequent deployments of the AVM module fail and the associated **Container Apps Managed Environment** transitions to a `Failed` provisioning state with:

```
Update environment AKS handler failed: 'R' is an invalid start of a value. Path: $ | LineNumber: 0 | BytePositionInLine: 0..
```

### Root cause

The `runtime` parameter was strictly typed against the GA `Microsoft.App/containerApps@2026-01-01` Bicep schema:

```bicep
param runtime resourceInput<'Microsoft.App/containerApps@2026-01-01'>.properties.configuration.runtime?
```

The GA `2026-01-01` schema only exposes `runtime.java`. However, the Azure backend (and Portal) supports `runtime.dotnet.autoConfigureDataProtection` — see the `2025-10-02-preview` API version where `RuntimeDotnet` is defined ([change log](https://learn.microsoft.com/en-us/azure/templates/Microsoft.App/change-log/containerapps#2025-10-02-preview)).

Because the strict type rejected `dotnet`, callers had no way to set or preserve `runtime.dotnet` in the AVM module, leading to the failure described above whenever the .NET stack had been enabled out-of-band.

### Fix

- Replaced the `runtime` parameter type with a new user-defined `runtimeType` that exposes both `java.enableMetrics` and `dotnet.autoConfigureDataProtection`.
- Added a `#disable-next-line BCP037` on the `runtime: runtime` assignment, since the `dotnet` property is supported by the underlying API but not by the GA `2026-01-01` Bicep type definition.
- Extended the `max` E2E test to include `runtime.dotnet.autoConfigureDataProtection` to provide validation coverage for the new property.
- Bumped the module version from `0.22` to `0.23`.

## Pipeline Reference

[avm.res.app.container-app](https://github.com/oZakari/bicep-registry-modules/actions/workflows/avm.res.app.container-app.yml)

## Type of Change

- [ ] Update to CI Environment or utilities
    - [ ] New Module
    - [x] Minor version bump (`0.22` → `0.23`) - non-breaking change
    - [ ] Major version bump - breaking change
    - [ ] Update to documentation

## Checklist

- [x] I have run `Set-AVMModule` locally to generate the supporting module files

### DIFF
--- a/avm/res/app/container-app/CHANGELOG.md
+++ b/avm/res/app/container-app/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 The latest version of the changelog can be found [here](https://github.com/Azure/bicep-registry-modules/blob/main/avm/res/app/container-app/CHANGELOG.md).
 
+## 0.23.0
+
+### Changes
+
+- The `runtime` parameter now uses a custom user-defined type that exposes both the `java` and `dotnet` runtime configurations. This allows callers to set `runtime.dotnet.autoConfigureDataProtection`, which is supported by the underlying Azure API but not yet part of the GA `Microsoft.App/containerApps@2026-01-01` Bicep type definition. Resolves [#7129](https://github.com/Azure/bicep-registry-modules/issues/7129) by allowing the .NET stack configuration to be preserved across deployments (e.g., when the .NET Development Stack has been enabled via the Azure Portal).
+
+### Breaking Changes
+
+- None
+
 ## 0.22.1
 
 ### Changes
--- a/avm/res/app/container-app/README.md
+++ b/avm/res/app/container-app/README.md
@@ -505,6 +505,9 @@ module containerApp 'br/public:avm/res/app/container-app:<version>' = {
       }
     ]
     runtime: {
+      dotnet: {
+        autoConfigureDataProtection: false
+      }
       java: {
         enableMetrics: true
       }
@@ -671,6 +674,9 @@ module containerApp 'br/public:avm/res/app/container-app:<version>' = {
     },
     "runtime": {
       "value": {
+        "dotnet": {
+          "autoConfigureDataProtection": false
+        },
         "java": {
           "enableMetrics": true
         }
@@ -821,6 +827,9 @@ param roleAssignments = [
   }
 ]
 param runtime = {
+  dotnet: {
+    autoConfigureDataProtection: false
+  }
   java: {
     enableMetrics: true
   }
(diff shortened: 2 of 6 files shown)


avm/res/app/container-app/CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 
 The latest version of the changelog can be found [here](https://github.com/Azure/bicep-registry-modules/blob/main/avm/res/app/container-app/CHANGELOG.md).
 
+## 0.23.0
+
+### Changes
+
+- The `runtime` parameter now uses a custom user-defined type that exposes both the `java` and `dotnet` runtime configurations. This allows callers to set `runtime.dotnet.autoConfigureDataProtection`, which is supported by the underlying Azure API but not yet part of the GA `Microsoft.App/containerApps@2026-01-01` Bicep type definition. Resolves [#7129](https://github.com/Azure/bicep-registry-modules/issues/7129) by allowing the .NET stack configuration to be preserved across deployments (e.g., when the .NET Development Stack has been enabled via the Azure Portal).
```
eriqua, line 9: Hey @oZakari this is an interesting one and great we can workaround the resource derived type's bug with UDT. Just a question: do you think it would be worthy to raise the bug with the right group? Maybe we did that already, I'm just curious. That would be to help improving the next version, so that if useful we can go back to RDT next. Interested in your thoughts. Looks good to me for the rest

eriqua: Another question: is the feature preview? I don't mean the API but the feature itself. Asking as if it is, it would be good to also include a note about it in the parameter description (as per [SFR1](https://azure.github.io/Azure-Verified-Modules/spec/SFR1/)).

jtracey93 (approved):

PR #178, opened 2022-10-09
Import ACR Images - Better output names

Gordonby:
## Description

Refining the outputs from the module, to ease the consumption of it by other modules.

#### Context 

Because of the uri portion to the names, it was getting stuck when being consumed in a quickstart, in the azure quickstart repo CI.

## Updating an existing module

- [x] I have run `brm validate` locally to verify the module files.
- [x] I have run deployment tests locally to ensure the module is deployable.
- [x] I have read the [Updating an existing module](../CONTRIBUTING.md#updating-an-existing-module) section in the contributing guide and updated the `version.json` file properly:
  - [x] The PR contains breaking changes, and I have bumped the MAJOR version in `version.json`.

PR #7233, opened 2026-07-23
ci: Refresh allowed e2e test regions from Azure capacity list

jtracey93:
## Description

Refreshes the default `AllowedRegionsList` used by the e2e validation region selector (`utilities/pipelines/e2eValidation/regionSelector/Get-AvailableResourceLocation.ps1`) to the currently "green" regions from the Azure capacity list (https://aka.ms/azurecapacity), as agreed in the AVM PG + Core Team sync on 23 Jul 2026.

`uksouth` is no longer a reliable default and has been a source of avoidable static/deployment test failures, so it has been removed (along with `northeurope` and `germanywestcentral`).

**Changes:**
- Replaced the allowed region set with the refreshed green regions: `swedencentral`, `belgiumcentral`, `norwayeast`, `eastasia` (retained as an edge region), `polandcentral`, `austriaeast`, `denmarkeast`, `koreacentral`, `eastus`, `centralus`, `newzealandnorth`.
- Removed `koreacentral` from `ExcludedRegions` so it can actually be selected for resource modules, since it is now part of the allowed green-region set. (For resource modules, excluded regions are filtered out *before* the allowed-list intersection, so leaving it in both lists would have kept it from ever being picked.)
- Added an inline note pointing to the capacity list and a reminder to re-review the list quarterly.

The effective region for resource modules remains the intersection of this allowed list and per-resource feature availability (via `Get-AzLocation`). Non-`Recommended`/excluded regions continue to be filtered as before.

> Tracked internally in the AVM PG + Core Team backlog (item #178). No `version.json`/module files are affected by this change.

## Pipeline Reference

| Pipeline |
| -------- |
| _N/A — CI/utilities change; no module pipeline is affected._ |

## Type of Change

  - [ ] Update to documentation

## Checklist

### DIFF
--- a/utilities/pipelines/e2eValidation/regionSelector/Get-AvailableResourceLocation.ps1
+++ b/utilities/pipelines/e2eValidation/regionSelector/Get-AvailableResourceLocation.ps1
@@ -34,11 +34,19 @@ function Get-AvailableResourceLocation {
 
         [Parameter(Mandatory = $false)]
         [array] $AllowedRegionsList = @(
+            # Refreshed to currently "green" regions from the Azure capacity list (https://aka.ms/azurecapacity)
+            # per the AVM PG + Core Team sync on 23 Jul 2026. Re-review this list quarterly against the capacity list.
+            'swedencentral',
+            'belgiumcentral',
+            'norwayeast',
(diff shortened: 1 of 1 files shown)


thomascoats (approved):
LGTM 👍

PR #7234, opened 2026-07-24
feat: Auto-rerun failed module pipelines on main with a hard stop

jtracey93:
## Description

Fixes transient/capacity CI failures on `main` that otherwise require someone to manually re-kick module pipelines. When a module is merged to `main` and its deploy step fails, the publish tag is never created and the module is never published to the registry.

This wraps the existing `utilities/tools/Invoke-WorkflowsFailedJobsReRun.ps1` (amended) in a scheduled platform workflow that re-runs **only the failed jobs** of the latest failed `main`-branch run of each module workflow, as a **new attempt of the same run** (so it stays tied to the original commit), with a **hard stop** once a run reaches 3 total attempts.

Addresses internal issue: [azure-cloud-native/Azure-Verified-Modules#179](https://msft.ghe.com/azure-cloud-native/Azure-Verified-Modules/issues/179)

### What changed

- **Amend `utilities/tools/Invoke-WorkflowsFailedJobsReRun.ps1`**
  - Add `MaxAttempts` (default `3`) and `TargetBranch` (default `main`) parameters.
  - Hard stop based on GitHub's `run_attempt`: a run is re-triggered only while `run_attempt < MaxAttempts`; at `>= 3` it is left alone for manual investigation.
  - Non-interactive: removed the `Read-Host` "apply? (y/n)" prompt so it is CI-safe; behaviour is driven by `-WhatIf`.
  - Guard the `Invoke-ReRun` call when there is nothing to re-trigger — fixes a latent bind error where the mandatory `[object[]]` parameter rejects an empty array, which would otherwise throw on every healthy scan (all runs green / already hard-stopped).
  - Add an end-of-run summary and fix a pre-existing PAT assignment bug.
- **Add `.github/workflows/platform.rerun-failed-workflows.yml`**
  - Scheduled every 90 minutes (two crons), plus `workflow_dispatch` with a `whatIf` boolean for dry-runs.
  - `permissions: { actions: write, contents: read }`, a `concurrency` guard, a fork/schedule guard, and the built-in `GITHUB_TOKEN`.
- **Add `utilities/tests/pipelines/Invoke-WorkflowsFailedJobsReRun.Tests.ps1`**
  - Pester coverage for the hard-stop logic (re-run below limit, hard stop at/above limit, success, in-progress, `-WhatIf` no-op, custom `MaxAttempts`, mixed set). Runs automatically via `Test-CI.ps1`.

### Notes for reviewers

- **Attempt semantics:** `MaxAttempts = 3` is the **total** attempt cap, inclusive of the original merge-to-`main` run (i.e. original + 2 automated re-runs). Once `run_attempt` reaches 3, no further re-runs.
- **Region is recalculated on each rerun (no code needed):** the deploy region is chosen at runtime by the "Get Resource Location" step (`Get-AvailableResourceLocation.ps1` → `Get-Random`), not frozen in the matrix at trigger time, so a rerun re-rolls the region — the intended "retry in a different region" recovery. Applies whenever `customLocation` is empty (always true for push-to-`main`).
- **Issue cleanup already handled:** the daily `platform.manage-workflow-issue.yml` → `Set-AvmGitHubIssueForWorkflow.ps1` already closes `[Failed pipeline] <name>` issues once the latest `main` run is green (no change here — confirmed only).

### Validation

- `Invoke-ScriptAnalyzer` (Warning + Error): clean on the amended script and the new test.
- Pester: **9/9 pass**.
- Workflow YAML parses; triggers `schedule` + `workflow_dispatch`; jobs `job_initialize_pipeline`, `rerun-failed-workflows`.
- Recommended post-merge: trigger via `workflow_dispatch` with `whatIf: true` for a live dry-run before relying on the schedule.

/cc @alexandersehr — could you review the approach please? 🙏

## Pipeline Reference

| Pipeline |
| -------- |
| N/A — platform/CI utility change (no module affected) |

## Type of Change

  - [ ] Update to documentation

## Checklist


.github/workflows/platform.rerun-failed-workflows.yml
```diff
@@ -0,0 +1,77 @@
...
+        with:
+          fetch-depth: 0
+      - name: "Set input parameters to output variables"
+        id: get-workflow-param
+        uses: ./.github/actions/templates/avm-getWorkflowInput
+        with:
+          workflowPath: "${{ env.workflowPath}}"
+    outputs:
+      workflowInput: ${{ steps.get-workflow-param.outputs.workflowInput }}
+  rerun-failed-workflows:
```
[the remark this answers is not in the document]
jtracey93: Good catch — added a top-level `permissions: { contents: read }` block so the default `GITHUB_TOKEN` is read-only across the whole workflow. `job_initialize_pipeline` inherits that read-only default (it only checks out the repo and reads workflow inputs), while `rerun-failed-workflows` keeps its explicit `actions: write` override, which is required for the `rerun-failed-jobs` REST endpoint.


> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

utilities/tools/Invoke-WorkflowsFailedJobsReRun.ps1
```diff
@@ -227,35 +238,53 @@ function Invoke-WorkflowsFailedJobsReRun {
...
-            $userInput = Read-Host -Prompt 'Should apply (y/n)?'
-        }
-        while ($userInput -notin @('y', 'n'))
-
-        switch ($userInput) {
-            'y' {
-                $null = Invoke-ReRun @reRunInputObject -WhatIf:$false
-            }
-            'n' { return }
-        }
```
AlexanderSehr, line 257: This should not be removed as I use it very regulary when invoking the function. The idea: Because the function analyzes every workflow (which can take some time) it is very handy that I don't have to run the function with a 'whatif' first and then again without (which would perform the analysis twice).

Instead, I'd recommend to leave this in here and add an extra `-Force` or `-NonInteractive` parameter which would automatically assume a `n` in line 250 if specified.

jtracey93: Good call — restored the interactive `Should apply (y/n)?` prompt so your run-once-with-`-WhatIf`, preview, then confirm-to-apply workflow works exactly as before (the confirmed apply reuses the already-analyzed `$reRunInputObject`, so it doesn't re-run the slow analysis).

For the automated path I added a `-NonInteractive` switch that assumes `n` (preview only, never applies) so the scheduled workflow's `-WhatIf` dry-run doesn't block on `Read-Host` on the runner. The real scheduled run uses `-WhatIf:$false`, which skips the prompt block entirely and applies directly. Added Pester coverage for both the interactive `y` apply and the `-NonInteractive` skip.


> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

utilities/tools/Invoke-WorkflowsFailedJobsReRun.ps1
```diff
@@ -230,6 +242,10 @@ function Invoke-WorkflowsFailedJobsReRun {
     $currentCount = 1
     $runsToReTrigger = [System.Collections.ArrayList]@()
     $hardStoppedRuns = [System.Collections.ArrayList]@()
+    $staleRuns = [System.Collections.ArrayList]@()
+    # When [LookbackDays] > 0, only runs created on/after this cutoff are eligible; older
+    # runs are skipped so an initial invocation does not re-trigger a large stale backlog.
+    $lookbackCutoff = if ($LookbackDays -gt 0) { ([datetimeoffset]::UtcNow).AddDays(-$LookbackDays) } else { $null }
```
AlexanderSehr: Would recommend to remove that feature again. 
If you check the `Get-GitHubModuleWorkflowLatest` script, it get's only exacty one workflow which is the latest for the given branch. So e.g., you get only the latest run for the main branch. If that one is 1 day old or 15 doesn't matter. 

The query url:
https://github.com/Azure/bicep-registry-modules/blob/1a0f77a459a22db1f6e3c6209de8ec7add6454b0/utilities/pipelines/platform/helper/Get-GitHubModuleWorkflowLatestRun.ps1#L48

I'd even argue that the lookback complicates the logic - without actually changing anything (as we've gotten only 1 run before and will still only get 1 run (or none)).
Or in other words, it makes to code less maintainable for no gain 😉

jtracey93: Good call — removed the lookback again in e3b09ffb4. You're right that `Get-GitHubModuleWorkflowLatestRun` only ever returns the single latest run per workflow, so the volume is already bounded to at most one re-run per module, and filtering that one run by age would actually work against the whole point here (an old failed `main` run is exactly the stuck module we want to recover). The initial-burst concern that prompted it can be previewed operationally via the workflow's `whatIf` dry-run instead. Hard-stop (`MaxAttempts`) is unchanged and tests are back to 11/11.


> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

AlexanderSehr (approved):

PR #7292, opened 2026-08-20
fix: Recovery Services Vault - Omit soft delete properties from `backupConfig` when `softDeleteSettings` is used

### LINKED ISSUE #7289: [AVM Module Issue]: Recovery-Services/vault softDeleteSettings vs backupConfig

### Issue Type?

Feature Request

### Module Name

avm/res/recovery-services/vault

### (Optional) Module Version

_No response_

### Description

Module `avm/res/recovery-services/vault` has two different options to configure Soft Delete:
- Parameter `softDeleteSettings`, which is defined as: 
  ```
  type softDeleteSettingType = {
      enhancedSecurityState: ('AlwaysON' | 'Disabled' | 'Enabled' | 'Invalid')
      softDeleteRetentionPeriodInDays: int
      softDeleteState: ('AlwaysON' | 'Disabled' | 'Enabled' | 'Invalid')
  }
  ```
- Parameter `backupConfig`, whose properties are used to create child resource `avm/res/recovery-services/vault/backup-config`, and (among others) has the properties:
  ```
  enhancedSecurityState
  softDeleteFeatureState
  isSoftDeleteFeatureStateEditable
  softDeleteRetentionPeriodInDays
  ```

If both parameters exist, the first two properties of the `backupConfig` parameter will not be set, effectively falling back to the equivalent properties from `softDeleteSettings`:
```
enhancedSecurityState: !empty(softDeleteSettings) ? null : backupConfig.?enhancedSecurityState
softDeleteFeatureState: !empty(softDeleteSettings) ? null : backupConfig.?softDeleteFeatureState
```
The latter two do not follow that behaviour, and will get set:
```
isSoftDeleteFeatureStateEditable: backupConfig.?isSoftDeleteFeatureStateEditable
softDeleteRetentionPeriodInDays: backupConfig.?softDeleteRetentionPeriodInDays
```

A few observations:
1. The `softDeleteSettings` parameter properties `enhancedSecurityState` and `softDeleteState` will overrule the corresponding properties in the `backupConfig` parameter, when both are passed as input to a deployment.
1. Both parameters contain a third shared property, `softDeleteRetentionPeriodInDays`, which can be abiguously passed with different values, and will not be overruled.
1. The `backupConfig` parameter has a property `isSoftDeleteFeatureStateEditable`, which appears to be relevant to the Soft Delete feature, but can't be configured using the `softDeleteSettings` parameter. According to the bicep/ARM documentation: "_This flag is no longer in use. Please use 'softDeleteFeatureState' to set the soft delete state for the vault_".

A few questions:
1. Which value for `softDeleteRetentionPeriodInDays` is leading, and shouldn't the override mechanism be implemented for this parameter as well?
1. Is the usage of a specific one of the parameters preferable over the other, for example because the other may be deprecated in the future?
1. Since we would like to use only one of these parameters whenever possible, do the other properties of `backupConfig`, such as `storageModelType`, `storageType` and `storageTypeState`, also map to `recovery-services/vault` parameters, so we can be sure to configure them only once?

### (Optional) Correlation Id

_No response_

alexanderojala:
## Description

Fixes #7289

Deploying `avm/res/recovery-services/vault` with both the vault-level `softDeleteSettings` and the `backupConfig` child configuration fails, because the two write the same underlying service state through two different APIs.

Once soft delete has been configured via the **Vault API** (`Microsoft.RecoveryServices/vaults`), the **backupconfig API** (`Microsoft.RecoveryServices/vaults/backupconfig`) permanently rejects any further write to the soft delete related properties with:

> Soft delete state for this vault cannot be modified using this API. Since the Vault API was previously used to update the soft delete property for this vault, you must again use the Vault API to make any further changes to this property. (Code: `BMSUserErrorSoftDeleteUseVaultApi`)

This happens even when the values being written are identical to the ones already set on the vault, so the two configurations cannot be reconciled by forwarding matching values. The only viable resolution is for the `backupconfig` request to not carry those properties at all.

### Changes

- **`main.bicep`** — when `softDeleteSettings` is provided, the parent no longer forwards `enhancedSecurityState`, `softDeleteFeatureState`, `isSoftDeleteFeatureStateEditable` and `softDeleteRetentionPeriodInDays` to the `backup-config` child module.
- **`backup-config/main.bicep`** — these four properties are now emitted conditionally, so an unset property is **absent** from the request payload rather than present with a `null` value. Sending them as `null` is itself rejected by the API with `UserErrorInvalidRequestParameter` ("Parameter NO_PARAM in request is invalid").
- **`backup-config/main.bicep`** — `isSoftDeleteFeatureStateEditable` changed from `bool` (default `true`) to a nullable `bool?`. When it is not specified, the service-side value is now left untouched instead of being implicitly set to `true`.
- **`backup-config/main.bicep`** — added `'Invalid'` to the allowed values of `enhancedSecurityState` to align with the API schema (`softDeleteFeatureState` already allowed it).
- **`tests/e2e/max/main.test.bicep`** — added `softDeleteRetentionPeriodInDays` and `isSoftDeleteFeatureStateEditable` to `backupConfig` so all four affected properties are exercised alongside `softDeleteSettings`.
- Updated both `CHANGELOG.md` files and regenerated the supporting module files via `Set-AVMModule`.

### Validation

Both failure modes were reproduced in CI before the fix, which brackets the solution:

| Payload for the soft delete properties | Result |
| -------------------------------------- | ------ |
| Present, value `null`                  | ❌ `UserErrorInvalidRequestParameter` |
| Present, values matching the vault     | ❌ `BMSUserErrorSoftDeleteUseVaultApi` |
| Absent                                 | ✅ Succeeds |

Both `max` and `waf-aligned` configure `softDeleteSettings` together with `backupConfig` soft delete properties, so both cover the reported scenario, and both now deploy successfully.

### Notes for reviewers

Two points that are deliberate but worth a second opinion:

1. If a consumer sets both `softDeleteSettings` and the soft delete properties of `backupConfig`, the latter are now **silently ignored** rather than causing a failure. The API offers no way to honour them, but an explicit `fail()` on the conflicting combination would be an alternative if preferred.
2. Removing the `true` default from `isSoftDeleteFeatureStateEditable` is a behaviour change for consumers deploying the `backup-config` module standalone. It has been classified under *Changes* rather than *Breaking Changes* since no deployment breaks and the property remains service-managed, but happy to reclassify.

## Pipeline Reference

| Pipeline |
| -------- |
| [avm.res.recovery-services.vault](https://github.com/alexanderojala/bicep-registry-modules/actions/workflows/avm.res.recovery-services.vault.yml) |

## Type of Change

  - [ ] Update to documentation

## Checklist

### DIFF
--- a/avm/res/recovery-services/vault/CHANGELOG.md
+++ b/avm/res/recovery-services/vault/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 The latest version of the changelog can be found [here](https://github.com/Azure/bicep-registry-modules/blob/main/avm/res/recovery-services/vault/CHANGELOG.md).
 
+## 0.13.1
+
+### Changes
+
+- Fixed the deployment of the `backup-config` child module failing when vault-level `softDeleteSettings` is provided together with `backupConfig`. The soft delete related properties of `backupConfig` are now omitted from the `backupconfig` request payload entirely, as the API rejects any write to these properties once they were set through the vault itself. [#7289](https://github.com/Azure/bicep-registry-modules/issues/7289)
+
+### Breaking Changes
+
+- None
+
 ## 0.13.0
 
 ### Changes
--- a/avm/res/recovery-services/vault/README.md
+++ b/avm/res/recovery-services/vault/README.md
@@ -561,7 +561,9 @@ module vault 'br/public:avm/res/recovery-services/vault:<version>' = {
     // Non-required parameters
     backupConfig: {
       enhancedSecurityState: 'AlwaysON'
+      isSoftDeleteFeatureStateEditable: true
       softDeleteFeatureState: 'AlwaysON'
+      softDeleteRetentionPeriodInDays: 7
     }
     backupPolicies: [
       {
@@ -968,7 +970,9 @@ module vault 'br/public:avm/res/recovery-services/vault:<version>' = {
     "backupConfig": {
       "value": {
         "enhancedSecurityState": "AlwaysON",
-        "softDeleteFeatureState": "AlwaysON"
+        "isSoftDeleteFeatureStateEditable": true,
+        "softDeleteFeatureState": "AlwaysON",
+        "softDeleteRetentionPeriodInDays": 7
       }
     },
     "backupPolicies": {
@@ -1401,7 +1405,9 @@ param name = 'rsvmax001'
 // Non-required parameters
 param backupConfig = {
   enhancedSecurityState: 'AlwaysON'
+  isSoftDeleteFeatureStateEditable: true
   softDeleteFeatureState: 'AlwaysON'
+  softDeleteRetentionPeriodInDays: 7
 }
 param backupPolicies = [
   {
--- a/avm/res/recovery-services/vault/backup-config/CHANGELOG.md
+++ b/avm/res/recovery-services/vault/backup-config/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 The latest version of the changelog can be found [here](https://github.com/Azure/bicep-registry-modules/blob/main/avm/res/recovery-services/vault/backup-config/CHANGELOG.md).
 
+## 0.1.2
+
+### Changes
+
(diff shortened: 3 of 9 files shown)


AlexanderSehr (approved):
Hola @alexanderojala,
please excuse the late approval. I was out on vacation and it seems no one else jumped on it. 
Anyways, looks good :)

PR #7330, opened 2026-09-05
fix: update skuCapacity parameter description

### LINKED ISSUE #7278: [AVM Module Issue]: API management `skuCapacity` default of `3` fails deployment when `sku` is `Basic` (max capacity 2)

### Issue Type?

Bug

### Module Name

avm/res/api-management/service

### (Optional) Module Version

0.14.4

### Description

The module declares:

```bicep
@description('Conditional. The scale units for this API Management service. Required if using Basic, Standard, or Premium skus. For range of capacities for each sku, reference https://azure.microsoft.com/en-us/pricing/details/api-management/.')
param skuCapacity int = 3
```

`skuCapacity` defaults to `3` regardless of the chosen `sku`. This default is invalid for the `Basic` tier, which only supports a capacity of `1` or `2`. When a consumer sets `sku: 'Basic'` without explicitly overriding `skuCapacity`, deployment fails with:

```
OutOfBoundSkuCapacityForSkuType: Specified Sku capacity of 3 is invalid. For Basic Sku it must be between 1 and 2, inclusive. Add CustomSettings if needed.
```

This is surprising because the module's `Developer` handling already accounts for tier-specific constraints (capacity is hardcoded to `1` for `Developer` in the `service` resource: `contains(sku, 'Developer') ? 1 : skuCapacity`), but no equivalent safeguard/default exists for `Basic`.

# Repro

```bicep
module apiManagement 'br/public:avm/res/api-management/service:0.14.4' = {
  params: {
    name: 'apim-example'
    location: location
    publisherEmail: 'api@example.com'
    publisherName: 'Example'
    sku: 'Basic'
    // skuCapacity omitted, defaults to 3 -> deployment fails
  }
}
```

# Expected behavior / suggested fixes

- Lower the default `skuCapacity` to a value valid across all SKU tiers (e.g. `1` or `2`), since `1` is valid for every tier including `Basic` and `Developer`.
- Alternatively, add tier-aware capacity clamping/fallback logic in the `service` resource (similar to the existing `Developer` special-case), e.g. defaulting to `2` when `sku == 'Basic'` and `skuCapacity` is left unset/exceeds the tier's max.
- At minimum, document this pitfall more prominently in the parameter description, since the "Conditional... Required if using Basic, Standard, or Premium" wording doesn't make clear that the *default itself* is invalid for `Basic`.

# Workaround

Explicitly pass `skuCapacity: 1` (or `2`) when using `sku: 'Basic'`.

### (Optional) Correlation Id

_No response_

abhishekaryams:
- Clarified that the `Basic` SKU supports a maximum capacity of `2`, clamping the default of `3` to `2`.
- Updated related descriptions in README and main.bicep files for consistency.

## Description

Fixes #7278 
## Pipeline Reference

| Pipeline |
| -------- |

## Type of Change

  - [ ] Update to documentation

## Checklist

### DIFF
--- a/avm/res/api-management/service/README.md
+++ b/avm/res/api-management/service/README.md
@@ -3674,7 +3674,7 @@ param virtualNetworkType = 'None'
 
 | Parameter | Type | Description |
 | :-- | :-- | :-- |
-| [`skuCapacity`](#parameter-skucapacity) | int | The scale units for this API Management service. Required if using Basic, Standard, or Premium skus. For range of capacities for each sku, reference https://azure.microsoft.com/en-us/pricing/details/api-management/. |
+| [`skuCapacity`](#parameter-skucapacity) | int | The scale units for this API Management service. Required if using Basic, Standard, or Premium skus. Note: the `Basic` SKU supports a maximum capacity of `2`, so the default of `3` is automatically clamped to `2` for `Basic`. For the range of capacities for each sku, reference https://azure.microsoft.com/en-us/pricing/details/api-management/. |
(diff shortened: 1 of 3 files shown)


ReneHezser:
Hi @abhishekaryams, 
Thanks for the fix. Can you please add a pipeline run and update the Changelog?

PR #7343, opened 2026-09-14
chore: generate per-module CODEOWNERS from module owners

jaredfholgate:
## Description

Add the initial template-generated CODEOWNERS snapshot so the result can be reviewed before enabling automated synchronization.

Companion workflow: [Azure/azure-verified-modules-tools#112](https://github.com/Azure/azure-verified-modules-tools/pull/112).

- Generate 267 top-level module entries from the official Bicep resource, pattern, and utility indexes at [Azure/Azure-Verified-Modules commit 6142d822](https://github.com/Azure/Azure-Verified-Modules/commit/6142d822b65db7013818d717519a562e67ce664f). Parent directory rules recursively cover child modules; no separate child-module rows are needed.
- 237 entries contain named owners plus `@Azure/azure-verified-modules-module-owners` as an alternative reviewer. The remaining 30 entries contain only the group.
- Preserve the tooling catch-all and final core-team test/deployment-exclusion overrides. Use the owners group as the shared `/avm/` default.
- Add a generated-file warning and a link to the source template. The template link will resolve after the tools implementation is merged.
- Update governance validation to accept generated top-level entries, reject malformed/duplicate/child patterns, and retain static-rule precedence. Include coverage of the actual checked-in file.

The file was rendered locally and copied byte-for-byte from the tools generator. Removing the 339 redundant child rows does not change effective ownership: each had the same owners as its parent. No live synchronization, ruleset changes, permission changes, or merges were performed.

## Owner access findings

GitHub's CODEOWNERS API currently reports 27 **Unknown owner** references across these eight CSV owners:

| Owner | References |
| ----- | ---------- |
| `didayal-msft` | 1 |
| `elizatargithub7` | 1 |
| `juancj` | 2 |
| `kpantos` | 1 |
| `ozakari` | 7 |
| `pankajagrawal16` | 3 |
| `peterbud` | 4 |
| `sebassem` | 8 |

GitHub requires code owners to exist and have repository write access. A direct permission lookup confirms `sebassem` currently has only `read` access. The generated file preserves the CSV owners rather than silently dropping them. Resolve the affected handles/access before enabling automatic merging; the shared owners group itself is not flagged.

## Rollout

Merge this initial snapshot and governance compatibility change before enabling the companion tools synchronization. Its schedule is `33 2-23/4 * * *` (02:33, 06:33, 10:33, 14:33, 18:33, and 22:33 UTC), offset two hours from the existing repo-sync slots. Scheduled merging requires `AVM_CODEOWNERS_SYNC_ENABLED=true`. Manual dispatch defaults to plan-only, which may create/update the stable sync change request but never merge it.

## Pipeline Reference

| Pipeline |
| -------- |
| Local Pester: `CodeOwnersGovernance.Tests.ps1` — 33 passed, including the generated file |
| Local Pester: `ModuleOwnership.Tests.ps1` — 54 passed |

## Type of Change

  - [ ] Update to documentation

## Checklist

- [ ] I have run `Set-AVMModule` locally to generate the supporting module files. (Not applicable: no module files changed.)
- [ ] I have updated the module's CHANGELOG.md file with an entry for the next version. (Not applicable: no module files changed.)


utilities/pipelines/staticValidation/compliance/module.tests.ps1
```diff
@@ -2140,22 +2140,33 @@ Describe 'Governance tests' {
...
             '/avm/ @Azure/azure-verified-modules-module-contributors'
+            '/avm/ @Azure/azure-verified-modules-module-owners'
+        )
+        $ownershipRules[-2..-1] | Should -Be @(
             '*avm.core.team.tests.ps1 @Azure/azure-verified-modules-tooling-contributors'
             '*.e2eignore @Azure/azure-verified-modules-tooling-contributors'
-        )
+        ) -Because 'tooling overrides must take precedence over module ownership.'
+
+        $moduleOwnershipRules = @($ownershipRules | Select-Object -Skip 2 | Select-Object -SkipLast 2)
```
AlexanderSehr: Maybe be a bit too easy to break. How about just a `$ownershipRules | Where-Object { $_ -like "/avm/*" }`?

utilities/pipelines/staticValidation/compliance/module.tests.ps1
```diff
@@ -2140,22 +2140,33 @@ Describe 'Governance tests' {
...
+        $individualOwnerPattern = '@(?=[a-zA-Z0-9-]{1,39}(?: |$))[a-zA-Z0-9]+(?:-[a-zA-Z0-9]+)*'
+        $moduleOwnershipPattern = "^/avm/(res|ptn|utl)/(?:[a-z0-9-]+/){2} ($individualOwnerPattern )*@Azure/azure-verified-modules-module-owners$"
```
AlexanderSehr, line 2171: Should this not also check for the actual owning team via the .csv - or, because we'll introduce the manifest sooner or later, should we keep it simple?

jaredfholgate: I'd keep this one simple. The sync in [Azure/azure-verified-modules-tools#112](https://github.com/Azure/azure-verified-modules-tools/pull/112) already reads the owners from the CSVs and checks the generated file before merging it. Repeating that lookup here could fail unrelated module checks between an owner change and the next sync. When the manifest is ready, we can switch the sync to that instead. This check can stay focused on the file format and tooling ownership.

utilities/pipelines/staticValidation/compliance/module.tests.ps1
jaredfholgate: Good point. I've changed this to select the `avm/res`, `avm/ptn` and `avm/utl` paths rather than relying on line positions. That also keeps the `/avm/` fallback out of the list. The tooling rules and override order are still checked separately.

jtracey93 (approved):

